PR #6, opened 2022-08-26
frontend: Update Easter egg

juliaogris:
Update Easter egg. The intention was to update this interactively in a
coding session with some kids, but we ran out of time before we got to
it, so tune the current Easter egg and keep it🐣.

### DIFF
--- a/frontend/index.js
+++ b/frontend/index.js
@@ -52,9 +52,9 @@ initWasm()
 // When source input string contains the sub string "confetti"
 // show confetti when clicking Run button.
 function showConfetti() {
-  const names = ['?']
-  const colors = ['red']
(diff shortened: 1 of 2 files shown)


frontend/index.js
```diff
@@ -52,9 +52,9 @@ initWasm()
 // When source input string contains the sub string "confetti"
 // show confetti when clicking Run button.
 function showConfetti() {
-  const names = ['?']
-  const colors = ['red']
-  let confetti = new Array(10)
+  const names = ['🦊', '🐐']
```
camh-, line 55: :blue_heart:

camh- (approved):
Love it! :green_heart:

PR #8, opened 2022-08-31
lexer: Initialise lexer

juliaogris:
Initialise lexer. All tokens according to to syntax grammar are covered.
Many pointers are taken from Thorsten Ball's [Interpreter book] code. The
function

	func lexer.Run(string) string

generates a stream of tokens for debug purposes.

Move assert package to pkg/assert. This has been overlooked in a
previous refactor.

[Interpreter book]: https://github.com/juliaogris/monkey/blob/master/lexer/lexer.go

Play with it at https://evy-lang--pr8-lexing-m4obheh1.web.app


pkg/lexer/lexer.go
```diff
@@ -1,5 +1,208 @@
...
+type Lexer struct {
+	input []rune
+	r     rune // current rune under examination
+	pos   int  // current position in input (points to current rune)
+	line  int
+	col   int
+}
+
+func New(input string) *Lexer {
+	return &Lexer{input: []rune(input), pos: -1, col: -1}
```
camh-, line 21: I think `col` should be `0` - it is reinitialised to zero at the start of every line, and columns should probably count from 1 like lines.

pkg/lexer/lexer.go
```diff
@@ -1,5 +1,208 @@
...
+}
+
+func (l *Lexer) peekRune() rune {
+	if l.pos+1 >= len(l.input) {
+		return 0
+	}
+	return l.input[l.pos+1]
+}
+
+func (l *Lexer) peekRune2() rune {
```
camh-, line 148: i think this should be called `peeekRune()` :tongue:

pkg/lexer/lexer.go
```diff
@@ -1,5 +1,215 @@
...
+		// EOF or NL before closing `"`
+		if l.r != '"' {
+			return tok.SetType(ILLEGAL).SetLiteral(`"`)
+		}
+		return tok.SetType(STRING_LIT).SetLiteral(literal)
+	case 0:
+		return tok.SetType(EOF)
+	}
+	if isLetter(l.r) {
+		lit := l.readIdent()
```
camh-, line 141: nit: you call this var `literal` a few lines above. perhaps keep the two the same?

pkg/lexer/lexer.go
```diff
@@ -1,5 +1,215 @@
...
+	if l.pos+2 >= len(l.input) {
+		return 0
+	}
+	return l.input[l.pos+2]
+}
+
+func (l *Lexer) readComment() string {
+	pos := l.pos
+	pr := l.peekRune()
+	for pr != 0 && pr != '\n' {
```
camh-, line 165: You could write this loop as:
```
        for pr := l.peekRune(); pr != 0 && pr != '\n'; pr = l.peekRune() {
                l.readRune()
        }
```

Not sure if you feel that is any clearer. It's no fewer expressions, but two of them can go onto the `for` line.

pkg/lexer/lexer.go
```diff
@@ -1,5 +1,215 @@
...
+	pos := l.pos
+	pr := l.peekRune()
+	for isDigit(pr) || pr == '.' {
+		l.readRune()
+		pr = l.peekRune()
+	}
+	return string(l.input[pos : l.pos+1])
+}
+
+func (l *Lexer) readIdent() string {
```
camh-, line 199: `readComment`, `readNum` and `readIdent` could be done as (just an idea):
```
func (l *Lexer) readWhile(pred func(rune) bool) string {
        pos := l.pos
        for pr := l.peekRune(); pred(pr); pr = l.peekRune() {
                l.readRune()
        }
        return string(l.input[pos : l.pos+1])
}

func (l *Lexer) readComment() string {
        return l.readWhile(func(r rune) bool { return r != 0 && r != '\n' })
}

func (l *Lexer) readNum() string {
        return l.readWhile(func(r rune) bool { return isDigit(r) || r == '.' })
}

func (l *Lexer) readIdent() string {
        return l.readWhile(func(r rune) bool { return isLetter(r) || unicode.IsDigit(r) })
}
```

It's a shame anonymous Go functions have so much boilerplate. Would be nicer with an arrow syntax.

pkg/lexer/lexer.go
```diff
@@ -1,5 +1,215 @@
...
+	line  int
+	col   int
+}
+
+func New(input string) *Lexer {
+	return &Lexer{input: []rune(input), pos: -1, line: 1}
+}
+
+func (l *Lexer) readRune() {
+	l.pos += 1
```
camh-, line 32: What do you think of this for the read/peek rune functions?

```
func (l *Lexer) readRune() {
        if l.r == '\n' {
                l.line++
                l.col = 0
        }
        l.pos++
        l.col++
        l.r = l.lookAt(l.pos)
}

func (l *Lexer) peekRune() rune {
        return l.lookAt(l.pos + 1)
}

func (l *Lexer) peekRune2() rune {
        return l.lookAt(l.pos + 2)
}

func (l *Lexer) lookAt(pos int) rune {
        if pos >= len(l.input) {
                return 0
        }
        return l.input[pos]
}
```

pkg/lexer/lexer_test.go
```diff
@@ -0,0 +1,252 @@
...
+		got := l.Next()
+		assert.Equal(t, &want, got, "want %s got %s", want.Type.String(), got.Type.String())
+	}
+}
+
+func TestStrings(t *testing.T) {
+	in := `
+s:string
+s = "abc"
+s = s[:1] // "bc"`
```
camh-, line 133: It would be good to have a test that has a comment and a line after it to ensure comments are read correctly wrt the following line.

camh- (approved):
I really like how the `Next()` function looks - very simple to read.

Just a few suggestions for slightly shorter code - not sure if you think some are better or not - feel free to take or discard whatever you want.

:dragon:

pkg/lexer/lexer.go
```diff
@@ -1,5 +1,208 @@
...
+	}
+	if isDigit(l.r) {
+		return tok.SetType(NUM_LIT).SetLiteral(l.readNum())
+	}
+
+	return tok.SetType(ILLEGAL)
+}
+
+func (l *Lexer) readRune() {
+	l.pos += 1
```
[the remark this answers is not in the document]
juliaogris: done.

pkg/lexer/lexer.go
juliaogris: done.

pkg/lexer/lexer.go
juliaogris: i tried this a couple of times with the readString function and gave up. looks goo here.
Done.

pkg/lexer/lexer.go
juliaogris: That's super nice, i like that a lot!

juliaogris: 🏆

pkg/lexer/lexer.go
juliaogris: so good. thanks.

I've updated this. Do you still think this should go before `Next` and all the `readString|Num|Ident|Comment` functions after?

pkg/lexer/lexer.go
camh-: Yeah, I tried readString but it doesn't fit the mould because it needs to look ahead.

pkg/lexer/lexer.go
camh-: Just thought about this a bit more and perhaps (not tested code, just written in the browser):
```
func (l *Lexer) skipWhitespace() {
    l.readWhile(func(r rune) bool { return r == ' ' || r == '\t' || r == '\r' })
```

Now, because skipWhitespace uses readWhile, it peeks rather than requiring `l.r` be read. This means the top of `Next()` can be written as:

````
func (l *Lexer) Next() *Token {
    l.skipWhitespace()
    l.readRune()

    ...
}
```

and I think this means you can now move the helper functions after `Next()`. This is because I found it weird that you would first read a rune, then skip whitespace, so I needed to know what they actually did. In the order of "skip whitespace" then "read rune", it just makes more sense I think - skip all the whitespace up to the next rune.

The only "oddity" now is that skipWhitespace does not skip newlines. Maybe (just maybe) skipWhitespace could be `skipHorizontalWhitespace`?

camh-:
A couple more unattached random comments:

1. You have a couple of rune predicates already. You could take the ones used as anonymous functions in `readComment`, `readNum` and `readIdent` and make them named functions. It may read a bit better:
```
    return l.readWhile(isBeforeEOL)
    ...
    return l.readWhile(isNumRune)
    ....
    return l.readWhile(isIdentRune)
```

2. You could re-use `isBeforeEOL` in `readString`: `if !isBeforeEOL(pr) { break }`

pkg/lexer/lexer.go
juliaogris: ok, thanks - found the same.

PR #12, opened 2022-09-09
github: Fix CI

juliaogris:
Fix CI for `make firebase-public`. The `firebase-public` target depends
on the `tiny` target which runs `tinygo`. Tinygo requires a `go`
installation, in specific the `GOROOT` environment variable. Having go
not explicitly installed has been the cause of recent CI failures.

Add `go-version` target, which runs `go version` to ensure go is
downloaded and installed via hermit. Have targets `tiny` and
`test-tiny` depend on it.


camh- (approved):
:green_apple: lgtm

PR #13, opened 2022-09-09
parser: Implement composite literal parsing

juliaogris:
Implement composite literal - array and map - parsing. Add insertion
order field to map for deterministic map printing and testing. Improve
error messages and type formatting to be more easily read and
understood.

The tests are a bit wanting, there is another PR in the works that improves on
error messages and test coverage and it was a bit too much effort to bring
it in before this one.

### DIFF
--- a/pkg/parser/parser_literal.go
+++ b/pkg/parser/parser_literal.go
@@ -1,0 +1,39 @@
+package parser
+
+import (
+	"strconv"
+
+	"foxygo.at/evy/pkg/lexer"
+)
+
+func (p *Parser) isLiteral() bool {
+	tt := p.cur.TokenType()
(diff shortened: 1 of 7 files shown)


pkg/parser/parser_literal.go
```diff
@@ -21,19 +21,93 @@ func (p *Parser) isLiteral() bool {
...
+	}
+	tt := p.cur.TokenType()
+	p.advance()
+	if tt != lexer.RCURLY {
+		p.appendError("invalid map end: " + tt.FormatDetails())
+		return nil, nil
+	}
+	// type check
+	for _, val := range pairs {
+		if !t.Accepts(val.Type()) {
```
camh-, line 108: Can this be checked inside the main loop where you build up the map?

camh- (approved):
:green_apple:

PR #15, opened 2022-09-12
parser: Add scoped typed analysis to declarations

juliaogris:
Add scoped typed analysis to declarations, ensuring that variables or
parameters do not get redeclared in same block. Variables may be
re-declared in inner scope of a new block with a new type which does not
affect the outer scope. Add detailed tests for various scenarios.

As the parser will also take care of type analysis this is a necessary
addition and patches a whole in the previous global type analysis for
redeclaration.

Add a minor test refactor for parse error assertions


camh- (approved):
:green_heart:

PR #16, opened 2022-09-13
go: Implement return statements

juliaogris:
Implement return statements in ast, parser and evaluator. With this we
can more carefully test valid scoping in the evaluator.

Add pointer to FuncDecl in FuncCall node for easier evaluation.


camh- (approved):
All looks pretty good to me :green_salad:

PR #18, opened 2022-09-14
go: Refactor builtins

juliaogris:
Refactor builtins, such that parser package doesn't have any default
builtins other than for testing. Builtins need to be provided to parser
by caller. This ensures builtins are always only defined in one place.


camh- (approved):
:green_circle:

PR #19, opened 2022-09-14
go: Add assignment evaluation and parsing

juliaogris:
Add assignment parsing and assignment AST node. Extract parseAssignable
from parseTerm to reuse here. This will be extended by indexed and
field selector expressions. Rework String() method for Node type
Var, so that it works better in errors messages.

Add assignment evaluation: assignment nodes only update scope for
existing variables.

### DIFF
--- a/pkg/evaluator/evaluator_test.go
+++ b/pkg/evaluator/evaluator_test.go
@@ -77,6 +77,49 @@ print f2
 	assert.Equal(t, want, b.String())
 }
 
+func TestAssignment(t *testing.T) {
+	prog := `
+f1:num
+f2:num
+f3 := 3
+print f1 f2 f3
+f1 = 1
+print f1 f2 f3
+f1 = f3
+f2 = f1
+f3 = 4
+print f1 f2 f3
+
+`
+	b := bytes.Buffer{}
+	fn := func(s string) { b.WriteString(s) }
+	Run(prog, fn)
+	want := "0 0 3\n1 0 3\n3 3 4\n"
+	assert.Equal(t, want, b.String())
+}
+
+func TestAssignmentAny(t *testing.T) {
+	prog := `
+f1:any
+f2:num
+print f1 f2
+f1 = f2
+print f1 f2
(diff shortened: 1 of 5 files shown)


pkg/parser/parser.go
```diff
@@ -196,9 +196,42 @@ func (p *Parser) parseStatement(scope *scope) Node {
...
+	return &Assignment{Token: tok, Target: target, Value: value}
+}
+
+func (p *Parser) parseAssignable(scope *scope) Node {
+	name := p.cur.Literal
+	p.advance()
+	v, ok := scope.get(name)
+	if !ok {
+		if _, ok := p.funcs[name]; ok {
+			p.appendError("function call must be parenthesized: (" + name + " ...)")
```
camh-, line 228: This error doesn't seem right. Isn't this the case of assigning to a function name and not a variable? Shouldn't this error say something like "cannot assign to <name> as it is a function not a variable"?

I think this is because an assignable (lvalue) is also valid as a term (rvalue), but in each context the error message should be different. On the left-hand side, you cannot assign to a function. On the right-hand side, a function name must be in parens.

You may want to add a test case for this too - assigning to a function.

Or should this just be an "unknown variable" error? Did you decide that a variable can shadow a function? I can't remember.

pkg/evaluator/evaluator_test.go
```diff
@@ -77,6 +77,49 @@ print f2
...
+f1:any
+f2:num
+print f1 f2
+f1 = f2
+print f1 f2
+f1 = fox
+print f1 f2
+
+func fox:string
+       return "🦊"
```
camh-, line 112: :balloon: :pineapple:

camh- (approved):
Just the one error message concern. :evergreen_tree:

pkg/parser/parser.go
juliaogris: yes, thanks for picking this up.
variables cannot shadow functions. unknown variable error seems perfect! I will add the test.

PR #20, opened 2022-09-15
docs: Enforce at least one statement in blocks

juliaogris:
Update grammar to enforce at least one statement in blocks. This means
each function declaration, if statement, while statement etc. need to
have at least one statement in the inner block. Like in Go's "variable
declared but not used" error message we want to catch accidental
omissions and limit the set of "valid programs" to provide more guide
rails.

The parser will be updated in a follow up PR.

While at it, reorder the production rules in the grammar to flow more
naturally from higher order constructs, to single line statements,
expressions and TERMINALS.


camh- (approved):
:green_book:

PR #21, opened 2022-09-15
go: Add if statement parsing and evaluation

juliaogris:
Add if statement parsing and evaluation - this is a pretty straight
forward PR, with only  minor refactors around BlockStatment parsing and
bool constants during evaluation.

### DIFF
--- a/pkg/evaluator/evaluator.go
+++ b/pkg/evaluator/evaluator.go
@@ -46,11 +46,13 @@ func (e *Evaluator) Eval(scope *scope, node parser.Node) Value {
 	case *parser.StringLiteral:
 		return &String{Val: node.Value}
 	case *parser.Bool:
-		return &Bool{Val: node.Value}
+		return e.evalBool(node)
 	case *parser.FunctionCall:
 		return e.evalFunctionCall(scope, node)
 	case *parser.Return:
 		return e.evalReturn(scope, node)
+	case *parser.If:
+		return e.evalIf(scope, node)
 	case *parser.BlockStatement:
 		return e.evalBlockStatment(scope, node)
 	}
@@ -66,6 +68,13 @@ func (e *Evaluator) evalProgram(scope *scope, program *parser.Program) Value {
 		}
 	}
 	return result
+}
+
+func (e *Evaluator) evalBool(b *parser.Bool) Value {
+	if b.Value {
+		return TRUE
+	}
+	return FALSE
 }
 
 func (e *Evaluator) evalDeclaration(scope *scope, decl *parser.Declaration) Value {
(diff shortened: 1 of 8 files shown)


pkg/evaluator/evaluator.go
```diff
@@ -126,6 +135,36 @@ func (e *Evaluator) evalReturn(scope *scope, ret *parser.Return) Value {
 	return &ReturnValue{Val: val}
 }
 
+func (e *Evaluator) evalIf(scope *scope, i *parser.If) Value {
+	val := e.evalCoditionedBlock(scope, i.IfBlock)
```
camh-, line 139: :pineapple: Thinking out aloud here: The AST for `If` could just have a single slice of condition blocks (i.e. no `if` or `else` parts). The `if` bit is just the first of the conditional blocks, and the `else` could just have a condition of `TRUE`. This would reduce the evaluation to just looping over that one slice and evaluating it all the same way.

pkg/evaluator/evaluator.go
```diff
@@ -126,6 +135,36 @@ func (e *Evaluator) evalReturn(scope *scope, ret *parser.Return) Value {
...
+	for _, elseif := range i.ElseIfBlocks {
+		val := e.evalCoditionedBlock(scope, elseif)
+		if val == TRUE || isError(val) {
+			return val
+		}
+	}
+	if i.Else != nil {
+		return e.Eval(newInnerScope(scope), i.Else)
+	}
+	return nil
```
camh-, line 152: This should be `return FALSE` to mirror the case when a branch is executed and `TRUE` is returned. Or the `TRUE` cases should return `nil` too.

pkg/parser/parser.go
```diff
@@ -378,7 +382,7 @@ func (p *Parser) parseTerm(scope *scope) Node {
 		}
 		return lit
 	}
-	p.appendError("unexpected character " + tt.FormatDetails())
+	p.appendError("unexpected " + tt.FormatDetails())
```
camh-, line 385: The tests for this change are in the previous commit. Either move this change there, or those tests here.

pkg/parser/parser.go
```diff
@@ -494,15 +508,27 @@ func (p *Parser) appendErrorForToken(message string, token *lexer.Token) {
...
-	for p.cur.TokenType() != lexer.END && p.cur.TokenType() != lexer.EOF {
+	for !endTokens[p.cur.TokenType()] {
 		stmt := p.parseStatement(scope)
 		if stmt != nil {
 			stmts = append(stmts, stmt)
 		}
 	}
-	p.advancePastNL()
+	if len(stmts) == 0 {
+		p.appendErrorForToken("block statements must have at least one statement, none found", tok)
```
camh-, line 530: I think using the term "block statements" is leaking the grammar into the error messages. This could just say "At least one statement is required jere" or something.

camh- (commented):
some comments to keep you going.

pkg/evaluator/evaluator.go
juliaogris: ok, done.

pkg/parser/ast.go
```diff
@@ -57,6 +57,19 @@ type FuncDecl struct {
...
 }
 
+type If struct {
+	Token        *lexer.Token
+	IfBlock      *ConditionedBlock
+	ElseIfBlocks []*ConditionedBlock
+	Else         *BlockStatement
+}
+
+type ConditionedBlock struct {
```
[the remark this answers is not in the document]
juliaogris: done.

pkg/parser/ast.go
```diff
@@ -174,6 +187,21 @@ func (f *FuncDecl) Type() *Type {
...
 }
 
+func (i *If) String() string {
+	result := "if " + i.IfBlock.String()
+	for _, elseif := range i.ElseIfBlocks {
+		result += "else if" + elseif.String()
+	}
+	if i.Else != nil {
+		result += "else {\n" + i.Else.String() + "}\n"
+	}
```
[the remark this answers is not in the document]
juliaogris: I don't fully follow this, but it seems not to be doing what we want. If I add the else statement I get:
```
--- FAIL: TestBlock (0.00s)
    parser_test.go:173: want != got
        "if (true) {\nprint('TRUE')\n}\n"
        "if (true) {\nprint('TRUE')\n}\n\n"
        ```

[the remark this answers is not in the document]
juliaogris: I actually really don't understand where that newline is coming from 🤷‍♀️

pkg/evaluator/evaluator.go
juliaogris: yeah, as discussed in person - this is a nice and accurate observation and will lead to less code :+1:, but it might be a bit confusing so for now I'll leave as is.

pkg/parser/ast.go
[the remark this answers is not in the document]
camh-: It looks like it is coming from `BlockStatement.String()`

pkg/parser/parser.go
juliaogris: yup, good idea. done.

pkg/parser/parser.go
juliaogris: fixed. thank you!

camh- (approved):
:green_heart: LGTM

PR #22, opened 2022-10-04
Implement while statement parsing and evaluation

juliaogris:
Implement while statement parsing and evaluation by reusing conditional
Block execution from If-statements parsing and evaluation and factor
out shared code.

Fix various bugs that popped up while implementing while statements:

- evaluator: Fix control flow for conditional return statements
- evaluator: Set variable value in correct scope
- parser: Intern basic parser types

### DIFF
--- a/pkg/evaluator/scope.go
+++ b/pkg/evaluator/scope.go
@@ -18,9 +18,19 @@ func (s *scope) get(name string) (Value, bool) {
 		return nil, false
 	}
 	if val, ok := s.values[name]; ok {
-		return val, ok
+		return val, true
 	}
 	return s.outer.get(name)
+}
+
+func (s *scope) getScope(name string) (*scope, bool) {
+	if s == nil {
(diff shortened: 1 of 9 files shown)


pkg/evaluator/scope.go
```diff
@@ -23,6 +23,16 @@ func (s *scope) get(name string) (Value, bool) {
 	return s.outer.get(name)
 }
 
+func (s *scope) getScope(name string) (*scope, bool) {
+	if s == nil {
+		return nil, false
+	}
+	if _, ok := s.values[name]; ok {
+		return s, ok
```
camh-, line 31: nit: I would make this `return s, true`. Since `ok` is always true in this branch, the literal makes it easier to read and understand at a glance.

camh- (approved):
Really nice. I like how simple it is to add the while construct. Confirms that you have a good parsing structure in place. :dragon:

PR #23, opened 2022-10-05
parser: Harden return statements

juliaogris:
Harden return statements such that there is an error when there is
unreachable code after a return statement or a return statement is of
the wrong type.

### DIFF
--- a/pkg/parser/ast.go
+++ b/pkg/parser/ast.go
@@ -13,7 +13,8 @@ type Node interface {
 }
 
 type Program struct {
-	Statements []Node
+	Statements    []Node
+	alwaysReturns bool
 }
 
 type FunctionCall struct {
@@ -86,8 +87,9 @@ type Var struct {
 }
 
 type BlockStatement struct {
-	Token      *lexer.Token // the NL before the first statement
-	Statements []Node
+	Token         *lexer.Token // the NL before the first statement
+	Statements    []Node
+	alwaysReturns bool
 }
 
 type Bool struct {
@@ -123,6 +125,9 @@ func (p *Program) String() string {
 }
 func (*Program) Type() *Type {
 	return NONE_TYPE
+}
+func (p *Program) AlwaysReturns() bool {
+	return p.alwaysReturns
 }
 
 func (f *FunctionCall) String() string {
@@ -162,6 +167,9 @@ func (r *Return) String() string {
 }
 func (r *Return) Type() *Type {
 	return r.T
+}
+func (*Return) AlwaysReturns() bool {
+	return true
 }
 
 func (a *Assignment) String() string {
@@ -205,6 +213,20 @@ func (i *If) String() string {
 func (i *If) Type() *Type {
 	return NONE_TYPE
 }
+func (i *If) AlwaysReturns() bool {
+	if i.Else == nil || !i.Else.AlwaysReturns() {
+		return false
+	}
+	if !i.IfBlock.AlwaysReturns() {
+		return false
+	}
+	for _, b := range i.ElseIfBlocks {
+		if !b.AlwaysReturns() {
+			return false
+		}
+	}
+	return true
+}
 
 func (e *EventHandler) String() string {
 	body := e.Body.String()
@@ -227,6 +249,14 @@ func (b *BlockStatement) String() string {
 func (b *BlockStatement) Type() *Type {
 	return NONE_TYPE
 }
+func (b *BlockStatement) AlwaysReturns() bool {
+	return b.alwaysReturns
+}
+
+func alwaysReturns(n Node) bool {
+	r, ok := n.(interface{ AlwaysReturns() bool })
+	return ok && r.AlwaysReturns()
+}
 
 func (w *While) String() string {
 	return "while " + w.ConditionalBlock.String()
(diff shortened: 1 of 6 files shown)


pkg/parser/ast.go
```diff
@@ -227,6 +250,14 @@ func (b *BlockStatement) String() string {
 func (b *BlockStatement) Type() *Type {
 	return NONE_TYPE
 }
+func (b *BlockStatement) AlwaysReturns() bool {
+	return b.alwaysReturns
+}
+
+func alwaysReturns(n Node) bool {
+	r, ok := n.(Returner)
```
camh-, line 258: :pineapple: If you wanted, you could get rid of the `Returner` interface and change this line to:

    r, ok := n.(interface{ AlwaysReturns() bool})

Essentially using an anonymous interface to check if the given type has this method. I would not normally do this, but since it is only used inside this function, you could get away with it. At the very least, I don't think `Returner` needs to be exported. It is only used internally for parsing correctness.

pkg/parser/ast.go
```diff
@@ -12,8 +12,13 @@ type Node interface {
...
 }
 
+type Returner interface {
+	AlwaysReturns() bool
+}
+
 type Program struct {
-	Statements []Node
+	Statements       []Node
+	guaranteedReturn *lexer.Token
```
camh-, line 21: This field is leftover from the previous implementation.

pkg/parser/parser.go
```diff
@@ -518,18 +527,28 @@ func (p *Parser) parseIfBlock(scope *scope) *BlockStatement {
...
+	var prev Node
 	for !endTokens[p.cur.TokenType()] {
+		tok := p.cur
 		stmt := p.parseStatement(scope)
-		if stmt != nil {
-			stmts = append(stmts, stmt)
+		if stmt == nil {
+			continue
+		}
+		if alwaysReturns(prev) {
```
camh-, line 538: We don't need `prev` any more. This can be `if block.AlwaysReturns()` (or `if block.alwaysReturns`) since we set that on the previous iteration.

pkg/parser/parser.go
```diff
@@ -87,6 +87,7 @@ func (p *Parser) Parse() *Program {
 func (p *Parser) parseProgram(scope *scope) *Program {
 	program := &Program{Statements: []Node{}}
 	p.advanceTo(0)
+	returnSeen := false
```
camh-, line 90: rather than this `returnSeen` local var, you could add `hasReturn` to the `Program` struct, and also implement `HasReturn()` on it. Then you can set it and check it in a similar way to what I suggested on `parseBlockWithEndTokens()`.

pkg/parser/parser.go
```diff
@@ -96,7 +97,15 @@ func (p *Parser) parseProgram(scope *scope) *Program {
 		case lexer.ON:
 			stmt = p.parseEventHandler(scope)
 		default:
+			tok := p.cur
 			stmt = p.parseStatement(scope)
+			if stmt != nil && returnSeen {
```
camh-, line 102: `if stmt != nil && p.HasReturn() {`

pkg/parser/parser.go
```diff
@@ -96,7 +97,15 @@ func (p *Parser) parseProgram(scope *scope) *Program {
...
+				p.appendErrorForToken("unreachable code", tok)
+				stmt = nil
+			}
+			if alwaysReturns(stmt) {
+				returnSeen = true
```
camh-, line 107: `p.hasReturn = true`

pkg/parser/parser_test.go
```diff
@@ -189,28 +189,44 @@ x=len('123')
...
 	end
 end
+func nums2:num n1:num n2:num
+	if true //TODO: > 10
+		return n1
+	else
+		return n2
+	end
+end
+func nums3:num
```
camh-, line 212: is this function for later? because it should be invalid since it does not return a value at the end, and you do not have that constant condition checking we've talked about. Right now, I would expect this to generate an error and only later when  you add that constant checking would it pass.

Do you have any checking for missing returns on functions yet?

camh- (commented):
some stuff to get you going, but i really like how this looks. again, reasonably simple to add means its in pretty good shape

pkg/parser/ast.go
camh-: oh, looks like the linter beat me to it

juliaogris: oups, sorry. fixed.

juliaogris: I should turn up the linter strictness.
Will do in follow up pr

pkg/parser/parser.go
juliaogris: yes, nice! I will make alwaysReturns private....

pkg/parser/parser_test.go
juliaogris: yeah, I don't have checks for missing returns yet. I should probably add this to this PR in a last commit.

juliaogris: yeah, I don't have checks for missing returns yet. I should probably add this to this PR in a last commit.

pkg/parser/parser.go
juliaogris: ok done (as `program.alwaysReturns = true`)

pkg/parser/parser.go
juliaogris: done as `program.AlwaysReturns()` - like block

pkg/parser/parser.go
juliaogris: ok, great idea. done.

camh- (approved):
:dragon_face: bloody beautiful

PR #24, opened 2022-10-05
parser: Add unused var check

juliaogris:
Add unused var check reporting an error if declared variable is never
used. This feature was yet again borrowed from go and piggy backed on
the scope data structure.

Refactor `validateVar` to `validateVarDecl` as it seems the more
appropriate name.

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -115,6 +115,7 @@ func (p *Parser) parseProgram(scope *scope) *Program {
 			program.Statements = append(program.Statements, stmt)
 		}
 	}
+	p.validateScope(scope)
 	return program
 }
 
@@ -222,7 +223,6 @@ func (p *Parser) parseAssignmentStatement(scope *scope) Node {
 		p.advancePastNL()
 		return nil
 	}
-
 	target := p.parseAssignable(scope)
 	tok := p.cur
 	if target == nil {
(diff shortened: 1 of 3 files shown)


juliaogris:
Hi from GitHub CLI

pkg/parser/parser_test.go
```diff
@@ -401,6 +422,103 @@ end
...
+	print "foo"
+else if true
+	x := 1
+end
+print x
+`: "line 6 column 2: 'x' declared but not used",
+		`
+x :=1
+for i := range 10
+	x :=2
```
camh-, line 474: nit: missing space before `2` - and the next few times.

pkg/parser/parser_test.go
```diff
@@ -401,6 +422,103 @@ end
...
+x := 1
+if true
+	print "foo"
+else if true
+	x := 1
+end
+print x
+`: "line 6 column 2: 'x' declared but not used",
+		`
+x :=1
```
camh-, line 472: nit: missing space before `1` - and the next few times

pkg/parser/parser.go
```diff
@@ -247,8 +247,8 @@ func (p *Parser) parseAssignmentStatement(scope *scope) Node {
 
 func (p *Parser) parseAssignable(scope *scope) Node {
 	name := p.cur.Literal
-	p.advance()
 	v, ok := scope.get(name)
+	p.advance()
```
camh-, line 251: did this line move for a reason? it splits up the assignment of `ok` and the test of it.

camh- (commented):
LGTM. just a couple of nits but otherwise nice how short it is.

pkg/parser/parser_test.go
juliaogris: thanks done. i need an evy formatter.

pkg/parser/parser.go
juliaogris: yeah, the thought was that the error message is linked to the position after the variable, rather on it. but i didn't fix it properly, i'll reworked it again in a separate commit.

juliaogris: ok, done.

PR #26, opened 2022-10-06
github: Replace firebase-deployment action

juliaogris:
Replace firebase-deployment action by hand-crafted make targets. This
has been motivated by an apparently broken firebase-tools release
(4 hours old) that cannot be down-graded when using the firebase GitHub
action and is a blocker. Aside: the foxygoat authors believe it is nice
to stay third-party GitHub action free where possible so that the bulk
of what's run on CI can also be run locally.

While at it upgrade the firebase CLI version to the second most recent
one - firebase-11.13.0 works, but the previous firebase-11.6.0 and the
even more recent firebase-11.14.0 do not it seems.


camh- (approved):
:green_apple: very nice

PR #29, opened 2022-10-07
github: Fix CI setup for master build

juliaogris:
Fix CI setup for master build, such that PR deploy is only executed on
CI run triggered by PR events. Master/prod deployment is executed as
part of the release job.
This is a bug as a result of the recent firebase deployment refactor.

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -14,6 +14,7 @@ jobs:
         fetch-depth: 0
     - run: ./bin/make ci
     - run: ./bin/make firebase-deploy
+      if: ${{ github.event_name == 'pull_request' }} # only run on PR, deploy prod/master in release job
       env:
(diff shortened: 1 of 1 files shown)


camh- (approved):

.github/workflows/cicd.yaml
```diff
@@ -14,6 +14,7 @@ jobs:
         fetch-depth: 0
     - run: ./bin/make ci
     - run: ./bin/make firebase-deploy
+      if: github.event_name == 'pull_request' # only run on PR, deploy prod/master in release job
```
camh-, line 17: oops shouldn't have approved. needs `${{ ... }}` I think

juliaogris: ok, I will add it down below too ( it doesn't really need it it seems.)

juliaogris: ok, I will add it down below too ( it doesn't really need it it seems.)

PR #30, opened 2022-10-07
go: Implement break statement

juliaogris:
Implement break statement in parser and evaluator ensuring it is only
used inside a loop. Use scope to keep track of enclosing AST node and
ensure that break's scope is within a loop (while) node's scope.
Implement unreachable code with block.AlwaysTerminates method on block.

Fix bare returns in evaluator as they accidentally got ignored and
didn't stop execution. This became apparent when thinking about breaks
in evaluator.

Also, fix AlwaysReturns method on While blocks. Loops may at times never
have their body executed so can _never_ always return.

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -104,12 +104,12 @@ func (p *Parser) parseProgram() *Program {
 		default:
 			tok := p.cur
 			stmt = p.parseStatement(scope)
-			if stmt != nil && program.AlwaysReturns() {
+			if stmt != nil && program.AlwaysTerminates() {
 				p.appendErrorForToken("unreachable code", tok)
 				stmt = nil
 			}
-			if alwaysReturns(stmt) {
-				program.alwaysReturns = true
+			if alwaysTerminates(stmt) {
+				program.alwaysTerminates = true
 			}
 		}
 		if stmt != nil {
@@ -138,7 +138,7 @@ func (p *Parser) parseFunc(scope *scope) Node {
 		p.appendError("redeclaration of function '" + funcName + "'")
 		return nil
 	}
-	if fd.ReturnType != NONE_TYPE && !block.AlwaysReturns() {
+	if fd.ReturnType != NONE_TYPE && !block.AlwaysTerminates() {
 		p.appendError("missing return")
 	}
 	p.assertEnd()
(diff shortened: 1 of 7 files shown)


pkg/parser/parser.go
```diff
@@ -647,52 +656,56 @@ func (p *Parser) parseForStatement(scope *scope) Node {
...
 }
 
-func (p *Parser) parseIfConditionalBlock(scope *scope) *ConditionalBlock {
-	tok := p.cur
+func (p *Parser) parseIfConditionalBlock(ifStmt *If, scope *scope) *ConditionalBlock {
+	result := &ConditionalBlock{Token: p.cur}
 	p.advance() // advance past IF token
-	scope = newInnerScope(scope)
-	condition := p.parseCondition(scope)
+	scope = scope.innerScope(ifStmt)
```
camh-, line 704: not sure this is right. The block should be the conditional block, not the if block. the if has multiple blocks and they do not share the scope. they each have their own.

pkg/parser/parser.go
```diff
@@ -647,52 +656,56 @@ func (p *Parser) parseForStatement(scope *scope) Node {
...
+	result.Condition = p.parseCondition(scope)
```
camh-, line 705: the scope of the condition should be the outer scope, not the inner one. it cannot refer to variables declared inside the block. it may not technically matter, because at this point the inner scope is empty and the condition cannot declare vars (although if it could, should they be in the scope of the block or half-way?an extra hidden middle scope? :pineapple: on the question)

camh- (commented):
some comments for now

pkg/parser/parser.go
juliaogris: a new scope is created, each time, just always containing the same outer block == ifStmt.
It is confusing I can see this, but otherwise it doesn't line up with the else statement.

I cannot have the block create its own scope, because in the case of the While `.block` has to be the While statement (I suppose I could, but again unnecessary extra scoping).

I've updated this now so that the conditional block creates a new block (which means there is an unnecessary extra scope for the while, but I guess that's ok).

pkg/parser/parser.go
juliaogris: ok, fixed. thanks.

juliaogris: hmm, I thought about this some more:
A statement has never access to the variables declared "later" in the same block - so in that sense there is no difference between the scope of the condition and the scope of the following block.

However, for a `for` loop the "condition" is what is going on inside the `for i := range 10` statement. i will be available inside the loop but not outside the loop, so from my perspective i is defined in the same block as the loop.

pkg/parser/parser.go
juliaogris: I still feel a bit wrong about this because of the asymmetry between the else branch and the if | if-else branches.

camh- (approved):
Quite the marathon, but this now LGTM! :tada: 

:green_circle:

PR #31, opened 2022-10-11
docs: Update docs for whitespace and inverted types

juliaogris:
Update docs for whitespace. While at it, remove a bunch of duplicated
sections in grammar. We are now making horizontal whitespace `WS`
significant in certain places which allows us to remove the somewhat
awkward array and map typing in favour of derived and inferred types
for literals.

Remove `term` from grammar as it is not used in code, we only work with
expression. Introduce `<- ->` and  `<+ +>` tokens to annotated
whitespace inside grammar: expressions in expression lists, such as
function calls or array elements may not contain whitespace, as
whitespace is used as list separator.

Invert types so that types read as variables are being access:

        m:[]{}string
        m[0].name = "Joe"

Also add note on pass by reference for maps and arrays and pass by value
for string, num and bool.

---
For easier review:
https://github.com/foxygoat/evy/blob/docs-ws/docs/syntax_by_example.md
https://github.com/foxygoat/evy/blob/docs-ws/docs/syntax_grammar.md

### DIFF
--- a/docs/syntax_by_example.md
+++ b/docs/syntax_by_example.md
@@ -1,10 +1,9 @@
 # Evy Syntax by Example
 
-The following examples on various aspects of `evy`'s syntax give an
-intuitive understanding of the language. For a formal language
-specification see the [syntax grammar].
+The following examples give an intuitive understanding of various
+aspects of `evy`'s syntax. For a formal language specification see the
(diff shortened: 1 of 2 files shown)


camh- (approved):
🪴

PR #32, opened 2022-10-11
parser: Implement whitespace sensitivity

juliaogris:
Update lexer to emit single WS token for horizontal whitespace.

Implement whitespace sensitivity in parser: 

1. Disallow whitespace after unary operator 
2. Disallow whitespace within the expressions of a list:
  - argument list of function call
  - elements of an array literal
  - values of map literal 
3. Regardless of context, allow whitespace within (), [] and {}

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -195,6 +195,9 @@ func (p *Parser) parseStatement(scope *scope) Node {
 	case lexer.NL, lexer.EOF, lexer.COMMENT:
 		p.advancePastNL()
 		return nil
+	case lexer.WS:
+		p.advance()
+		return nil
 	case lexer.IDENT:
 		switch p.peek.Type {
 		case lexer.ASSIGN, lexer.DOT:
@@ -519,6 +522,17 @@ func (p *Parser) parseBlockWithEndTokens(scope *scope, endTokens map[lexer.Token
 }
 
 func (p *Parser) advance() {
+	p.advanceWSS()
+	if p.cur.Type == lexer.WS {
+		p.advanceWSS()
+	}
+	if p.peek.Type == lexer.WS {
+		p.peek = p.lookAt(p.pos + 2)
+	}
+}
+
+// advanceWSS advances to the next token in whitespace sensitive (wss) manner.
+func (p *Parser) advanceWSS() {
 	p.pos++
 	p.cur = p.lookAt(p.pos)
 	p.peek = p.lookAt(p.pos + 1)
@@ -528,6 +542,9 @@ func (p *Parser) advanceTo(pos int) {
 	p.pos = pos
 	p.cur = p.lookAt(pos)
 	p.peek = p.lookAt(pos + 1)
+	if p.peek.Type == lexer.WS {
+		p.peek = p.lookAt(p.pos + 2)
+	}
 }
 
 func (p *Parser) lookAt(pos int) *lexer.Token {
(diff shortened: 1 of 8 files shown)


pkg/parser/parser.go
```diff
@@ -567,6 +570,17 @@ func (p *Parser) parseBlockWithEndTokens(scope *scope, endTokens map[lexer.Token
 }
 
 func (p *Parser) advance() {
+	p.advanceOne()
+	for p.cur.Type == lexer.WS {
+		p.advanceOne()
+	}
+	inc := 2
+	for p.peek.Type == lexer.WS {
```
juliaogris, line 578: I hope there is a better way for doing this than these two loops :(

pkg/parser/parser.go
```diff
@@ -567,6 +570,17 @@ func (p *Parser) parseBlockWithEndTokens(scope *scope, endTokens map[lexer.Token
 }
 
 func (p *Parser) advance() {
```
juliaogris, line 572: This function should now probably be renamed, but to what?
`advanceSkippingWS()` that's a mouth full...

pkg/parser/parser.go
```diff
@@ -567,6 +570,17 @@ func (p *Parser) parseBlockWithEndTokens(scope *scope, endTokens map[lexer.Token
...
+	for p.cur.Type == lexer.WS {
+		p.advanceOne()
+	}
+	inc := 2
+	for p.peek.Type == lexer.WS {
+		p.peek = p.lookAt(p.pos + inc)
+	}
+}
+
+func (p *Parser) advanceOne() {
```
juliaogris, line 583: Needs a better name.

pkg/parser/parser.go
```diff
@@ -576,6 +590,10 @@ func (p *Parser) advanceTo(pos int) {
 	p.pos = pos
 	p.cur = p.lookAt(pos)
 	p.peek = p.lookAt(pos + 1)
+	inc := 2
+	for p.peek.Type == lexer.WS {
```
juliaogris, line 594: yuck.

camh-: Is this yuck because you don't advance `inc` so the loop never terminates if there are two consecutive whitespace tokens? I know there can't be because they are folded, but then why the loop?

pkg/parser/parser.go
camh-: if `advanceOne` is already peeking, why do you need this here. :thinking:

camh-: oh, sorry, github confused me with the diff. it is advanceTo() that has the same loop. ignore me.

pkg/parser/parser.go
juliaogris: yeah, I thought about this, but advancePastNL means: keep on advancing until you hit NL and then one more. In this case we want to _only_ advance if there is WS....

pkg/parser/parser.go
juliaogris: yeah you are right. WS is collapsed into a single token so I should just advance one more if it is WS. I will update this.

juliaogris: ok, I replaced the loops with if's. That feels a little less yuck I guess.

I don't think I can just assign `cur = peek` as I also need to know the position (of peek), I could track the peekPos, but not sure if that makes things easier...

pkg/parser/parser.go
juliaogris: yeah, but I think the WS insensitive is the more common one so I'd prefer to reserve `advance` for it i think.... to be seen.

camh- (approved):
:green_heart:

camh-:
sorry, should have done :white_heart:

PR #33, opened 2022-10-12
parser: Implement Pratt Expression parser

juliaogris:
Implement Top Down Operator Precedence or Pratt parser, replacing the
existing but very limited parseExpression method.

This parser is based on Thorston Ball's Interpreter book and his monkey
interpreter.

Functional changes include some error messages and the removal of type
prefix for array and map literals. The type is now inferred:

	arr := [1] // num array; arr := num[1] is now invalid

All arithmetic and logical expressions are now parsed. Outstanding work
includes array slices, expression parsing for the left hand side of an
assignment, e.g.

	arr[0] = 2

The evaluator implementation is also not included in this commit as it
has become fairly large as is.

Documentation updates are in flight in parallel at PR 31

Link: https://github.com/juliaogris/monkey/blob/master/parser/parser.go


pkg/parser/parser.go
```diff
@@ -199,6 +207,9 @@ func (p *Parser) parseStatement(scope *scope) Node {
 		if p.isFuncCall(p.cur) {
 			return p.parseFunCallStatement(scope)
 		}
+		if p.peek.Type == lexer.LBRACKET || p.peek.Type == lexer.DOT {
```
juliaogris, line 210: I had to move this down, otherwise `len [1]` was interpreted as ASSIGNMENT in errored. I just moved the p.isFuncCall line before assignment I got less useful error messages for `print = 5` etc.

pkg/parser/expression.go
```diff
@@ -0,0 +1,402 @@
...
+			p.appendError(leftType.String() + " index expects num, found " + index.Type().Format())
+			return nil
+		}
+		return &IndexExpression{Token: tok, Left: left, Index: index, T: left.Type().Sub}
+	}
+	if index.Type() != STRING_TYPE {
+		p.appendError("map index expects string, found " + index.Type().Format())
+		return nil
+	}
+	return &IndexExpression{Token: tok, Left: left, Index: index}
```
camh-, line 162: should this be setting `T` like you do for the array/string indexed expression? If so, I'd reorder the above code to:
```
if (leftType == ARRAY || leftType == STRING) && index.Type() != NUM_TYPE {
    return error...
}
if leftType == MAP && index.Type() != STRING_TYPE {
    return error...
}
return &IndexExpression{Token: tok, Left: left, Index: index, T: left.Type().Sub}
```

Happy path and all, but I also find that error checking simpler to read - it spells out exactly what is required in one line each (i.e. array and string need a numeric index, map needs a string index)

pkg/parser/expression.go
```diff
@@ -0,0 +1,402 @@
...
+}
+
+func isComparisonOp(tt lexer.TokenType) bool {
+	return tt == lexer.EQ || tt == lexer.NOT_EQ || tt == lexer.LT || tt == lexer.GT || tt == lexer.LTEQ || tt == lexer.GTEQ
+}
+
+func (p *Parser) validateUnaryType(unaryExp *UnaryExpression) {
+	tok := unaryExp.Token
+	op := unaryExp.Op
+	if op != OP_MINUS && op != OP_BANG {
```
camh-, line 193: how about:
```
switch unaryExp.Op {
case OP_MINUS:
    if unaryExp.Right.Type() != NUM_TYPE {
        p.appendError...
    }
case OP_BANG:
    if unaryExp.Right.Type() != BOOL_TYPE {
        p.appendError...
    }
default:
    p.appendError... // invalid unary token
}
```

seems more straightforward to me.

pkg/parser/expression.go
```diff
@@ -0,0 +1,402 @@
...
+		return
+	}
+
+	leftType := binaryExp.Left.Type()
+	rightType := binaryExp.Right.Type()
+	if !leftType.Equals(rightType) {
+		p.appendErrorForToken("mismatched type for "+op.String()+": "+leftType.Format()+", "+rightType.Format(), tok)
+		return
+	}
+	if op == OP_PLUS {
```
camh-, line 223: i think i'd use a switch again - then you don't need the `return` statements. it would look a lot like the unary validation

pkg/parser/expression.go
```diff
@@ -0,0 +1,402 @@
...
+	t := &Type{Name: MAP, Sub: p.combineTypes(types)}
+	return &MapLiteral{Token: tok, Pairs: pairs, Order: order, T: t}
+}
+
+func (p *Parser) parseMapPairs(scope *scope) (map[string]Node, []string) {
+	pairs := map[string]Node{}
+	var order []string
+	tt := p.cur.TokenType()
+
+	for !p.isAtEOL() && tt != lexer.RCURLY {
```
camh-, line 360: i thought a map literal could span multiple lines, in which case it seems wrong to terminate parsing at EOL?

camh- (commented):
some comments for now. Leaving the screen for the evening so posting what I have so far.

pkg/parser/parser.go
juliaogris: I did not. Will move it back up.

pkg/parser/ast.go
```diff
@@ -22,13 +22,34 @@ type FunctionCall struct {
...
-	Value Node
+type UnaryExpression struct {
+	Token *lexer.Token // The unary token, e.g. !
+	Op    Operator
+	Right Node
+}
+
+type BinaryExpression struct {
+	T     *Type
+	Token *lexer.Token // The unary token, e.g. !
```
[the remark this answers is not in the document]
juliaogris: done.

pkg/parser/expression.go
```diff
@@ -0,0 +1,402 @@
...
+	}
+	if op == OP_AND || op == OP_OR {
+		if leftType != BOOL_TYPE {
+			p.appendErrorForToken("'"+op.String()+"' takes bool type, found "+leftType.Format(), tok)
+		}
+		return
+	}
+}
+
+func (p *Parser) curPrecedence() precedence {
```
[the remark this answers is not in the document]
juliaogris: ok, done.

pkg/parser/expression.go
```diff
@@ -0,0 +1,402 @@
...
+	case lexer.STRING_LIT, lexer.NUM_LIT, lexer.TRUE, lexer.FALSE, lexer.LBRACKET, lexer.LCURLY:
+		left = p.parseLiteral(scope)
+	case lexer.BANG, lexer.MINUS:
+		left = p.parseUnaryExpr(scope)
+	case lexer.LPAREN:
+		left = p.parseGroupedExpr(scope)
+	default:
+		p.appendError("unexpected " + p.cur.FormatDetails())
+	}
+	for left != nil && !p.isAtExprEnd() && prec < p.curPrecedence() {
```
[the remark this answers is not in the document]
juliaogris: done.

pkg/parser/expression.go
```diff
@@ -0,0 +1,402 @@
...
+}
+
+func (p *Parser) parseBinaryExpr(scope *scope, left Node) Node {
+	tok := p.cur
+	expType := left.Type()
+	if isComparisonOp(tok.Type) {
+		expType = BOOL_TYPE
+	}
+	binaryExp := &BinaryExpression{Token: tok, T: expType, Op: op(tok), Left: left}
+	prec := p.curPrecedence()
```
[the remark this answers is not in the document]
juliaogris: done.

pkg/parser/expression.go
juliaogris: ok, agreed that's nicer, done.

pkg/parser/expression.go
juliaogris: ok, that's an outright bug - sorry and thanks for finding. I need to do more type checking but was lazily assuming the evaluation stage will shake all of this out...
fixed

pkg/parser/expression.go
juliaogris: yes - I haven't implemented this yet. I think this needs more thought, especially for array literals:  what if they part of an expression?
I'll circle back on this later if that's ok?

camh-: sure, that's ok. i suspected this was the case.

pkg/parser/expression.go
juliaogris: yeah agreed, looks a lot better.

camh- (approved):
:leafy_green:

PR #34, opened 2022-10-16
parser: Invert composite type order

juliaogris:
Invert composite type order, such that an array of number is now
[]num. This lines up with how Go defines slices / arrays and the order in
which elements are accessed. An array of maps of string values is now:

    a:[]{}string
    a[0].name = "Joe"

As part of this refactor remove parser_type.go file completely as it has
been inconvenient when looking for parser_test.go.

Only the last commit of this PR is relevant. It requires PR 33 to be
merged first.


camh- (approved):
:green_apple:

PR #35, opened 2022-10-16
parser: Update parseAssignmentStatment for index and dot

juliaogris:
Update parseAssignmentStatment for index and dot expression as assignment
target.


camh- (approved):
:green_book: LGTM

PR #37, opened 2022-10-17
evaluator: Add expression evaluation for basic types

juliaogris:
Add expression evaluation for basic types, including arithmetic operations,
logical operations, comparison operations, string concatenations, unary
operators and grouping with (). Still outstanding are all array and map
operations.

### DIFF
--- a/pkg/evaluator/evaluator.go
+++ b/pkg/evaluator/evaluator.go
@@ -59,6 +59,10 @@ func (e *Evaluator) Eval(scope *scope, node parser.Node) Value {
 		return e.evalWhile(scope, node)
 	case *parser.BlockStatement:
 		return e.evalBlockStatment(scope, node)
+	case *parser.UnaryExpression:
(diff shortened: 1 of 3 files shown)


pkg/evaluator/evaluator.go
```diff
@@ -214,3 +218,97 @@ func (e *Evaluator) evalExprList(scope *scope, terms []parser.Node) []Value {
...
+	}
+	op := expr.Op
+	switch right := right.(type) {
+	case *Num:
+		if op == parser.OP_MINUS {
+			return &Num{Val: -right.Val}
+		}
+	case *Bool:
+		if op == parser.OP_BANG {
+			return &Bool{Val: !right.Val}
```
camh-, line 235: should this use `boolVal()` like the comparison operators? Otherwise this creates a new Bool value and is not one of `TRUE` or `FALSE`

camh- (approved):
LGTM :green_salad:

PR #40, opened 2022-10-18
evaluator: Array concatenation

juliaogris:
Add array concatenation for `+` operator. Ensure a copy of both operands is
made.

Fix parse error for array concatenation with empty arrays.

---

Only the last two commits of this PR are relevant. The rest depends on merging PRs 36-39.


camh- (approved):
:green_heart:

PR #41, opened 2022-10-18
go: Add slice expression

juliaogris:
Add slice parsing and evaluation for array and string. Ensure copy is made of array
elements when slicing. Dis-allow string indexing on left side of assignment:

    s := "abc"
    s[0] = "ABC" // doesn't make sense.


camh- (approved):
:leafy_green:

PR #42, opened 2022-10-20
eval: Fix any's wrapping any's

juliaogris:
Fix `any`'s wrapping `any`'s so that `any` only ever contains a
non-`any` type. We have added two test cases that failed before the code fix.

Sneak in a `make ci` dependency reorder so that linter errors surface earlier.


camh- (approved):
:green_salad:

PR #43, opened 2022-10-23
Add for ... range loop parsing and evaluation

juliaogris:
Add `for` statement parsing according to grammar.

    for_stmt   = "for" ident := "range" expr [ expr [expr]]
                    statements
                 "end" NL .

Range Expression can be an expression evaluating to a string, a map or
an array or it can consist of 1 to 3 numbers forming a StepRange.

Add 2 new AST nodes: `For` and `StepRange`. Fix up previous for loop tests
that now fail because of unused variable errors in parser package.

Add `for ... range` loop evaluation using the new AST nodes `for` and `stepRange`.
In order to properly capture the loopVariable, e.g. `x` in `for
x := range [1 2 3]` introduce the creation of a zeroValue from
a *parser.Type in evaluator.

### DIFF
--- a/pkg/evaluator/evaluator.go
+++ b/pkg/evaluator/evaluator.go
@@ -61,6 +61,8 @@ func (e *Evaluator) Eval(scope *scope, node parser.Node) Value {
 		return e.evalIf(scope, node)
 	case *parser.While:
 		return e.evalWhile(scope, node)
+	case *parser.For:
+		return e.evalFor(scope, node)
 	case *parser.BlockStatement:
 		return e.evalBlockStatment(scope, node)
 	case *parser.UnaryExpression:
@@ -74,7 +76,7 @@ func (e *Evaluator) Eval(scope *scope, node parser.Node) Value {
 	case *parser.DotExpression:
 		return e.evalDotExpr(scope, node)
 	}
-	return nil
+	return nil // TODO: panic?
 }
 
 func (e *Evaluator) evalProgram(scope *scope, program *parser.Program) Value {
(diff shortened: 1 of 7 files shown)


pkg/evaluator/evaluator.go
```diff
@@ -213,6 +215,94 @@ func (e *Evaluator) evalWhile(scope *scope, w *parser.While) Value {
...
+func (e *Evaluator) newStepRange(scope *scope, r *parser.StepRange, loopVar *parser.Var) (ranger, Value) {
+	start, err := e.numValWithDefault(scope, r.Start, 0.0)
+	if err != nil {
+		return nil, err
+	}
+	stop, err := e.numVal(scope, r.Stop)
+	if err != nil {
+		return nil, err
+	}
+	step, err := e.numValWithDefault(scope, r.Step, 1.0)
```
camh-, line 271: hmmm, the second return value is of type `Value` not `error`, so calling it `err` is a little misleading. I know it is the error case, but still... Perhaps call it errValue or something?

pkg/evaluator/evaluator.go
```diff
@@ -213,6 +215,94 @@ func (e *Evaluator) evalWhile(scope *scope, w *parser.While) Value {
...
+	if err != nil {
```
camh-, line 272: should it also be an eval error if `step` is zero? or is that just an infinite loop up to the programmer to deal with?

pkg/evaluator/ranger.go
```diff
@@ -0,0 +1,78 @@
...
+	return true
+}
+
+func (a *arrayRange) next() bool {
+	elements := *a.array.Elements
+	if a.cur >= len(elements) {
+		return false
+	}
+	a.loopVar.Set(elements[a.cur])
+	a.cur += 1
```
camh-, line 52: Could be `a.cur++`. I sometimes wonder why Go has `++` - may as well do this `+= 1` since you can't use `++` in an expression. Up to you.

pkg/evaluator/evaluator.go
```diff
@@ -213,6 +215,94 @@ func (e *Evaluator) evalWhile(scope *scope, w *parser.While) Value {
...
+		return &arrayRange{loopVar: loopVar, array: v, cur: 0}, nil
+	case *String:
+		loopVar := &String{}
+		scope.set(f.LoopVar.Name, loopVar)
+		return &stringRange{loopVar: loopVar, str: v, cur: 0}, nil
+	case *Map:
+		loopVar := &String{}
+		scope.set(f.LoopVar.Name, loopVar)
+		order := make([]string, len(*v.Order))
+		copy(order, *v.Order)
```
camh-, line 255: Don't forget to document this - that iteration over a map only iterates over the entries present at the start, and not new ones added. And that it is safe to delete entries.

pkg/parser/parser.go
```diff
@@ -587,14 +587,75 @@ func (p *Parser) parseBreakStatement(scope *scope) Node {
...
+		p.assertEOL()
+	case ARRAY:
+		forNode.LoopVar.T = t.Sub
+		forNode.Range = n
+		p.assertEOL()
+	case NUM:
+		forNode.LoopVar.T = NUM_TYPE
+		forNode.Range = p.parseStepRange(scope, n)
+	default:
+		p.appendError("expected num, string, array or map after range, found " + t.Format())
```
camh-, line 630: On this branch, `forNode.LoopVar.T` is nil (the loopVar does not have a type). However, you go on to parse the block, and the code in that block can reference the var. That reference ends up failing because `forNode.LoopVar` (as set in the scope) is not a proper var, and you get a nil pointer dereference when trying to determine if the type is compatible when used.

Perhaps you could set the type to `any` for the purposes of continuing to parse the code? You may generate unrelated errors due to the type, but you won't crash.

camh- (approved):
Just some trivial things and a bug. :green_apple:

pkg/evaluator/evaluator.go
juliaogris: fair, done.

pkg/evaluator/evaluator.go
juliaogris: hmmm... i did think about it and decided to go for infinite loop. maybe i shouldn't.

juliaogris: ok added.

pkg/evaluator/ranger.go
juliaogris: ok, updated.

pkg/evaluator/evaluator.go
juliaogris: yeah... there were some docs updates, but there is already a doc update in flight. should get that merged...

juliaogris: yeah... there were some docs updates, but there is already a doc update in flight. should get that merged maybe and rework it again ;)

pkg/parser/parser.go
juliaogris: ok, I've set it to NONE_TYPE on generation. hope that will do.

pkg/evaluator/evaluator.go
juliaogris: I've added a note to my longer term todos here:
https://gist.github.com/juliaogris/05d459cbdb475c53cced8c06dde71072

pkg/parser/parser.go
camh-: yeah that would be fine I expect. As long as it doesn't crash - it still doesn't compile - that error is correctly detected.

PR #45, opened 2022-10-24
evaluator: Add map builtins `has` and `del`

juliaogris:
Add builtin function `del` for entry removal from map:

    del {a: 1} "a" // {}

Add builtin function `has` for entry existence check in map:

    has {a: 1} "a" // true

### DIFF
--- a/pkg/evaluator/builtin.go
+++ b/pkg/evaluator/builtin.go
@@ -34,6 +34,7 @@ func DefaultBuiltins(printFn func(string)) Builtins {
 		"move":  {Func: moveFunc(printFn), Decl: moveDecl},
 		"line":  {Func: lineFunc(printFn), Decl: lineDecl},
 		"has":   {Func: BuiltinFunc(hasFunc), Decl: hasDecl},
+		"del":   {Func: BuiltinFunc(delFunc), Decl: delDecl},
 	}
 }
 
@@ -89,6 +90,22 @@ func hasFunc(args []Value) Value {
 	key := args[1].(*String)
 	_, ok := m.Pairs[key.Val]
 	return &Bool{Val: ok}
+}
+
+var delDecl = &parser.FuncDecl{
+	Name: "del",
+	Params: []*parser.Var{
+		{Name: "m", T: parser.GENERIC_MAP},
+		{Name: "key", T: parser.STRING_TYPE},
+	},
+	ReturnType: parser.NONE_TYPE,
+}
+
+func delFunc(args []Value) Value {
+	m := args[0].(*Map)
+	keyStr := args[1].(*String)
+	m.Delete(keyStr.Val)
+	return nil
 }
 
 var moveDecl = &parser.FuncDecl{
--- a/pkg/evaluator/builtin.go
+++ b/pkg/evaluator/builtin.go
@@ -33,6 +33,7 @@ func DefaultBuiltins(printFn func(string)) Builtins {
 		"len":   {Func: BuiltinFunc(lenFunc), Decl: lenDecl},
 		"move":  {Func: moveFunc(printFn), Decl: moveDecl},
 		"line":  {Func: lineFunc(printFn), Decl: lineDecl},
+		"has":   {Func: BuiltinFunc(hasFunc), Decl: hasDecl},
 	}
 }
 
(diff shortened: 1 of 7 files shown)


pkg/evaluator/builtin.go
```diff
@@ -74,6 +75,25 @@ func lenFunc(args []Value) Value {
 	return newError("'len' takes 1 argument of type 'string', array '[]' or map '{}' not " + args[0].Type().String())
 }
 
+var hasDecl = &parser.FuncDecl{
+	Name: "has",
+	Params: []*parser.Var{
+		{Name: "map", T: parser.ANY_TYPE},
```
camh-, line 82: not sure i understand why this is `ANY_TYPE`. Is it because it is a generic function in a language without generics? If so, perhaps needs a comment, or maybe you could somehow make use of `NONE_TYPE` like you do for empty arrays and maps?

pkg/evaluator/builtin.go
```diff
@@ -74,6 +75,25 @@ func lenFunc(args []Value) Value {
...
+		{Name: "key", T: parser.STRING_TYPE},
+	},
+	ReturnType: parser.BOOL_TYPE,
+}
+
+func hasFunc(args []Value) Value {
+	if len(args) != 2 {
+		return newError("'has' takes 2 argument not " + strconv.Itoa(len(args)))
+	}
+	m := args[0].(*Map)
```
camh-, line 92: Since the param is ANY, won't you need to do type validating here instead of assuming a `(*Map)` type assertion will work?

pkg/evaluator/evaluator_test.go
```diff
@@ -690,6 +690,27 @@ end
 	}
 }
 
+func TestHas(t *testing.T) {
```
camh-, line 693: I think you need some error tests too for wrong types if you are going to do the type checking in the evaluator.

pkg/evaluator/builtin.go
```diff
@@ -94,6 +95,38 @@ func hasFunc(args []Value) Value {
 	return &Bool{Val: ok}
 }
 
+var delDecl = &parser.FuncDecl{
+	Name: "del",
+	Params: []*parser.Var{
+		{Name: "map", T: parser.ANY_TYPE},
```
camh-, line 101: some comments here as for `has`, below too in the same way.

pkg/evaluator/builtin.go
```diff
@@ -94,6 +95,38 @@ func hasFunc(args []Value) Value {
...
+}
+
+func delFunc(args []Value) Value {
+	if len(args) != 2 {
+		return newError("'del' takes 2 argument not " + strconv.Itoa(len(args)))
+	}
+	m := args[0].(*Map)
+	keyStr := args[1].(*String)
+	key := keyStr.Val
+	order := make([]string, len(*m.Order)-1)
```
camh-, line 114: by making a copy of order and replacing it, aren't you violating the reference semantics of maps? If you have:
```
a := {hello: "world"}
b := a
del a "hello"
```
then it should also be removed from b?

Thinking even more, I think b will be inconsistent - it will have `hello` in `Order`, but not in `Pairs` (because that is shared).

Instead I think you need to remove in-place and not allocate a new map - search for `key` in order and slice assign the element away - something like https://stackoverflow.com/a/37335777/23744 perhaps (although off the top of my head, I'm not 100% sure if that will not reallocate). It's a shame you can't use generics because x/slices has a delete function.

pkg/evaluator/builtin.go
juliaogris: I think it needs to be ANY because we can't say map of ANY value (just like in go where you for example can't type assert a slice of Any to a slice of int).
none_type does not work - this is real parameter type checking at the parser level.
The Name "map" is just the variable / parameter name (maybe I should change that).

juliaogris: renamed to `m`

pkg/evaluator/builtin.go
juliaogris: Fixed.

pkg/evaluator/evaluator_test.go
juliaogris: ok, done.

pkg/evaluator/builtin.go
juliaogris: ok, done.

pkg/evaluator/builtin.go
juliaogris: hmmm. Not sure this will / can work. It is more the other way round: when I pass `{}` to a function that expects `{}int` that will work.

I think it's not really possible to create type: map of anything - or at least it is not possible in go.... maybe i should allow it.

let's chat in person tomorrow.

pkg/evaluator/builtin.go
juliaogris: Ok, I have addressed this finally. I've moved the delete functionality into the value.go alongside insertKey and Get. seems like the better place.

camh- (approved):
:green_apple:

PR #48, opened 2022-11-03
github: Disable app_token script and use marketplace action

juliaogris:
Disable `app_token` script and use marketplace action 
`peter-murray/workflow-application-token-action@v1`
to create a temporary App `installation_token`.

This seems to work. 

Let's see if we can get brew formula update on merge to master and 
then circle back to work out the difference between this action and
our `app_token` bash script.


camh- (approved):
LGTM

PR #50, opened 2022-11-03
github: Fix app_token script

juliaogris:
Fix `app_token` script by using `openssl enc -A -base64` instead of
`base64` because `base64` adds linebreaks on linux but not on mac. The
flag to change this doesn't work on mac.

Reinstate usage on with `make release` and remove node based third-party
app. We had only temporarily changed this to see if the rest of the
brew formula update process in repo foxygoat/homebrew-tap would work.

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -27,14 +27,11 @@ jobs:
     - uses: actions/checkout@v3
       with:
         fetch-depth: 0
-    - id: app_token
-      uses: peter-murray/workflow-application-token-action@v1
-      with:
-        application_id: ${{ secrets.BOXYGOAT_GITHUB_APP_ID }}
-        application_private_key: ${{ secrets.BOXYGOAT_GITHUB_APP_PEM }}        
     - run: ./bin/make release
       env:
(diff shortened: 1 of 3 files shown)


.github/workflows/cicd.yaml
```diff
@@ -27,14 +27,11 @@ jobs:
...
-      uses: peter-murray/workflow-application-token-action@v1
-      with:
-        application_id: ${{ secrets.BOXYGOAT_GITHUB_APP_ID }}
-        application_private_key: ${{ secrets.BOXYGOAT_GITHUB_APP_PEM }}        
     - run: ./bin/make release
       env:
-        GITHUB_TOKEN: ${{ steps.app_token.outputs.token }}
+        GITHUB_APP_ID: ${{ secrets.BOXYGOAT_GITHUB_APP_ID }}
+        GITHUB_APP_PEM: ${{ secrets.BOXYGOAT_GITHUB_APP_PEM }}
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
juliaogris, line 34: This is still needed for `git push origin $(NEXTTAG)`

camh- (approved):
lgtm

PR #52, opened 2022-11-09
builtin: Add join and sprint functions

juliaogris:
Add join and sprint functions for easier string manipulation. Sprint, Join
and Print function have similarities for so they were refactored to use the
same underlying `join(args, sep) string` function in the same commit.

Join joins the string representations of the elements of an array with the
specified separator:

	join [1 "a" true] ", " // 1, a, true

`sprint` concatenates the arguments given with a space separator, like
`print`, but returns a string rather than printing to a specified output
location. Unlike print it does not add a newline at the end as this seems
to be a more likely use case - to be seen and changed if deemed bad.


camh- (approved):
:green_circle:

camh- (approved):
🍏

PR #54, opened 2022-11-10
frontend: Fix styles for iPad

juliaogris:
Fix styles for iPad, the output was dim grey on the right. Also, turn
off change `i` being spell-corrected to `I`.


camh- (approved):
iLGTM. :green_apple:

PR #57, opened 2022-11-12
evaluator: Fix variable initialisation

juliaogris:
Fix variable initialisation for inferred declaration and parameter
passing from another variable. Previously we would just use the same
value as the underlying variable which resulted in values of bath
variables changing even when only one was updated.

All newly added tests accept for `TestAssign3` failed prior to this fix.

### DIFF
--- a/pkg/evaluator/evaluator.go
+++ b/pkg/evaluator/evaluator.go
@@ -107,7 +107,7 @@ func (e *Evaluator) evalDeclaration(scope *scope, decl *parser.Declaration) Valu
 	if decl.Type() == parser.ANY_TYPE && val.Type() != ANY {
 		val = &Any{Val: val}
 	}
-	scope.set(decl.Var.Name, val)
+	scope.set(decl.Var.Name, copyOrRef(val))
 	return nil
 }
 
@@ -139,7 +139,7 @@ func (e *Evaluator) evalMapLiteral(scope *scope, m *parser.MapLiteral) Value {
 		if isError(val) {
 			return val
 		}
-		pairs[key] = val
+		pairs[key] = copyOrRef(val)
 	}
 	order := make([]string, len(m.Order))
 	copy(order, m.Order)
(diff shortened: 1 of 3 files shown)


pkg/evaluator/evaluator.go
```diff
@@ -107,7 +107,7 @@ func (e *Evaluator) evalDeclaration(scope *scope, decl *parser.Declaration) Valu
 	if decl.Type() == parser.ANY_TYPE && val.Type() != ANY {
 		val = &Any{Val: val}
```
camh-, line 108: trying to figure out why this nested val does not also need copyOrRef:
```
a := 1
b:any
b = a
a = 2
print a b
```
It prints `2 1` as expected, but from this code, it looks like it would share a reference?

camh- (commented):
You also need a `copyOrRef` in `evalArrayLiteral` and `evalMapLiteral`:

```
n := 1
a := [n n]
m := {n: n}
n = 2
print n a m
```

pkg/evaluator/evaluator.go
juliaogris: I just added:
```go
func TestAnyAssignment(t *testing.T) {
	prog := `
a := 1
b:any
b = a
a = 2
print a b
`
	b := bytes.Buffer{}
	fn := func(s string) { b.WriteString(s) }
	Run(prog, fn)
	want := "1 2\n"
	assert.Equal(t, want, b.String())
}
```
and it fails with:
```
    evaluator_test.go:943: want != got
        "1 2\n"
        "2 2\n"
```

juliaogris: same on evy.dev - any idea where you saw it print `2 1` ?

juliaogris: and of course I mean `2 1` above...

camh-: oh, maybe I didn't properly unfix it. I fixed it locally, implemented the test, but thought I unfixed it too.

Checking the test - oops. the test is wrong. I did it slightly differently in the test `print b`, but did a `want := "2\n"` instead of `1`. Ok, so I'm just clumsy, not mad. good to know

PR #59, opened 2022-11-12
tinygo: Upgrade to version 0.26.0

camh-:
Upgrade tinygo to version 0.26.0. We need to add `//export` directives
the Go function declarations for JavaScript implementations, according
to a tinygo commit message. And when you do that, you can no longer use
those functions as a value in the `evaluator.Runtime` struct. So wrap
the function references in anonymous funcs to make them Go functions.

The symbols in the Go environment for WebAssembly no longer have the
`main.` prefix, so drop them from the JS/Go/WebAssembly binding setup.
This may be because exported functions never had the Go package name
prefix?

Link: https://github.com/tinygo-org/tinygo/commit/8bbfb1ee68245a967081165c7161cd3421f97823


juliaogris (approved):
💚👗

Thank you so much!
Maybe a comment on the purpose of the bounce functions?

camh-:
> Maybe a comment on the purpose of the bounce functions?

I've removed the bounce functions and used anonymous functions instead. Added a comment as to why. Is this better?

juliaogris (approved):
🇳🇬

PR #61, opened 2022-12-06
Tweak CSS to make output look less faded on ios

juliaogris:
It seems ios creates some custom styles for disabled textareas. Fix
found on Stackoverflow.

Link: https://stackoverflow.com/a/23511280


camh- (approved):
:opaque_apple:

PR #63, opened 2023-01-24
evaluator: Refactor scopes

juliaogris:
Rework evaluator scopes so that they are part of the evaluator state.
A scope is in essence a linked list to it's outer scope so the data
structure can directly be used as a "stack" to push and pop scopes where 
needed. This work was initially done to re-work the evaluator execution
model so that we can react events such as OnClick and "step" through the
evaluation. However, we will explore this via channels next and keep the
actual execution call stack as is for now.

### DIFF
--- a/main.go
+++ b/main.go
@@ -43,8 +43,10 @@ func (c *cmdRun) Run() error {
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


camh- (approved):
:four_leaf_clover: LGTM

PR #64, opened 2023-01-31
frontend: Fix erroneously overridden go wasm env object

juliaogris:
Fix erroneously overridden go wasm env object. Instead of setting
go.importObject.env to a new object, extend existing one. This was the
result of somewhat misleading docs. While add it at better error handling
for wasm instantiation.

Housekeeping:

* Add prettier format and checks to make/CI
* Fix typo


camh- (approved):
:green_apple: LGTM

PR #65, opened 2023-02-04
all: Rework evy-evaluation to be hoisted on main

juliaogris:
Rework evy-evaluation to be hoisted on main rather than the separately
exported `evaluate` function. Some things, such as time.Sleep and
fmt.Sprintf only get properly initialised if the main*() function is
the entry point to the wasm code.

Additional, we had to solve to problem of having a JS function
returning a string in memory - we've done this in a hacky way by
returning Ptr and Length in separate function calls.

In preparation remove Lex and Parse button and wasm bindings.

### DIFF
--- a/frontend/index.js
+++ b/frontend/index.js
@@ -6,13 +6,13 @@ var wasm
 function initWasm() {
   const go = new Go() // see wasm_exec.js
   const evyEnv = {
-    jsPrint: jsPrint,
-    move: move,
-    line: line,
-    width: width,
-    circle: circle,
-    rect: rect,
-    color: color,
+    jsPrint,
+    move,
+    line,
+    width,
+    circle,
+    rect,
+    color,
   }
   go.importObject.env = Object.assign(go.importObject.env, evyEnv)
   WebAssembly.instantiateStreaming(fetch("evy.wasm"), go.importObject)
--- a/frontend/index.js
+++ b/frontend/index.js
@@ -23,11 +23,9 @@ function initWasm() {
     .catch((err) => {
       console.error(err)
     })
-  document.querySelectorAll("header button").forEach((button) => {
-    button.onclick = handleRun
-    button.disabled = false
-    window.location.hash.includes("debug") && button.classList.remove("hidden")
-  })
+  const button = document.getElementById("run")
+  button.onclick = handleRun
+  button.disabled = false
 }
 
 // jsPrint converts wasm memory bytes from ptr to ptr+len to string and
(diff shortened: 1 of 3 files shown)


camh- (commented):
LGTM I think. Not a fan of the global source variables - wondering if they can be somehow hoisted into the wasm instance, but its fine to get going I think.

camh- (approved):

frontend/index.js
```diff
@@ -1,34 +1,23 @@
...
-  WebAssembly.instantiateStreaming(fetch("evy.wasm"), go.importObject)
+  WebAssembly.compileStreaming(fetch("evy.wasm"))
     .then(function (obj) {
-      wasm = obj.instance
-      go.run(wasm)
+      wasmModule = obj
     })
     .catch((err) => {
       console.error(err)
     })
```
camh-, line 15: nit: This could be formatted with fewer syntactical elements I think:
```
    .then(obj => wasmModule = obj)
    .catch(err => console.error(err))
```

frontend/index.js
```diff
@@ -1,34 +1,23 @@
 "use strict"
 
-var wasm
+let wasmModule, wasmInst
+let sourcePtr, sourceLength
+const runButton = document.getElementById("run")
```
camh-, line 5: How come you make this a global? It's only used in `initWasm()` it looks like.

frontend/index.js
```diff
@@ -40,25 +29,47 @@ function jsPrint(ptr, len) {
...
+  sourceLength = bytes.length
+
   document.getElementById("output").textContent = ""
   resetCanvas()
-  wasm.exports.evaluate(ptr, bytes.length)
+
+  go.run(wasmInst)
 }
 
+function reset() {}
```
camh-, line 71: This does not seem to be used.

camh- (approved):

frontend/index.js
juliaogris: done.

frontend/index.js
juliaogris: sorry, that was from my follow up work where I changed it to a `Stop....` while running and made it a cancel button as well. Reverted.

frontend/index.js
juliaogris: Same. Removed.

juliaogris: SORRY!

camh- (approved):
:green_book: LGTM

PR #67, opened 2023-02-06
all: Implement stop

juliaogris:
Implement stop in frontend by turning the Run button into a Stop button
during execution. For this we also need an onExit callback into JS from
go when execution finishes so that the button label gets reset.

Additionally we need to occasionally yield to JS event processing via a
`time.Sleep` in the evaluator code. This operation is abstracted via
the Yielder interface.

Execution is stopped in Eval if the Stopped flag is set, which gets set
on click via the exported Go function stop(). The access to the stop
flag is not synchronised, but as JS is single threaded so we don't need
to worry about it.

In a preparatory commit refactor evaluator and builtins so that there is 
a separate `NewEvaluator` function and `e.Run` method. This allows for
`wasm/main.go` to hold an evaluator as global variable on which we set the 
`evaluator.Stopped` flag. 

Evy program for testing.
```
while true 
   rect 10 10
end
```
another one with sleep:
```
for i := range 0 100 0.1
  clear
  dot i 50

  sleep 0.01
end

func clear
  color "white"
  move 0 0 
  rect 100 100
end

func dot x:num y:num
  color "red"
  move x y
  circle 10
end
```

### DIFF
--- a/pkg/evaluator/evaluator.go
+++ b/pkg/evaluator/evaluator.go
@@ -16,10 +16,21 @@ func NewEvaluator(builtins Builtins) *Evaluator {
 }
 
 type Evaluator struct {
+	Stopped  bool
+	Yielder  Yielder
 	print    func(string)
 	builtins map[string]Builtin
 
 	scope *scope // Current top of scope stack
+}
+
+// Yielder abstracts the method of yielding the CPU so that
+// JavaScript/browser events get a chance to be processed. Currently
+// (Feb 2023) it seems that you can only yield to JS by sleeping for at
+// least 1ms but having that delay is not ideal. Other methods of
+// yielding can be explored by implementing a different Yielder.
+type Yielder interface {
+	Yield()
 }
 
 func (e *Evaluator) Run(input string) {
@@ -35,7 +46,13 @@ func (e *Evaluator) Run(input string) {
 	}
 }
 
+var ErrStopped = newError("stopped")
+
 func (e *Evaluator) Eval(node parser.Node) Value {
+	if e.Stopped {
+		return ErrStopped
+	}
+	e.yield() // Yield to give JavaScript/browser events a chance to run.
 	switch node := node.(type) {
 	case *parser.Program:
 		return e.evalProgram(node)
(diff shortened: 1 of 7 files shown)


pkg/evaluator/evaluator.go
```diff
@@ -35,7 +41,13 @@ func (e *Evaluator) Run(input string) {
 	}
 }
 
+var ErrStopped = newError("stopped")
+
 func (e *Evaluator) Eval(node parser.Node) Value {
+	if e.Stopped {
+		return ErrStopped
+	}
+	e.yield()
```
camh-, line 50: Perhaps a comment here to say why this yield exists?

```
// Yield to give JavaScript/browser events a chance to run.
```

pkg/evaluator/evaluator.go
```diff
@@ -16,12 +16,18 @@ func NewEvaluator(builtins Builtins) *Evaluator {
...
 type Evaluator struct {
+	Stopped  bool
+	Yielder  Yielder
 	print    func(string)
 	builtins map[string]Builtin
 
 	scope *scope // Current top of scope stack
 }
 
+type Yielder interface {
```
camh-, line 27: a comment perhaps?

```
// Yielder abstracts the method of yielding the CPU so that JavaScript/browser
// events get a chance to be processed. Currently (Feb 2023) it seems that you
// can only yield to JS by sleeping for at least 1ms but having that delay is
// not ideal. Other methods of yielding can be explored by implementing a
// different Yielder.
```

pkg/wasm/main.go
```diff
@@ -24,6 +27,26 @@ func getSource() string {
 	return getString(ptr, length)
 }
 
+type yielder struct {
```
camh-, line 30: perhaps call this `sleepingYielder` or something?

camh- (approved):
:green_salad: LGTM

pkg/wasm/main.go
juliaogris: done.

pkg/evaluator/evaluator.go
juliaogris: nice thanks. done.

pkg/evaluator/evaluator.go
juliaogris: done.

PR #68, opened 2023-02-08
evaluator: Tighten error handling with panics

juliaogris:
Tighten error handling with panics rather than returning false or nil so
that we can track down errors more easily in when they are a result of
internal coding errors.


camh- (approved):
:leaves: LGTM

PR #69, opened 2023-02-08
parser: Error on function redeclaration

juliaogris:
Error on function redeclaration, including builtin functions. Before we
would just let new functions override existing functions which is
pretty confusing, err early err often.

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -51,7 +51,11 @@ func (e Error) String() string {
 
 func New(input string, builtins map[string]*FuncDecl) *Parser {
 	l := lexer.New(input)
-	p := &Parser{funcs: builtins, wssStack: []bool{false}}
+
+	p := &Parser{funcs: map[string]*FuncDecl{}, wssStack: []bool{false}}
+	for name, funcDecl := range builtins {
+		p.funcs[name] = funcDecl
+	}
 
 	// Read all tokens, collect function declaration tokens by index
 	// funcs temporarily holds FUNC token indices for further processing
@@ -79,8 +83,13 @@ func New(input string, builtins map[string]*FuncDecl) *Parser {
 	for _, i := range funcs {
 		p.advanceTo(i)
 		fd := p.parseFuncDeclSignature()
-		if fd != nil {
+		switch {
+		case p.funcs[fd.Name] == nil:
 			p.funcs[fd.Name] = fd
+		case builtins[fd.Name] == nil:
+			p.appendErrorForToken("redeclaration of function "+fd.Name, fd.Token)
+		case builtins[fd.Name] != nil:
+			p.appendErrorForToken("cannot override builtin function "+fd.Name, fd.Token)
 		}
 	}
 	return p
--- a/pkg/parser/parser_test.go
+++ b/pkg/parser/parser_test.go
@@ -967,6 +967,32 @@ end
 	}
 }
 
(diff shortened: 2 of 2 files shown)


pkg/parser/parser.go
```diff
@@ -51,7 +51,11 @@ func (e Error) String() string {
 
 func New(input string, builtins map[string]*FuncDecl) *Parser {
 	l := lexer.New(input)
-	p := &Parser{funcs: builtins, wssStack: []bool{false}}
+
+	p := &Parser{funcs: map[string]*FuncDecl{}, wssStack: []bool{false}}
+	for name, funcDecl := range builtins {
```
camh-, line 56: :pineapple: shame you can't use generics - you could use `x/exp/maps.Clone()` here (or on the line before even: `funcs: maps.Clone(builtins)`)

pkg/parser/parser.go
```diff
@@ -80,6 +84,14 @@ func New(input string, builtins map[string]*FuncDecl) *Parser {
 		p.advanceTo(i)
 		fd := p.parseFuncDeclSignature()
 		if fd != nil {
+			if p.funcs[fd.Name] != nil {
```
camh-, line 87: perhaps a switch?:
```
switch {
case p.funcs[fd.Name] == nil:
  p.funcs[fd.Name] = fd
case builtins[fd.Name] == nil:
  p.appendErrorForToken("redeclaration of function "+fd.Name, fd.Token)
case builtins[fd.Name] != nil: // could be `default:`, but maybe this reads better?
  p.appendErrorForToken("cannot override builtin function "+fd.Name, fd.Token)
}
```

camh- (approved):
:leafy_green: LGTM

pkg/parser/parser.go
camh-: or perhaps swap the last two cases?

pkg/parser/parser.go
juliaogris: oh nice!! i think tinygo 1.27.0 is coming out soon and it works with got 1.20.
I wonder if the issue was again that I didn't hoist things on main.
I've got a cleanup PR due that replaces all errormessages with Printfs now that that's possible. In fact, I should maybe consider exposing them in evy!?

pkg/parser/parser.go
juliaogris: ok.

juliaogris:
```go
switch {
case  p.funcs[fd.Name] != nil && builtins[fd.Name] == nil:
  p.appendErrorForToken("redeclaration of function "+fd.Name, fd.Token)
case p.funcs[fd.Name] != nil && builtins[fd.Name] != nil: 
  p.appendErrorForToken("cannot override builtin function "+fd.Name, fd.Token)
default:
  p.funcs[fd.Name] = fd
}
```

juliaogris: any of them. i'll got with your suggestion.

camh-: I deliberately didn't do it your way since you can take advantage of the case ordering to put the happy path first and not require the more complicated case expressions. but up to you

juliaogris: but we don't do the happy path first typically.
we do it left and last, don't we?

camh-: but rules are made to be broken. happy path is more about less nesting than what comes first. it just ends up putting the happy path last mostly.

PR #71, opened 2023-02-09
Parser: Add variable validation in for statements

juliaogris:
Add variable validation in for statements. In specific, variable
declared in the as `for VAR :=  range ...` should follow the same
validation rules as other variable declarations. It's been an
oversight.


juliaogris:
Folding this work in #70

juliaogris:
I've folded this PR in to #70 - easier to keep track of.

PR #73, opened 2023-02-10
all: Wire eventHandler to frontend

juliaogris:
Wire eventHandler to frontend so that when `on <event>` block appears in
evy code the corresponding frontend event triggers the execution of
that block. Currently implemented events: `down`, `up`, `move`, `key`.

Only `frontend:` and `all:` commits are relevant, the `parser:` commits are 
covered in #70 

Sample code
https://evy-lang--73-51oq7tbx.web.app#draw

```
on down x:num y:num
  print "down" x y
end

on up
  print "up"
end

on key c:string
   print "key" c
end
```

simple drawing tool

```
d := false
width 1

on down x:num y:num
  d = true
  move x y
end

on up
  d = false
end

on move x:num y:num
  if d
    line x y
  end
end
```

### DIFF
--- a/frontend/index.js
+++ b/frontend/index.js
@@ -58,6 +58,7 @@ async function handleRun(event) {
 
 // onStopped is exported to evy go/wasm and called when execution finishes
 function onStopped() {
+  removeEventHandlers()
   runButton.innerText = "Run"
 }
 
@@ -71,6 +72,7 @@ function newEvyGo() {
     rect,
     color,
     onStopped,
+    registerEventHandler,
     sourcePtr: () => sourcePtr,
     sourceLength: () => sourceLength,
   }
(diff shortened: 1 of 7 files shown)


frontend/samples/draw.evy
```diff
@@ -0,0 +1,17 @@
+down := false
+width 1
+
+on down x:num y:num
+  down = true
+  move x y
+end
+
+on up
+  down= false
```
camh-, line 10: nit: missing space before equals

pkg/evaluator/evaluator.go
```diff
@@ -16,21 +18,28 @@ func NewEvaluator(builtins Builtins) *Evaluator {
...
 type Evaluator struct {
-	Stopped  bool
-	Yielder  Yielder // Yield to give JavaScript/browser events a chance to run.
-	print    func(string)
-	builtins Builtins
+	Stopped       bool
+	Yielder       Yielder // Yield to give JavaScript/browser events a chance to run.
+	print         func(string)
+	builtins      Builtins
+	eventHandlers map[string]*parser.EventHandler //
```
camh-, line 25: nit: empty comment at EOL

pkg/evaluator/evaluator.go
```diff
@@ -16,21 +18,28 @@ func NewEvaluator(builtins Builtins) *Evaluator {
...
 
 	scope *scope // Current top of scope stack
 }
 
+type Event interface {
```
camh-, line 30: Does this interface add much? What if it were:

```
type Event struct {
  Name string
  Params []any
}
```
and instead of the `xyEvent` and `stringEvent` structs, you could just have
```
func newXYEvent(name string, x, y float64) *Event {
  return &Event{Name: name, Params: []any{x, y}}
}
func newStringEvent(name, str string) *Event {
  return &Event{Name:name, Params: []any{str}}
}
```
and the use these functions in the Go handlers?

Because nothing seems to access the individual fields of the `xyEvent` and `stringEvent` structs.

camh- (approved):
I put a comment here before about a commit message - not sure if it will show up when I submit the review, so if not, i'll follow up with another comment.

camh-:
The commit message for "frontend: Refactor canvas scale parameters" says:

> logical top right left: 100, 100

I think there's an extra "left" at the end that should not be there.

frontend/index.js
```diff
@@ -244,5 +244,43 @@ function circle(r) {
...
+
+function ctrlEnterListener(e) {
+  if ((e.metaKey || e.ctrlKey) && event.key === "Enter") {
+    handleRun()
+  }
+}
+
+async function fetchSource() {
+  // parse url fragment into object
+  // e.g. https://exmple.com#a=1&b=2 into {a: "1", b"2"}
```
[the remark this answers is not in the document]
juliaogris: thanks, done.

frontend/samples/draw.evy
juliaogris: done, really need a formatter...

pkg/evaluator/evaluator.go
juliaogris: done, thanks.

pkg/evaluator/evaluator.go
juliaogris: ok, agreed.

juliaogris: really much better actually. thank you.

juliaogris:
ok all addressed i think. I will merge now, but if you can be bothered to look again and you find something I've missed, please let me know. 🍪

PR #74, opened 2023-02-12
wasm: Rework yielding so that it doesn't interfere with sleep

juliaogris:
Rework yielding so that is doesn't interfere with sleep. With the
currently deployed evy playground an endless loop without a `sleep`
crashes the browser tab, as the most recent Yielding bug-fix has
introduced a new bug. sigh.

In this fix we reset the yield timeout and eval counter every time we run
`sleep` as part of the evy program. This too seems fragile, but I'm running
out of ideas and empirically it seems to work better.

Add URL fragment shortcut to evy samples, #s=myprog tries to load
"samples/myprog.evy", again introduced for easier testability.

Only the last 2 commits are relevant, but to avoid merge conflicts I've branched
off `on-eval`

Both of these examples are stoppable now 😅
https://evy-lang--74-wc49wsco.web.app/#s=forever
https://evy-lang--74-wc49wsco.web.app/#s=movingdot

```
func target x:num y:num
  move x y
  colors := ["yellow" "gold" "orange"  "red" "maroon"]
  radius := 5
  for c := range colors
     color c
     circle radius
     radius = radius - 1
     print "color" c "radius" radius
  end
end


target 40 40 
```
causes:
```
evy.wasm:0x592a Uncaught (in promise) RuntimeError: unreachable
    at runtime.runtimePanic (evy.wasm:0x592a)
    at runtime.nilPanic (evy.wasm:0x2894)
    at internal/task.Pause (evy.wasm:0x5c29)
    at runtime.deadlock (evy.wasm:0x15d26)
    at runtime.run$1$gowrapper (evy.wasm:0x164c0)
    at tinygo_launch (evy.wasm:0x5e9ac)
    at (*internal/task.Task).Resume (evy.wasm:0x57a7)
    at runtime.scheduler (evy.wasm:0x166af)
    at _start (evy.wasm:0x1640e)
    at _start.command_export (evy.wasm:0x5ecde)
    ```


juliaogris:
ok, this was another failed attempt. #76 finally fixed it (`-stack-size=32kb`, slack convo https://gophers.slack.com/archives/CDJD3SUP6/p1676264760211069 )

PR #77, opened 2023-02-14
builtin: Add printf and sprintf functions

juliaogris:
Add printf and sprintf functions using Go's fmt.Sprintf (which now that we
hoist wasm code via main works 🚀). To allow for formatted printed of
numbers and bools add and `unwrap` function to value.go, unwrapping the
value of the Value interface implementation for basic types only.

In the process of creating printf, we discovered some missing unescaping in
the lexing stage as the goal is to keep string handling consistent with Go
and print "\n" as newlines. Fix it and improve error message for invalid
strings.

While at fix evaluator test output for multiline test cases and failed
tests.

### DIFF
--- a/pkg/evaluator/builtin.go
+++ b/pkg/evaluator/builtin.go
@@ -1,6 +1,7 @@
 package evaluator
 
 import (
+	"fmt"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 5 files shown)


camh- (commented):
This PR has a stray commit on the end I think "builting: str2num & str2bool wip"

camh- (approved):
LGTM apart from the stray commit

PR #78, opened 2023-02-14
evalutor: Add str2num and str2bool functions

juliaogris:
Add builtin functions `str2num` and `str2bool` converting strings to
numbers and bools. They set the evy global variables `err` and `errmsg` in
case of an error.

Rework builtins to take scope so that they can set `err` and `errmsg`.

Disallow redeclaration of builtin globals in parser.

For good housekeeping, fix inconsistent error messages in parser


camh- (approved):
:green_apple: LGTM

PR #79, opened 2023-02-15
all: Add read function

juliaogris:
Add blocking read function for line-by-line reading from text input.

Rework access to "returned" strings for function imported from JS to
go/wasm, used in getting evy source and readline. To achieve blocking
reads while allowing interactivity in JS code sleep for 50ms between
every unsuccessful read attempt in go/wasm.

Try it with
https://evy-lang--79-se5aqk1b.web.app#echo


camh- (approved):
:green_book: LGTM

PR #80, opened 2023-02-16
all: Add `on input` handler

juliaogris:
Add `on input id:string val:string` handler, so that evy can react to any
kind of input event, depending on how the frontend has wired it. By default
we have added and hidden to sliders in the index.html / DOM.

An example can be explored at
https://evy-lang--80-k6pkxs0k.web.app#show=.slider&source=samples/sliders.evy

In preparation change visibility of evy-widgets in dom via URL fragment.

### DIFF
--- a/frontend/index.css
+++ b/frontend/index.css
@@ -168,6 +168,18 @@ textarea:disabled {
   font-weight: bold;
 }
 
+.slider {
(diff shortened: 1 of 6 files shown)


camh- (approved):
:green_circle: LGTM

PR #82, opened 2023-02-17
builtin: Add `rand` and strings functions

juliaogris:
Add random number generation functions:

* `rand n` returns an integer in the half open interval [0,n)
* `rand1` returns an float64 in the half open interval [0,1)

Add string functions `upper`, `lower`, `index`, `trim`,
`startswith`, `endswith`, `replace`.

This functions are mapped to Go's math/rand package. It seems that
tinygo still requires seeding (tested).

Add a demos for clarity.

Minor bug fix for non 2xx source response: don't load random HTLM,
write error to console instead.

https://evy-lang--82-r0m72zv4.web.app/#source=samples/rand.evy&show=.read
https://evy-lang--82-r0m72zv4.web.app/#strings

### DIFF
--- a/frontend/samples/rand.evy
+++ b/frontend/samples/rand.evy
@@ -1,0 +1,37 @@
+while true
+  print "╔════════════════════════╗"
+  print "║ Guess my number (1-10) ║"
+  print "╚════════════════════════╝"
+
+  n := 1 + (rand 10)
+  guess := readn
+
+  while guess != n
+    if guess > 10
+      print guess "Guess lower (1-10)."
+    else if guess < 1
+      print guess " Guess higher (1-10)."
+    else if guess < n
+       print guess "Guess higher."
+    else
+       print  guess "Guess lower."
+    end
+    guess = readn
+  end
+  print guess "You got it 🥳."
+  print
+end
+
+// readn reads a number.
+// If you enter a string that is not number
+func readn:num
+  while true
+    str := read
+    n := str2num str
+    if !err
+      return n
+    end
+    print "💣 '"+str+"'is not a number. Try again."
+  end
+  return -1
+end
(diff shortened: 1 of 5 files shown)


juliaogris:
From Mali: hey mum, the sample code would be better like this:

```
while true
  print "Guess my number (1-6)."

  n := 1 + (rand 6)
  guess := readn

  while guess != n
    if n > guess
       print "You've guessed" guess "- guess higher."
    else
       print "You've guessed" guess "- guess lower."
    end
    guess = readn
  end
  print "You've guessed" guess " - You got it right 🥳."
end

// readn reads a number. 
// If you enter a string that is not number
func readn:num
  while true
    str := read
    n := str2num str
    if !err
      return n
    end
    print str "is not a number, try again."
  end
  return -1
end

```

frontend/samples/rand.evy
```diff
@@ -0,0 +1,37 @@
...
+  print "╔════════════════════════╗"
+  print "║ Guess my number (1-10) ║"
+  print "╚════════════════════════╝"
+
+  n := 1 + (rand 10)
+  guess := readn
+
+  while guess != n
+    if guess > 10
+      print guess "Guess higher (1-10)."
```
camh-, line 11: should this be "Guess lower" and the next "Guess higher"?

camh- (approved):
:peacock: LGTM

I noticed when I tried the guessing game that the input line was not focused, so I typed a number (nothing happened) and I pressed enter, and that stopped the program because the Stop button was still focused. Perhaps calling `read` should focus on the input?

Also, I guessed the right number 3 times, so I think I deserve a cookie.

frontend/samples/rand.evy
juliaogris: :cookie:
thanks for the review!

juliaogris: I fixed this up. 
Your comment re input focus is a really good one. also fixed.

PR #83, opened 2023-02-17
go: Add modulo operator `%` to evy

juliaogris:
Add modulo operator to evy, which includes extending lexer, parser and
evaluator. As in other programming languages, use `%` symbol and give
the operator the same precedence as multiplication or division.


camh- (approved):
:sauropod:

PR #84, opened 2023-02-17
go: Allow `for` statement without var declaration

juliaogris:
Allow for statement without var declaration, e.g.:

	for range 2
       print "Hello"  // executed twice
	end

This fell in place quite easily, however, the nesting might be deeper
than ideal - looking forward to feedback.


camh- (approved):
:four_leaf_clover: This seems fine how you did it. Can't think of a better way.

PR #85, opened 2023-02-18
parser: Implement anonymous variable `_`

juliaogris:
Implement anonymous variable `_` which cannot be read and only be
declared in function or event handler parameters. In a for loop prefer
a range statement without assignment.


pkg/parser/parser_test.go
```diff
@@ -991,6 +1001,11 @@ func fox
    print "fox overridden"
 end
 `: "line 6 column 1: redeclaration of function 'fox'",
+		`
+func fox _:string
+   print "fox" _
+end
+`: "line 3 column 16: anonymous variable '_' cannot be read",
```
camh-, line 1008: fox can read your mind. surely fox can read the underscore. :pineapple:

camh- (approved):
:cucumber:

PR #86, opened 2023-02-18
parser: Allow for multiline array and map declarations

juliaogris:
Allow for multiline array and map declarations so that it is easier to
declare large blocks of data. We want to get this in before we look
into `evy format` or syntax colouring as this might impact things.


camh- (approved):
:pear:

PR #87, opened 2023-02-20
parser: Prepare for `evy format`

juliaogris:
Prepare for `evy format` to print parse-error free source code 
in canonical way.

We will attach comments to the Stmt nodes in a follow up PR
when adding formatting, for this reason we introduce distinct
node for `FunCallStmt`, `InferredDeclStmt`, `TypedDeclStmt` and
`GroupedExpression` (expressions with `()`).

Remove the `Format` method from parser.Type to free up the Format
name purely for formatting and clean up its usage a little.

### DIFF
--- a/pkg/evaluator/builtin.go
+++ b/pkg/evaluator/builtin.go
@@ -12,18 +12,18 @@ import (
 
 type Builtin struct {
 	Func BuiltinFunc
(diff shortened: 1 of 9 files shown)


camh- (approved):
:recycle: LGTM

PR #88, opened 2023-02-22
parser: Fix empty array bug

juliaogris:
Fix empty array bug mutating the GENERIC_ARRAY and GENERIC_MAP global
variable. These types are used when using the inferred declaration

    a := []

At this point it is not clear which type or array a will be, but to say
it will be of type `any` is restricting it early. When we then assign a
typed array to a, e.g. with `a = [1 2]` a becomes an array of num. The
type is updated via type.Infer(), however, the Infer method mutated
itself (the GENERIC_ARRAY). Fix this by returning a copy.

This in turn made obvious that

    for i := []
      print i
    end

caused an error, because we talk for ArrayRanging `array.Type.Sub`,
which we've now changed to `array.Type.Infer().Sub`

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -752,7 +752,7 @@ func (p *Parser) parseForStatement(scope *scope) Node {
 		forNode.Range = n
 	case ARRAY:
 		if forNode.LoopVar != nil {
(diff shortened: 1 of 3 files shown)


camh- (approved):
:beetle:

PR #89, opened 2023-02-22
parser: Fix variadic parameter bug

juliaogris:
Fix variadic parameter bug, where the variadic parameter gets added to
the scope in its original form rather than as an array of it. Before
we'd add for

    func x n:num...

x with type `num` rather than `[]num` to the function body's scope.
Fix and add tests to evaluation and parsing.

### DIFF
--- a/pkg/evaluator/evaluator_test.go
+++ b/pkg/evaluator/evaluator_test.go
@@ -998,6 +998,48 @@ print 5 errmsg
 	}
 }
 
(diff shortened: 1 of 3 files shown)


camh- (approved):
:cactus: LGTM, although the evy test code indenting is all over the place - the evaluator code is 3 spaces for the first level and 2 for the next. For the parser, one block is indented 8, the other is 3. Up to you if you care.

PR #90, opened 2023-02-23
frontend: Fix read textfield and hashchange

juliaogris:
Fix read UI, so that it does not keep on grabbing the focus while
running. Ensure read input text field is shown.

Fix hashchange listener by attaching it to window, rather than document. We
now actually get a source change when updating the URL fragment.


camh- (approved):
:cucumber: LOKTM

PR #91, opened 2023-02-24
all: Add formatting

juliaogris:
Implement formatting in parser and wire in frontend prior to every
evaluation. Formatting was purposefully kept mostly to the 
`parser/format.go` file.

This is feature complete formatting, the only outstanding work
is to add `evy format [-w] FILE...` to evy CLI.

### DIFF
--- a/frontend/index.js
+++ b/frontend/index.js
@@ -53,6 +53,12 @@ function evySource() {
   return stringToMemAddr(code)
 }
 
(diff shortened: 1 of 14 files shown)


juliaogris:
It seems that tinygo a map issue with interfaces, so we'll work around this by storing comments, wss and mulitine info with ast nodes in #92.

PR #92, opened 2023-02-25
all: Implement formatting of evy code v3

juliaogris:
Implement formatting in parser and wire in frontend prior to every
evaluation. Formatting was purposefully kept mostly to the
`parser/format.go` file.

This is feature complete formatting, the only outstanding work
is to add `evy format [-w] FILE...` to evy CLI.

Reformat sample code by running:

* https://evy-lang--92-crcf8aqh.web.app/#animate
* https://evy-lang--92-crcf8aqh.web.app/#dot
* https://evy-lang--92-crcf8aqh.web.app/#draw
* https://evy-lang--92-crcf8aqh.web.app/#echo
* https://evy-lang--92-crcf8aqh.web.app/#rand
* https://evy-lang--92-crcf8aqh.web.app/#sliders
* https://evy-lang--92-crcf8aqh.web.app/#strings

Initially we kept all additional info with the `formatting` which
revealed a but in tinygo, that has since been fixed, see
[gophers slack](https://gophers.slack.com/archives/CDJD3SUP6/p1677288675055869) 
for details.

I could not reopen that PR, but I've raised a new one for comparison #94 .

PR #93, opened 2023-02-25
all: Keep UX and go code tidy

juliaogris:
Keep UX in frontend for frequent sample switching. K
Keep go code tidy with ForStmt.String() fix and dead code removal.

This PR was shaved off #92 as there were a few too many unrelated fixes.


camh- (approved):
:potted_plant:

PR #94, opened 2023-02-25
all: Implement formatting of evy code

juliaogris:
Implement formatting in parser and wire in frontend prior to every
evaluation. Formatting was purposefully kept mostly to the
`parser/format.go` file.

This is feature complete formatting, the only outstanding work
is to add `evy format [-w] FILE...` to evy CLI.

Reformat sample code by running:

https://evy-lang--94-x5xuqzz3.web.app/#animate
https://evy-lang--94-x5xuqzz3.web.app/#dot
https://evy-lang--94-x5xuqzz3.web.app/#draw
https://evy-lang--94-x5xuqzz3.web.app/#echo
https://evy-lang--94-x5xuqzz3.web.app/#rand
https://evy-lang--94-x5xuqzz3.web.app/#sliders
https://evy-lang--94-x5xuqzz3.web.app/#strings

We had to hack around not using interfaces as map keys by converting them
`uinptr` to work around a tinygo-0.27.0 bug which has been fixed on 
dev, see gophers slack link for details. Thanks for the idea and implementation
help @camh- !

Link: https://gophers.slack.com/archives/CDJD3SUP6/p1677288675055869

### DIFF
--- a/pkg/parser/format.go
+++ b/pkg/parser/format.go
@@ -8,8 +8,9 @@ import (
 
 func newFormatting() *formatting {
 	return &formatting{
-		wss:      map[*BinaryExpression]bool{},
-		comments: map[uintptr]string{},
+		wss:       map[*BinaryExpression]bool{},
+		comments:  map[uintptr]string{},
+		multiline: map[uintptr][]multilineItem{},
 	}
 }
 
@@ -24,7 +25,8 @@ type formatting struct {
 	//
 	// the current implementation is a hack to work around a tinygo but, details:
 	// https://gophers.slack.com/archives/CDJD3SUP6/p1677288675055869
-	comments map[uintptr]string
+	comments  map[uintptr]string
+	multiline map[uintptr][]multilineItem
 
 	indentLevel int
 }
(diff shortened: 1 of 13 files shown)


pkg/parser/format.go
```diff
@@ -0,0 +1,343 @@
...
+		comments:    map[Node]string{},
+		indentLevel: -1,
+	}
+}
+
+type formatting struct {
+	w io.StringWriter
+
+	wss         map[*BinaryExpression]bool
+	comments    map[Node]string
```
juliaogris, line 22: Here we save comments and whitespace concerns purely with the formatting struct and leave the AST alone. Unfortunately this causes an error with tinygo 0.27.0 which has since been fixed on it's dev branch.

pkg/parser/multiline.go
```diff
@@ -0,0 +1,84 @@
...
+//	    1 // commment1
+//	    2
+//
+//	     // comment2
+//	     3
+//	]
+//
+// has the representation:
+//
+//	[ "el", "// comment1", "el", "", "", "// comment2", "el", ""]
```
camh-, line 31: Should the first slice element be `""` ? The first element is on the second line. With the map example below, it starts with `""` - yeah, looking at the test cases, this should start with `""`

camh- (approved):
:green_heart: Sorry it took me so long to get to. All looks good. It must be nice to have formatted code now.

pkg/parser/multiline.go
juliaogris: Thanks, fixed (I had it correct but then mucked it up in the re-format of the comment....)

PR #95, opened 2023-02-26
main: Add `evy fmt` command

juliaogris:
Implement `evy fmt` command via parser.Program.Format() and add it to
Makefile for formatting and CI checks.

While at it clean up evy CLI tool a little bit.

### DIFF
--- a/main.go
+++ b/main.go
@@ -4,6 +4,7 @@ package main
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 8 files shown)


main.go
```diff
@@ -15,23 +16,35 @@ import (
...
+	Parse    parseCmd    `cmd:"" help:"Parse evy program" hidden:""`
 }
 
 type runCmd struct {
 	Source string `arg:"" help:"Source file. Default stdin" default:"-"`
 }
 
+type fmtCmd struct {
+	Write     bool     `short:"w" help:"update .evy file" xor:"mode"`
+	ExitError bool     `short:"e" help:"" xor:"mode"`
```
juliaogris, line 44: Not sure if this a good name....

main.go
```diff
@@ -51,6 +64,66 @@ func (c *runCmd) Run() error {
 	return nil
 }
 
+func (c *fmtCmd) Run() error {
+	if len(c.Files) == 0 {
+		return format(os.Stdin, os.Stdout, c.Check)
```
camh-, line 69: You should probably return an error here if `c.Write` is true - can't write to files that are not listed

main.go
```diff
@@ -51,6 +64,66 @@ func (c *runCmd) Run() error {
...
+		if err != nil {
+			return err
+		}
+		if c.Write {
+			out, err = os.CreateTemp("", "evy")
+			if err != nil {
+				return fmt.Errorf("%s: %w", filename, err)
+			}
+		}
+		if err := format(in, out, !c.Write); err != nil {
```
camh-, line 83: should `!c.Write` be `c.Check` ? You should be able to run `evy fmt foo.evy` and it will write the output to stdout. This is what `goimports` does

main.go
```diff
@@ -51,6 +64,66 @@ func (c *runCmd) Run() error {
...
+			}
+			if err := os.Rename(tempFile.Name(), filename); err != nil {
+				return fmt.Errorf("%s: %w", filename, err)
+			}
+		}
+	}
+	return nil
+}
+
+func format(r io.Reader, w io.StringWriter, exitError bool) error {
```
camh-, line 102: nit: I think I would call `exitError` `checkOnly` - uses the same terminology as the flags

camh- (approved):
:broccoli:

main.go
```diff
@@ -15,32 +16,44 @@ import (
...
-type cmdRun struct {
+type runCmd struct {
 	Source string `arg:"" help:"Source file. Default stdin" default:"-"`
 }
 
-type cmdTokenize struct {
+type fmtCmd struct {
+	Write bool     `short:"w" help:"update .evy file" xor:"mode"`
+	Check bool     `short:"c" help:"check if already formatted" xor:"mode"`
+	Files []string `arg:"" help:"Source files. Default stdin"`
```
camh-, line 45: oops. missed this. this needs `optional:""` otherwise `len(c.Files)` is never zero and cannot be.

juliaogris: oops indeed. fixed.

PR #96, opened 2023-02-26
parser: Insert NL before `func`

juliaogris:
Insert NL before `func` and `on` statements for function and eventhandler
declarations. Honour leading comments. Insert newline after func and event
handler declarations if directly followed by a stmt.


camh- (approved):
:broccoli:

PR #101, opened 2023-03-14
parser: Fix formatting bug

juliaogris:
Fix two formatting bugs that rendered evy program un-executable but
required only minor source updates:

- escape `\n` in stringliterals with strconv.Quote.
  Previously a literal newline was printed inside the string.
- ensure unary operators get formatted properly.
  Previously `while !(check)` formatted to `while !check()`

Add test for the fixes and ensure tests failed previous to fix.

### DIFF
--- a/pkg/parser/format.go
+++ b/pkg/parser/format.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"reflect"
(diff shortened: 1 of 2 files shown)


camh- (approved):
:bug:

PR #104, opened 2023-03-15
frontend: Rework and extend courses/samples

juliaogris:
Rename samples to courses to align them with UI intent and add new
courses.


juliaogris:
I user acceptance tested the new examples with Mali. 
She got all but the last example right.

camh- (approved):

PR #106, opened 2023-03-16
all: Implement output UI initialisation

juliaogris:
Implement output UI initialisation from evy source code analysis:
* show/hide canvas if there are some/no drawing functions in evy code.
* show/hide read input text field
* show/hide sliders

Update the parser.Program ast node to expose `CalledBuiltinFuncs` which
determine UI changes. Include eventhandler implementation in code analysis.

Note: there is a race condition when navigating back/forwards between
stopping the program and reinstantiating wasmInst for formatting / ui initialisation of new code.
wasmInst is asynchronously set to undefined after the wasm
execution has finished and called afterStop function.

### DIFF
--- a/pkg/wasm/imports.go
+++ b/pkg/wasm/imports.go
@@ -52,9 +52,12 @@ func newSleepingYielder() *sleepingYielder {
 func (y *sleepingYielder) Yield() {
 	y.count++
 	if y.count > 1000 && time.Since(y.start) > 100*time.Millisecond {
-		time.Sleep(minSleepDur)
-		y.Reset()
+		y.ForceYield()
 	}
+}
+
+func (y *sleepingYielder) ForceYield() {
+	y.Sleep(minSleepDur)
 }
 
 func (y *sleepingYielder) Sleep(dur time.Duration) {
--- a/pkg/wasm/main.go
+++ b/pkg/wasm/main.go
@@ -102,7 +102,7 @@ func handleEvents(yielder *sleepingYielder) {
 			yielder.Reset()
 			eval.HandleEvent(event)
 		} else {
-			yielder.Sleep(minSleepDur)
+			yielder.ForceYield()
 		}
 	}
 }
--- a/pkg/wasm/main.go
+++ b/pkg/wasm/main.go
@@ -32,6 +32,9 @@ func main() {
 			setEvySource(formattedInput)
 		}
 	}
+	if actions["ui"] {
+		prepareUI(ast)
+	}
 	if actions["eval"] {
 		// The ast does not correspond to the formatted source code. For
 		// now this is acceptable because evaluator errors don't output
@@ -65,6 +68,13 @@ func parse(input string, rt evaluator.Runtime) (*parser.Program, error) {
 		return nil, parser.TruncateError(err, 8)
 	}
 	return prog, nil
+}
+
+func prepareUI(prog *parser.Program) {
+	funcNames := prog.CalledBuiltinFuncs
+	eventHandlerNames := parser.EventHandlerNames(prog.EventHandlers)
+	names := append(funcNames, eventHandlerNames...)
+	jsPrepareUI(strings.Join(names, ","))
 }
 
 func evaluate(prog *parser.Program, rt *jsRuntime) {
--- a/pkg/wasm/main.go
+++ b/pkg/wasm/main.go
@@ -3,6 +3,8 @@
 package main
 
 import (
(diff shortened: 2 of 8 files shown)


pkg/evaluator/builtin.go
```diff
@@ -21,30 +21,27 @@ type Builtins struct {
 	Print         func(s string)
```
camh-, line 21: It does not look like `Evaluator.print` is used any more, and this `Builtins.Print` is only used to initialise `Evaluator.print`. I think both of these can be removed.

pkg/wasm/exports.go
```diff
@@ -0,0 +1,68 @@
...
+// alloc pre-allocates memory used in string parameter passing.
+//
+//export alloc
+func alloc(size uint32) *byte {
+	buf := make([]byte, size)
+	return &buf[0]
+}
+
+// stop is called when JS/UI interactions triggers stop
+//export stop
```
camh-, line 16: When I save this file in my editor, it runs `goimports` and that adds a blank comment line before `//export stop`, and all the other comments with exports. What command do you run on save, and is it up-to-date? (there were recent changes with the updated go doc changes for 1.19 that reformatted comments, so you need to be up-to-date with your installed goimports.

It seems there is not a make recipe for formatting the Go source in `make tidy`, so CI passes even if code is not formatted.

I install it with:
```
go install golang.org/x/tools/cmd/goimports@master
```

pkg/wasm/imports.go
```diff
@@ -0,0 +1,139 @@
...
+// implementation. The `jsRuntime` struct wraps these functions to
+// implement evaluator.Runtime.
+
+import (
+	"time"
+
+	"foxygo.at/evy/pkg/evaluator"
+)
+
+type jsRuntime struct {
```
camh-, line 16: // jsRuntime implements evaluator.Runtime.

pkg/wasm/imports.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func (y *sleepingYielder) Reset() {
+	y.start = time.Now()
+	y.count = 0
+}
+
+// evySource is imported from JS. The float64 return value encodes the
+// ptr (high 32 bits) and length (low 32 bts) of the source string.
+//export evySource
```
camh-, line 82: Same here and rest of file regarding blank comment lines and `goimports` on save.

pkg/wasm/string.go
```diff
@@ -0,0 +1,45 @@
...
+
+package main
+
+import (
+	"strings"
+	"time"
+	"unsafe"
+)
+
+const minSleepDur = time.Millisecond
```
camh-, line 11: I dont think this belongs in `string.go` - it is used in `main.go` and `imports.go`

camh- (commented):
Some comments...

pkg/wasm/main.go
```diff
@@ -20,44 +20,44 @@ var (
...
 		builtins.Print(err.Error())
 		return
 	}
-	evaluate(source, builtins)
+	formattedInput := ast.Format()
+	if formattedInput != input {
+		setEvySource(formattedInput)
+	}
+
+	evaluate(ast, builtins)
```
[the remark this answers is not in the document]
camh-: oh, i think this is outdated now?

pkg/evaluator/builtin.go
juliaogris: ah cool, right you are, thanks, removed.

pkg/wasm/imports.go
juliaogris: thanks!

pkg/wasm/exports.go
juliaogris: i think I use gofumports, i will investigate.

pkg/wasm/string.go
juliaogris: yes. thank you. Maybe I should have sleepMin() function on the sleepingYielder or maybe forceYield( is the better term?

pkg/wasm/exports.go
juliaogris: ok, I will investigate. it should be made part of local tooling with local GOBIN and part of the build process, do you agree?

juliaogris: ok. gofumpt adds this too. The linter should have found this. that's sad :(

camh- (approved):
:green_salad:

PR #107, opened 2023-03-17
frontend: Extend events sample

juliaogris:
Extend events sample to include animation with key events. While testing
this sample it became obvious that the cmd+enter shortcut to start a
program should remove focus from the editor so that key events get
registered, fixed.

Thanks @fcostin for the idea!


frontend/courses/sample/tour/events.evy
```diff
@@ -0,0 +1,62 @@
...
+dy := 2
+radius := 5
+stopped := false
+speed := 0.5
+hue := 0
+
+dot x y radius hue
+
+on animate
+    if stopped
```
camh-, line 12: This is very tricksy! I was trying to figure out how an evy program actually stopped itself and wondered if there was something magical about this `stopped` variable. Turns out it was just that the run/stop button had focus so it also triggers when you press space. If you move the focus away manually and press space, the dot stops but not the program.

Just commenting.

camh- (approved):
:four_leaf_clover:

frontend/courses/sample/tour/events.evy
juliaogris: Hmm, good point! I always use cmd+enter to start the program these days.
I've added an unfocusRunButton function if we register the key-listener.
Let me know if you like that.

PR #109, opened 2023-03-17
lexer: Fix string tokens for double backslash

juliaogris:
Fix string tokens for 2n backslashes before the closing `"` quotation
mark.


camh- (approved):
\ :green_apple:

PR #113, opened 2023-03-21
evaluator: Refactor Runtime into interface

juliaogris:
Refactor evaluator.Runtime into interface as it appears to be the more
appropriate structure and works well with jsRuntime and cliRuntime. This
also allows for jsRuntime to track its own sleepingYielder.


camh- (approved):
:leafy_green:

PR #114, opened 2023-03-23
frontend: Highlight lines with error

juliaogris:
Highlight lines with evy errors in editor. This is done by finding the
correct token by its line and columns and adding the "err" CSS class to it.
It is becoming increasingly obvious that a more fully fledged web code
editor will be needed. Evy files with more than 700ish lines are starting
to get janky when editing. I could work around this by only updating the
relevant lines, but it all seems to get a bit to hacky.

housekeeping: Do a one-off lint pass with stylelint of CSS(requires node
to which I'm not yet ready to commit.)

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -127,15 +127,15 @@ func newParser(input string, builtins Builtins) *parser {
 		if p.builtins.Globals[fd.Name] != nil {
 			// We still go on to add `fd` to the funcs map so that the
 			// function can be parsed correctly even though it has an invalid name.
-			p.appendErrorForToken("cannot override builtin variable '"+fd.Name+"'", fd.Token)
+			p.appendErrorForToken("cannot override builtin variable '"+fd.Name+"'", fd.token)
 		}
 		switch {
 		case p.funcs[fd.Name] == nil:
 			p.funcs[fd.Name] = fd
 		case builtins.Funcs[fd.Name] == nil:
-			p.appendErrorForToken("redeclaration of function '"+fd.Name+"'", fd.Token)
+			p.appendErrorForToken("redeclaration of function '"+fd.Name+"'", fd.token)
 		case builtins.Funcs[fd.Name] != nil:
-			p.appendErrorForToken("cannot override builtin function '"+fd.Name+"'", fd.Token)
+			p.appendErrorForToken("cannot override builtin function '"+fd.Name+"'", fd.token)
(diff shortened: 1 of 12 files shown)


camh- (approved):
LGTM. just a comment; feel free to ignore

pkg/parser/parser.go
```diff
@@ -556,11 +556,12 @@ func (p *parser) assertArgTypes(decl *FuncDeclStmt, args []Node) {
...
 	}
-	for i := range args {
+	for i, arg := range args {
 		paramType := decl.Params[i].Type()
-		argType := args[i].Type()
+		argType := arg.Type()
 		if !paramType.Accepts(argType) && !paramType.Matches(argType) {
-			p.appendError("'" + funcName + "' takes " + ordinalize(i+1) + " argument of type '" + paramType.String() + "', found '" + argType.String() + "'")
+			msg := fmt.Sprintf("'%s' takes %s argument of type '%s', found '%s'", funcName, ordinalize(i+1), paramType.String(), argType.String())
+			p.appendErrorForToken(msg, arg.Token())
```
camh-, line 564: in my local proof of concept, I went a bit further and also added the token to the two other errors in this function:

```
                        if !paramType.Accepts(argType) && !paramType.Matches(argType) {
-                               p.appendError("'" + funcName + "' takes variadic arguments of type '" + paramType.String() + "', found '" + argType.String() + "'")
+                               p.appendErrorForToken("'"+funcName+"' takes variadic arguments of type '"+paramType.String()+"', found '"+argType.String()+"'", arg.Tok())
                        }

```

and

```
        if len(decl.Params) != len(args) {
-               p.appendError("'" + funcName + "' takes " + quantify(len(decl.Params), "argument") + ", found " + strconv.Itoa(len(args)))
+               tok := p.cur
+               if len(args) > len(decl.Params) {
+                       tok = args[len(decl.Params)].Tok()
+               }
+               p.appendErrorForToken("'"+funcName+"' takes "+quantify(len(decl.Params), "argument")+", found "+strconv.Itoa(len(args)), tok)
                return
        }

```

camh- (approved):
LGTM. just a comment; feel free to ignore...

pkg/parser/parser.go
juliaogris: all addressed. thank you!
(I had a mental note of doing this and then I forgot. busy mind)

PR #115, opened 2023-03-28
frontend: Add code sharing

juliaogris:
Add code sharing via gzip-base64 encoded content in URL hash e.g.

   https://evy.dev#content=H4sIAAAAAAAAEysoyswrUVDyyFTiAgDlkfOoCwAAAA==

This is just a stop gap for storing content server side and and sharing via
shorter ids. Note: sharing does not (yet) show on mobile.

In preparation, fix broken Run button for mobile.


camh- (approved):
:cucumber: seems fine. seems to work too, even better!

PR #116, opened 2023-03-29
frontend: Fix editor buglets

juliaogris:
* Highlight first, not last error token in line
* Fix empty/delete-all editor bug


camh- (approved):
:cucumber:

PR #118, opened 2023-04-16
lexer: Fix empty string lit in middle of line

juliaogris:
Fix empty string lit in middle of line. As we advance at the beginning
of reading string and then again when peekRune is `"` we overshoot for
empty strings `""`. Fix this be doing the outer advance only at the end
of the iteration.

### DIFF
--- a/pkg/lexer/lexer_test.go
+++ b/pkg/lexer/lexer_test.go
@@ -353,3 +353,37 @@ func TestString(t *testing.T) {
 		})
 	}
 }
+
+func TestStringLit(t *testing.T) {
+	in := `fn "" 0`
+	wantTokens := []*Token{
+		{Type: IDENT, Literal: "fn", Offset: 0, Line: 1, Col: 1},
+		{Type: WS, Literal: "", Offset: 2, Line: 1, Col: 3},
+		{Type: STRING_LIT, Literal: "", Offset: 3, Line: 1, Col: 4},
+		{Type: WS, Literal: "", Offset: 5, Line: 1, Col: 6},
+		{Type: NUM_LIT, Literal: "0", Offset: 6, Line: 1, Col: 7},
+	}
+	l := New(in)
+	for _, want := range wantTokens {
+		got := l.Next()
+		assert.Equal(t, want, got)
+	}
+}
+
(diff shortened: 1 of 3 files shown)


pkg/lexer/lexer_test.go
```diff
@@ -353,3 +353,19 @@ func TestString(t *testing.T) {
...
 }
+
+func TestStringLit(t *testing.T) {
+	in := `fn "" 0`
+	wantTokens := []*Token{
+		{Type: IDENT, Literal: "fn", Offset: 0, Line: 1, Col: 1},
+		{Type: WS, Literal: "", Offset: 2, Line: 1, Col: 3},
+		{Type: STRING_LIT, Literal: "", Offset: 3, Line: 1, Col: 4},
+		{Type: WS, Literal: "", Offset: 5, Line: 1, Col: 6},
+		{Type: NUM_LIT, Literal: "0", Offset: 6, Line: 1, Col: 7},
```
camh-, line 364: Since this fix is all about correct advancing, which affects what follows, and errors have their own handling, I think a test for that would be useful too:

```
func TestStringLitErr(t *testing.T) {
        in := `fn "unterminated
"hello"`
        wantTokens := []*Token{
                {Type: IDENT, Literal: "fn", Offset: 0, Line: 1, Col: 1},
                {Type: WS, Literal: "", Offset: 2, Line: 1, Col: 3},
                {Type: ILLEGAL, Literal: "invalid string", Offset: 3, Line: 1, Col: 4},
                {Type: NL, Literal: "", Offset: 16, Line: 1, Col: 17},
                {Type: STRING_LIT, Literal: "hello", Offset: 17, Line: 2, Col: 1},
                {Type: EOF, Offset: 24, Line: 2, Col: 8},
        }
        l := New(in)
        for _, want := range wantTokens {
                got := l.Next()
                assert.Equal(t, want, got)
        }
}
```

You could do these table-driven in the same test function, but i personally like to avoid table driven functions unless there are a lot of similar cases.

camh- (approved):
LGTM

pkg/lexer/lexer_test.go
juliaogris: Nice, thank you! updated

PR #119, opened 2023-04-16
frontend: Reset error highlight on start

juliaogris:
Reset error highlight on start (run), so that the error highlight so
that we accurately display new errors or no errors if they are gone.
Previously errors only got reset if the line was updated which
incorrectly kept error line highlight for follow-on errors, e.g.:

	x := "foo"  // line 1: 'x' declared but not used
	print y     // line 2: unknown variable name 'y'

Fixed as

	x := "foo"  // 💣 error remained: line 1: 'x' declared but not used
	print x     // no error

Now all errors get reset.

### DIFF
--- a/frontend/module/yace-editor.js
+++ b/frontend/module/yace-editor.js
@@ -77,6 +77,7 @@ export default class Yace {
 
     if (
       (value === this.value || value == null) &&
+      Object.keys(this.errorLines).length === 0 &&
       (!errorLines || Object.keys(errorLines).length === 0)
     ) {
       return
--- a/frontend/index.js
+++ b/frontend/index.js
@@ -203,6 +203,7 @@ async function start() {
   stopped = false
   errors = false
   wasmInst = await WebAssembly.instantiate(wasmModule, go.importObject)
+  editor.update({ errorLines: {} })
   clearOutput()
 
   const runButton = document.querySelector("#run")


frontend/module/yace-editor.js
```diff
@@ -77,6 +77,7 @@ export default class Yace {
 
     if (
       (value === this.value || value == null) &&
+      Object.keys(this.errorLines).length === 0 &&
```
camh-, line 80: Isn't this already handled in the next line of the condition? Or is my logic failing me? Either way, I don't really understand this condition. Are these two lines really what is needed to check if an object is empty? Isn't that what this condition is saying? "if the value has not changed or it is null AND there are no errors to highlight then return"?

Even so, this logic is "if foo AND (foo OR bar)" (reordering some things). If foo is true, then bar doesn't matter. If foo is false, then bar doesnt matter. Yeah, something seems off here, but maybe that's me missing something.

juliaogris: the commit message describes the error case - maybe you have a better idea on how to fix it?

essentially i want to say: if the new `value` and the `errLines` are the same then don't do anything....

the issue I had was (and I really need tests):

```
x := "foo"  // line 1: 'x' declared but not used
print y     // line 2: unknown variable name 'y'
```
Fixed as
```
x := "foo"  // 💣 error remained: line 1: 'x' declared but not used
print x     // no error
 ```

juliaogris: ok, sorry for the late response. The difference is `this.errorLines` vs `errorLines`.

juliaogris: This means:
If we haven't had any errorHighlights (`this.errorLines`) and we are also not any new errors (`errorlines`).

camh-: ok. sorry I didn't see the difference between those two. :moneybag:

camh- (approved):

PR #120, opened 2023-04-16
parser: Fix panic on nil token in error

juliaogris:
Fix panic on nil token in error when overriding builtin function names.
This was noted when implementing `clear` which had a different
signature from the hand written clear in the samples. Add panic in
error appending for nil tokens so that we get traceable panics.

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -129,14 +129,12 @@ func newParser(input string, builtins Builtins) *parser {
 			// function can be parsed correctly even though it has an invalid name.
 			p.appendErrorForToken("cannot override builtin variable '"+fd.Name+"'", fd.token)
 		}
-		switch {
-		case p.funcs[fd.Name] == nil:
-			p.funcs[fd.Name] = fd
-		case builtins.Funcs[fd.Name] == nil:
-			p.appendErrorForToken("redeclaration of function '"+fd.Name+"'", fd.token)
-		case builtins.Funcs[fd.Name] != nil:
+		if builtins.Funcs[fd.Name] != nil {
 			p.appendErrorForToken("cannot override builtin function '"+fd.Name+"'", fd.token)
+		} else if p.funcs[fd.Name] != nil {
+			p.appendErrorForToken("redeclaration of function '"+fd.Name+"'", fd.token)
 		}
+		p.funcs[fd.Name] = fd // override anyway so the signature is correct for parsing the function
 	}
 	return p
 }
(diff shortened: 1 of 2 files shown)


pkg/parser/parser.go
```diff
@@ -130,13 +130,12 @@ func newParser(input string, builtins Builtins) *parser {
 			p.appendErrorForToken("cannot override builtin variable '"+fd.Name+"'", fd.token)
 		}
 		switch {
-		case p.funcs[fd.Name] == nil:
-			p.funcs[fd.Name] = fd
-		case builtins.Funcs[fd.Name] == nil:
-			p.appendErrorForToken("redeclaration of function '"+fd.Name+"'", fd.token)
 		case builtins.Funcs[fd.Name] != nil:
 			p.appendErrorForToken("cannot override builtin function '"+fd.Name+"'", fd.token)
+		case p.funcs[fd.Name] != nil && builtins.Funcs[fd.Name] == nil:
```
camh-, line 135: I think maybe you explained last time, but I forget...

This is tautological. From the previous case, we know `builtins.Func[fd.Names]` must be nil - why do you need to check it here? Perhaps you could change this into an `if...else if...` (I seem to remember it was originally and I suggested switching it to a `switch`. Yeah, I'm fickle. sue me :-) : 

```
if builtin.Funcs[fd.Name] != nil {
    p.appendErrorForToken("cannot override builtin function '"+fd.Name+"'", fd.token)
} else if p.funcs[fd.Name] != nil {
    p.appendErrorForToken("redeclaration of function '"+fd.Name+"'", fd.token)
}
p.funcs[fd.Name] = fd // override anyway so the signature is correct for parsing the function
```

Although now I think about it, a function could both try to override a built-in AND redeclare a function. Perhaps that `else` should not be there at all, and you generate two errors in this case?

camh- (approved):
:broccoli: Looks good. My comment is just a suggestion - take it or leave it.

pkg/parser/parser.go
juliaogris: hmmm - I'm not sure which one is better. one message per one error or all (maybe the second one is the conceptually bigger one for the user? or maybe sometimes only the second makes sense?

juliaogris: p.funcs contains all builtin.Funcs.
This is the reason why I had `p.funcs[fd.Name] != nil && builtins.Funcs[fd.Name] == nil`
to mean _not_ a builtin funcs, but and explicitly declared one.

juliaogris: i think i had the tautological ` p.funcs[fd.Name] != nil && builtins.Funcs[fd.Name] == nil` in case you re-arrange the if-else block. if you do it without the extra term it will be wrong. and I did that before and it stung me. so i put it back in now.

camh-: I think one error - "cannot override built-in" is sufficient. If they define the same function twice that overrides a built-in, they just get this error for each.

juliaogris: ok. so what about keeping the tautological condition?

camh-: well, they don't make sense to me - i see it and just think WTF - logically this makes no sense, am I missing something? The current if/elseif reads well to me and it is clear. But up to you - whatever makes sense for you

PR #121, opened 2023-04-18
all: Cleanup lexing, comments and sample

juliaogris:
This PR continains various minor cleanups:

- lexing: Make empty string lexing consistent and clearer in JS and go
- index.js: Group exported funcs, fix comments, remove debug prints
- improve "guess my number" sample


pkg/lexer/lexer.go
```diff
@@ -191,15 +191,15 @@ func (l *Lexer) readIdent() string {
 
 func (l *Lexer) readString() (string, error) {
 	pos := l.pos
-	backslashCnt := 0
+	escaped := false
 	for {
 		if l.cur == '\\' {
```
camh-, line 196: you could get "fancy" and just say `escaped = l.cur == '\\' && !escaped`. Perhaps that is a little too obtuse? (same in the JS too)

camh- (approved):
:dragon:

pkg/lexer/lexer.go
juliaogris: yeah, i think even though this is a little "fancy" it is worth doing for the reduction in code size. updated.

PR #122, opened 2023-04-18
evy: Add `clear` to drawing functions

juliaogris:
Add `clear` to drawing functions as a builtin because it is very
commonly used. Use an optional second parameter to specify the clear or
background colour.


camh- (approved):
:green_square:

PR #123, opened 2023-04-23
evy: Add drawing style functions

juliaogris:
Add drawing style functions stroke, fill, dash, and linecap defined in
the same terms as HTML canvas styling, with the addition of `"none"`
for stroke and fill.

New examples:

- https://evy-lang--123-x8t7erj2.web.app/#linestyle
- https://evy-lang--123-x8t7erj2.web.app/#fill

### DIFF
--- a/frontend/index.js
+++ b/frontend/index.js
@@ -55,6 +55,11 @@ function newEvyGo() {
     rect,
     color,
     clear,
+    // advanced canvas
+    stroke,
+    fill,
+    dash,
+    linecap,
   }
   const go = new Go() // see wasm_exec.js
   go.importObject.env = Object.assign(go.importObject.env, evyEnv)
@@ -77,7 +82,20 @@ function jsPrepareUI(ptr, len) {
 }
 
 function needsCanvas(f) {
-  return f.move || f.line || f.width || f.circle || f.rect || f.color || f.colour
+  return (
+    f.move ||
+    f.line ||
+    f.width ||
+    f.circle ||
+    f.rect ||
+    f.color ||
+    f.colour ||
+    f.clear ||
+    f.stroke ||
+    f.fill ||
+    f.dash ||
+    f.linecap
+  )
 }
 
 // jsPrint converts wasmInst memory bytes from ptr to ptr+len to string and
@@ -373,6 +391,8 @@ function newCanvas() {
     scale: { x: 1, y: -1 },
     width: 100,
     height: 100,
+    fill: true,
+    stroke: true,
 
     offset: { x: 0, y: -100 }, // height
   }
@@ -392,6 +412,10 @@ function resetCanvas() {
   ctx.fillStyle = "black"
   ctx.strokeStyle = "black"
   ctx.lineWidth = 1
+  canvas.fill = true
+  canvas.stroke = true
+  ctx.lineCap = "butt"
+  ctx.setLineDash([])
   move(0, 0)
 }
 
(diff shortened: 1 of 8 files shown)


pkg/wasm/imports.go
```diff
@@ -34,7 +37,22 @@ func (rt *jsRuntime) Circle(r float64)           { circle(r) }
...
+func (rt *jsRuntime) Fill(s string)              { fill(s) }
+func (rt *jsRuntime) Dash(segments []float64)    { dash(floatsToString(segments)) }
+func (rt *jsRuntime) Linecap(s string)           { linecap(s) }
+
+func floatsToString(floats []float64) string {
+	if len(floats) == 0 {
+		return ""
+	}
+	var sb strings.Builder
+	sb.WriteString(fmt.Sprintf("%f", floats[0]))
```
camh-, line 50: you could just range over all floats and then return `sb.String()[1:]` to strip off the space rather than treating the first one specially here. just an idea

frontend/index.js
```diff
@@ -476,6 +504,33 @@ function clear(ptr, len) {
...
+function fill(ptr, len) {
+  const s = memToString(ptr, len)
+  canvas.fill = s !== "none"
+  canvas.ctx.fillStyle = s
+}
+
+// dash is exported to evy go/wasm.
+function dash(ptr, len) {
+  const s = memToString(ptr, len)
+  const nums = s.split(" ").map(Number).map(transformX)
```
camh-, line 524: :pineapple:  oooh, look at you being all functional and all :sun_with_face:

pkg/evaluator/builtin.go
```diff
@@ -155,16 +166,24 @@ func (rt *UnimplementedRuntime) Unimplemented(s string) {
...
+func (rt *UnimplementedRuntime) Move(x, y float64)       { rt.Unimplemented("move") }
+func (rt *UnimplementedRuntime) Line(x, y float64)       { rt.Unimplemented("line") }
+func (rt *UnimplementedRuntime) Rect(x, y float64)       { rt.Unimplemented("rect") }
+func (rt *UnimplementedRuntime) Circle(r float64)        { rt.Unimplemented("circle") }
+func (rt *UnimplementedRuntime) Width(w float64)         { rt.Unimplemented("width") }
+func (rt *UnimplementedRuntime) Color(s string)          { rt.Unimplemented("color") }
+func (rt *UnimplementedRuntime) Clear(color string)      { rt.Unimplemented("clear") }
+func (rt *UnimplementedRuntime) Stroke(s string)         { rt.Unimplemented("stroke") }
+func (rt *UnimplementedRuntime) Fill(s string)           { rt.Unimplemented("fill") }
+func (rt *UnimplementedRuntime) Text(s string)           { rt.Unimplemented("text") }
```
camh-, line 181: I think you've accidentally snuck in some bits of the next PR (i.e. Text et al are not implemented yet, although technically thats exactly what this function says).

camh- (approved):
LGTM, but I have no idea how these new functions are called - e.g. what parameters does `linecap` take and what do they mean. Should this get documented somewhere? I guess so, but documentation is boring, so ok.

pkg/evaluator/builtin.go
juliaogris: fixed, thanks.

pkg/wasm/imports.go
juliaogris: ok done.

juliaogris:
> LGTM, but I have no idea how these new functions are called - e.g. what parameters does `linecap` take and what do they mean. Should this get documented somewhere? I guess so, but documentation is boring, so ok.

Yes, you are right. There are lots of undocumented builtins now. This needs a separate pass. (hope that's ok with you).

PR #124, opened 2023-04-23
evy: Add canvas text builtins

juliaogris:
Add canvas text builtins `text`, `textsize`, `fontfamily`, and `font`.
Add new example #text. These new primitives are only concerned with
text and text styles on the evy output canvas.

### DIFF
--- a/frontend/courses/sample/draw/text.evy
+++ b/frontend/courses/sample/draw/text.evy
@@ -1,0 +1,17 @@
+move 10 80
+text "Hello"
+
+move 10 60
+textsize 10
+text "all 🖖"
+
+move 10 40
+fontfamily "Luminari, fantasy"
+text "Cheriolas!"
+
+move 10 35
+fill "grey"
+// font maps directly to the CSS font property; see:
+// https://developer.mozilla.org/en-US/docs/Web/CSS/font
+font "italic max(2rem, 2vh) \"Comic Sans MS\", cursive"
+text "(Cheering Charm)"
--- a/frontend/courses/courses.json
+++ b/frontend/courses/courses.json
@@ -24,7 +24,8 @@
         { "id": "squares", "title": "Squares" },
         { "id": "rainbow", "title": "Rainbow" },
         { "id": "fill", "title": "Stroke and Fill" },
-        { "id": "linestyle", "title": "Line Styles" }
+        { "id": "linestyle", "title": "Line Styles" },
+        { "id": "text", "title": "Text" }
       ]
     },
     {
--- a/frontend/index.js
+++ b/frontend/index.js
@@ -60,6 +60,10 @@ function newEvyGo() {
     fill,
     dash,
     linecap,
+    text,
+    textsize,
+    fontfamily,
+    font,
   }
   const go = new Go() // see wasm_exec.js
   go.importObject.env = Object.assign(go.importObject.env, evyEnv)
@@ -94,7 +98,11 @@ function needsCanvas(f) {
     f.stroke ||
     f.fill ||
     f.dash ||
-    f.linecap
+    f.linecap ||
+    f.text ||
+    f.textsize ||
+    f.fontfamily ||
+    f.font
   )
 }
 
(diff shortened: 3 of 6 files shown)


frontend/courses/sample/draw/text.evy
```diff
@@ -0,0 +1,15 @@
+move 10 80
+text "Hello"
+
+move 10 60
+textsize 10
+text "all 🖖"
+
+move 10 40
+fontfamily "Luminari, fantasy"
```
camh-, line 9: I dont have this font. Is this a mac thing? Perhaps you should stick to common web fonts?

frontend/courses/sample/draw/text.evy
```diff
@@ -0,0 +1,15 @@
...
+text "Cheriolas!"
+
+move 10 35
+fill "grey"
+font "italic max(2rem, 2vh) \"Comic Sans MS\", cursive"
```
camh-, line 14: :pineapple: I dont understand this `max(2rem, 2vh)` thing - I know it's from the canvas or web or something, but i feel whatever it is is going against the simplicity you're going for with evy (I thought the same with the colour definitions too, but I know you felt differently - still feel I need to say it though).

I've always though of font sizes in points - that is the sizing that is pretty much universal when it comes to text. `px`, `rem`, `vh`, etc all seem to be webby things and I dont know what they mean. Should you be expecting leaning programmers to know?

camh- (approved):

frontend/courses/sample/draw/text.evy
juliaogris: fantasy is the fallback. I searched for "common web fantasy fonts" and this is what came up, e.g.: https://blog.hubspot.com/website/web-safe-html-css-fonts or https://www.w3.org/Style/Examples/007/fonts.en.html

What does it display for you?

frontend/courses/sample/draw/text.evy
juliaogris: I know you have point.
For me all the recent additions are "advanced".
A lot of things can be done with `move`, `line`, `rect`, `text`, `textsize`.
However, if there is more nuance desired this is what you can use.
Maybe that's a mistake, maybe I will get rid of it again.
I have noticed however, that sometimes for Mali at least it is nice to consolidate knowledge without learning "hard" new things. So if she can draw circles it might be fun to challenge her with an ellipse, without requiring abstraction / higher order data structures or even the understanding of functions. Maybe all of this is a mistake. Not sure. I think most text things can be done with `text`, `textsize` and `fontfamily` (I added this separately because of your suggestion even though it is a little hacky). However, if you want to do bold and/or cursive, `font` on canvas is the way.

Maybe I should add a comment on where to read up on it?

juliaogris: It's again: make easy things easy but hard stuff possible.

juliaogris: But also to more properly answer: I believe there are layers of progression to evy - we have discussed this before. Start with simple drawing / printing to given an intuition on how code is executed sequentially. Add basic variables, while loops, reading input, if's. once that understanding is solid: functions, arrays. Lastly functions with parameters and return values, maps, recursion, list trees etc.

My feeling is, giving enough drawing primitives to write games is desirable, e.g.: https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Drawing_shapes#making_combinations

Imo that includes curves and flexible fonts but I reserve the rights to change my mind.

juliaogris: Last `max(2rem, 2vh)` is a CSS thing. it is literally the maximum of `2rem and 2vh` (2*root element size; by default a rem is 16px in most browser vs. 2*viewport height - viewport height is the browser window height). 
This are things to be read up on at 

- https://developer.mozilla.org/en-US/docs/Web/CSS/font and 
- https://developer.mozilla.org/en-US/docs/Web/CSS/font-size

frontend/courses/sample/draw/text.evy
camh-: I dont see the font changing at all. Just the same as the previous line (although it's a bit hard to tell for sure because the previous line is just "all", which is not much to compare with).

juliaogris: That's kinda sad.
This is what it looks like for me:
Screenshot 2023-04-24 at 1 02 22 PM

PR #125, opened 2023-04-24
evy: Add canvas poly builtin

juliaogris:
Add canvas poly builtin to draw polylines and polygons.

New sample:
https://evy-lang--125-ofi7tbe4.web.app/#poly

### DIFF
--- a/frontend/module/yace-editor.js
+++ b/frontend/module/yace-editor.js
@@ -467,7 +467,6 @@ const builtins = new Set([
   "rand",
   "rand1",
   "move",
-  "read",
   "line",
   "rect",
   "circle",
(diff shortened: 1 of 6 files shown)


frontend/module/yace-editor.js
```diff
@@ -467,14 +467,14 @@ const builtins = new Set([
   "rand",
   "rand1",
   "move",
-  "read",
```
camh-, line 470: read is still a builtin, isn't it?

camh- (approved):
LGTM, even if you do insist on calling it poly.

frontend/module/yace-editor.js
juliaogris: yeah, it was a duplicate. I'll break it out into a separate commit (ie should be deleted).

juliaogris:
> LGTM, even if you do insist on calling it poly.

Happy to call it something short and better if you can think of anything.

PR #129, opened 2023-05-17
evaluator: Implement short-circuit evaluation of AND/OR

camh-:
Implement short-circuiting of the AND and OR operators so that the
right-hand side is not evaluated if the result can be determined by the
left-hand side. For AND, this means left == false, since the whole
expression will be false regardless of what right is. For OR, this means
left == true, for the same reason.

Update the spec to say these operators perform short-circuit evaluation.

Add a Towers of Hanoi evy program with a new "Games" section in the
course. This program uses short-circuit evaluation to protect against
indexing an array out of bounds, hence its presence with this change.

### DIFF
--- a/docs/syntax_grammar.md
+++ b/docs/syntax_grammar.md
@@ -473,10 +473,12 @@ division and the [modulo operator]. `+` may also be used as
 concatenation operator for `string` and `array` types. 
 
 Boolean operators `and`, `or` stand for [logical conjunction (AND)] and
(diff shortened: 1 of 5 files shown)


juliaogris (approved):
Really lovely, all of it - can't find anything that should be changed. Thanks for doing this.
Maybe consider adding the Towers of Hanoi sample as a separate commit? It is really cool and was what prompted this fix.... 🍪

juliaogris (approved):
Looks still great to me. It's so addictive playing it. 💚

PR #130, opened 2023-05-18
docs: Clean up grammar spacing, etc

camh-:
Clean up the spacing of the grammar and add missing period production
terminators. This changes nothing of substance, just cleanliness.


juliaogris (approved):
Thank you. 🍵
I'm about to shower to stick with the theme (also cause i did a three hour pottery class with Abuja)

PR #131, opened 2023-05-20
frontend: Clean-up samples

juliaogris:
This is mostly a housekeeping PR around samples, removing course material
and rearranging sample sections.

Reclassify and update same samples. Remove all reference to courses as
we are doing courses in the wiki now and they'll get a different UI in
the future.
While at it, add some minor style fixes and fix the nav menu bug.


camh- (approved):
:rainbow:

PR #132, opened 2023-05-26
frontend: Capture pointerleave event on canvas

juliaogris:
Capture pointerleave event on canvas by using exact values `0` or `100`
for `x` or `y` to use as condition to stop animations, for example in
the interactive ellipse 🍉 sample. This addition was in response to
watching Mali play with the example and being confused with what it did
when leaving the mouse left the canvas.


camh- (approved):
:leaves:

PR #134, opened 2023-05-31
main: Drop trailing newline in CLI read function

camh-:
The CLI runtime `Read()` function was returning the line read with the
trailing newline. This is different to how the browser/wasm
implementation works which does not include the trailing newline. This
makes for different behaviour when running an evy program with the
different runtimes.


juliaogris (approved):
💚 Thank you!

PR #136, opened 2023-06-01
all: Fix various buglets

juliaogris:
This is another housekeeping PR fixing various little issues throughout
the code base that were found when working on the `tictactoe` sample:

- Fix lost animated run button
- Add error checking for rand argument range
- Fix `read` with empty string
- Add no-op yielder in evy CLI


camh- (approved):
:leafy_green:

PR #137, opened 2023-06-03
docs: Fix grammar of for loops

camh-:
Fix the grammar docs to specify that the variable declaration is
optional. This removes the assignment option as that is not implemented,
nor is it planned to be implemented.

### DIFF
--- a/docs/syntax_grammar.md
+++ b/docs/syntax_grammar.md
@@ -104,7 +104,7 @@ not allowed.
     for_stmt   = "for" range NL
                     statements
                  "end" NL .
-    range      = ident ( ":=" | "=" ) "range" range_args .
+    range      = [ ident ":=" ] "range" range_args .
     range_args = <- expr -> [ <- expr -> [ <- expr -> ] ] .
     while_stmt = "while" toplevel_expr NL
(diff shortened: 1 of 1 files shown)


docs/syntax_grammar.md
```diff
@@ -104,7 +104,7 @@ not allowed.
     for_stmt   = "for" range NL
                     statements
                  "end" NL .
-    range      = ident ( ":=" | "=" ) "range" range_args .
+    range      = [ ident ( ":=" | "=" ) ] "range" range_args .
```
juliaogris, line 107: It seem like `=` is not allowed, this does not compile:
```
i:num
for i = range 3 
    print i
end
```

juliaogris (approved):
Definitely an improvement, maybe also remove `"="` ?
(Or fix parser?)

sorry, you always get the trivial not so trivial ones...

docs/syntax_grammar.md
camh-: ok. I have removed that from the grammar, and (you're not going to like this) added a TODO to allow assignment in for loops. :moneybag:

PR #138, opened 2023-06-03
eval: Fix break to not break more than one level

camh-:
Fix the evaluation of the `break` statement so as to not break out of
more than one loop (`for` or `while`). We do this by turning a `Break`
`Value` into a `nil` `Value` when it is handled, breaking a loop.
Further outer loops will no longer break out.

Add tests for this case.

### DIFF
--- a/pkg/evaluator/evaluator_test.go
+++ b/pkg/evaluator/evaluator_test.go
@@ -168,6 +168,22 @@ while true
     end
     continue = false
 end
+`, `
+while true
+    while true
+        break
+    end
+    print "🎈"
+    break
+end
+`, `
+for range 3
+    for range 3
+        break
+    end
+    print "🎈"
+    break
+end
 `,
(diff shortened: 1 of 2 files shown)


pkg/evaluator/evaluator_test.go
```diff
@@ -168,6 +168,22 @@ while true
     end
     continue = false
 end
+`, `
+while true
+    while true
+	break
```
juliaogris, line 174: Needs one more indent?

juliaogris (approved):
lgtm modulo evy code indentation.

pkg/evaluator/evaluator_test.go
camh-: Ah, it was a tab, so it looked ok to me, but it looked 4 spaces to you. This language within language stuff is annoying.

PR #139, opened 2023-06-09
go: Use fmt.Sprintf where sensible

juliaogris:
Use fmt.Sprintf where sensible - it was used before with the assumption
that tinygo doesn't support it (runtime error), but it turns out this
was only because we didn't use main() as mount point. While at it,
change all single to double quotes.

### DIFF
--- a/pkg/evaluator/builtin.go
+++ b/pkg/evaluator/builtin.go
@@ -231,11 +231,11 @@ func printFunc(printFn func(string)) BuiltinFunc {
 func printfFunc(printFn func(string)) BuiltinFunc {
 	return func(_ *scope, args []Value) (Value, error) {
 		if len(args) < 1 {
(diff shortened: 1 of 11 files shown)


camh- (approved):
:broccoli: LGTM. Seems you got all the single quotes from what I can tell. I'm not sold on using printf everywhere - it is more overhead for arguably simple stuff, but most of that seems to be in the parser which is not the evaluation hot path. So fine by me if that's what you prefer.

juliaogris:
> 🥦 LGTM. Seems you got all the single quotes from what I can tell. I'm not sold on using printf everywhere - it is more overhead for arguably simple stuff, but most of that seems to be in the parser which is not the evaluation hot path. So fine by me if that's what you prefer.

Yeah, thought about that too and i left the consternation in some place where it was just 2 strings iirc

PR #140, opened 2023-06-10
parser: Move scope stack to parser state

juliaogris:
Refactor parser private function order a little and break up newParser (bit long).

Move scope stack into parser state rather than passing it
around as parameters.


camh- (approved):
:leafy_green: LGTM. I scanned it, and it seemed fine... ok. i checked it out and grepped for scope and seems complete.

juliaogris:
> 🥬 LGTM. I scanned it, and it seemed fine... ok. i checked it out and grepped for scope and seems complete.

Thanks for being so thorough! 🍪👗

PR #141, opened 2023-06-12
evy: Rework `font`

juliaogris:
Rework `font` function to take a map, rather than a string that matches
less closely to the CSS font property but is hopefully a bit more intuitive
to use. Remove `textsize` and `fontfamily` along the way. `font` usage with
exhaustive example:

    font {
       size:4
       style:"italic" // "normal"
       family:"Comic Sans MS, cursive"
       letterspacing:2 // extra inter-character space. negative allowed. default:0
       align:"center" // "left", "right"
       baseline:"middle" // "top", "bottom", "alphabetic" (default)
    }

In preparation add a couple of tweaks in tictactoe that popped up when
watching Anika play and pull out evaluator.Runtime interface and
utilities into separate file evaluator/runtime.go as builtins.go is
getting rather big.


pkg/evaluator/builtin.go
```diff
@@ -597,6 +595,71 @@ func dashFunc(dashFn func([]float64)) BuiltinFunc {
 	}
 }
 
+var fontDecl = &parser.FuncDeclStmt{
+	Name:       "font",
+	Params:     []*parser.Var{{Name: "properties", T: parser.GENERIC_MAP}},
```
camh-, line 600: is `GENERIC_MAP` correct here? I thought that was only for literals. For a formal argument list, I would have thought this should be `&Type{Name: MAP, Subtype: ANY_TYPE}`

pkg/evaluator/builtin.go
```diff
@@ -597,6 +595,71 @@ func dashFunc(dashFn func([]float64)) BuiltinFunc {
...
+		"align":         "string",
+		"letterspacing": "num",
+	}
+	for key, val := range arg.Pairs {
+		propType, ok := propTypes[key]
+		if !ok {
+			return nil, fmt.Errorf("%w: unknown property %q", ErrBadArguments, key)
+		}
+		anyVal, ok := val.(*Any)
+		if ok {
```
camh-, line 621: nit: I'd combine these lines because `anyVal` and `ok` are not needed after this:

    if a, ok := val.(*Any); ok {
        val = a.Val
    }

camh- (commented):
The commit message of the "Rework font function" commit is not wrapped at 72 chars, and the indenting is done with tabs not spaces, so it is extra indented (you would not notice as you have tab == 4, but at tab == 8, it is rather wide.

pkg/evaluator/builtin.go
juliaogris: If I don't put this in I cant use:

```
font {size:10} 
```

as it is interpreted as `{}num`.
So yes, I believe this is as good as it gets for what I want here.

pkg/evaluator/builtin.go
juliaogris: ok, agreed.

juliaogris:
> The commit message of the "Rework font function" commit is not wrapped at 72 chars, and the indenting is done with tabs not spaces, so it is extra indented (you would not notice as you have tab == 4, but at tab == 8, it is rather wide.

Apologies, fixed.

pkg/evaluator/builtin.go
juliaogris: done.

pkg/evaluator/builtin.go
camh-: I see. This seems to be a side-effect of how GENERIC works. so I guess this is fine for now, but I think the correct way is for `{size:10}` as a literal to remain GENERIC somehow so it can match either `{}num` or `{}any`. It is what we do for the empty composite literal, perhaps we need to do that for others?

juliaogris: yes agreed!
but maybe not as part of this PR.

camh- (approved):
:green_book:

PR #142, opened 2023-06-13
eval: Replace ValueType with *parser.Type

camh-:
Remove the `ValueType` type in favour of `*parser.Type` to enable deeper 
type knowledge of composite types (arrays, maps) that do not have values.
The `ValueType` type only modelled an array or map but not the type of its
contents. If the composite value was empty, it was not possible to tell
what the full composite type was. This makes it impossible to properly do
type assertions.

An `Array` and `Map` now need to store the actual parser type within the 
value. The simple types can just return the parser "constants" for those 
types without needing any extra storage.

This adds a `None` `Value` implementation that is returned instead of
`nil` when an evaluation does not result in a value, making it a little
safer to use everywhere. One place, `evalConditionalBlock` still does
return a `nil` `Value`, but that special case is handled in all the
places that call it and ensure that a real `Value` is returned past
those points.

### DIFF
--- a/pkg/evaluator/builtin.go
+++ b/pkg/evaluator/builtin.go
@@ -315,7 +315,7 @@ func splitFunc(_ *scope, args []Value) (Value, error) {
 	for i, s := range slice {
 		elements[i] = &String{Val: s}
 	}
-	return &Array{Elements: &elements}, nil
+	return &Array{Elements: &elements, T: stringArrayType}, nil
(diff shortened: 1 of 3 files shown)


juliaogris (approved):
🟢 lgtm
even without the pending typeassertion / typeof func this seems like a net improvement to unify the type.
Thank you!

PR #144, opened 2023-06-13
parser/eval: Implement type assertions and typeof builtin

camh-:
Implement type assertions as defined in the spec in the parser as a new
AST node type and in the evaluator to evaluate that AST node. Implement
the `typeof` builtin so that type assertions can be used safely and robustly
over a range of types.

Update the spec to make `typeof` more specified in the value returned for
arrays and maps, and cover the special case of empty array and map
literals that do not yet have a subtype.

A couple of small bugs have been fixed were AST nodes for literals were
not fully constructed, and where types were not inferred deeply enough
in the case of array and map elements of arrays and maps.

### DIFF
--- a/pkg/parser/ast.go
+++ b/pkg/parser/ast.go
@@ -712,20 +712,20 @@ func newlineList(nodes []Node) string {
 	return strings.Join(lines, "\n") + "\n"
 }
 
-func zeroValue(t TypeName) Node {
-	switch t {
+func zeroValue(t *Type, tt *lexer.Token) Node {
+	switch t.Name {
 	case NUM:
-		return &NumLiteral{Value: 0}
+		return &NumLiteral{Value: 0, token: tt}
 	case STRING:
-		return &StringLiteral{Value: ""}
+		return &StringLiteral{Value: "", token: tt}
 	case BOOL:
-		return &Bool{Value: false}
+		return &Bool{Value: false, token: tt}
 	case ANY:
-		return &Bool{Value: false}
+		return &Bool{Value: false, token: tt}
 	case ARRAY:
-		return &ArrayLiteral{}
+		return &ArrayLiteral{T: t, token: tt}
 	case MAP:
-		return &MapLiteral{}
+		return &MapLiteral{T: t, token: tt}
 	}
 	return nil
 }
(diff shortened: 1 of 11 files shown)


pkg/parser/ast.go
```diff
@@ -394,6 +400,18 @@ func (d *DotExpression) Type() *Type {
 	return d.T
 }
 
+func (t *TypeAssertion) Token() *lexer.Token {
+	return t.token
+}
+
+func (t *TypeAssertion) String() string {
+	return "(" + t.Left.String() + "." + "(" + t.T.String() + ")" + ")"
```
camh-, line 408: Hmmm, should I be using `Sprintf` here now?

juliaogris: Up to you, I think this is fine.

pkg/parser/expression.go
```diff
@@ -269,6 +271,34 @@ func (p *parser) parseDotExpr(left Node) Node {
...
+	case ILLEGAL_TYPE:
+		p.appendErrorForToken("invalid type in type assertion of '"+left.String()+"'", tok)
+	case ANY_TYPE:
+		p.appendErrorForToken("cannot type assert to type 'any'", tok)
+	}
+	if p.assertToken(lexer.RPAREN) {
+		p.advance() // advance past )
+	}
+	if left.Type() != ANY_TYPE {
+		p.appendErrorForToken("value of type assertion must be of type 'any', not '"+left.Type().String()+"'", tok)
```
juliaogris, line 297: I think we don't use single quotes anymore and i never quotes types, only var and func names.

juliaogris (approved):
💚 

Awesome Cam, especially all the little cleanups (tokens for literals) on the side. Thank you so much!!

This looks complete and well tested.

I wonder - if there ever a case where we want type assert anything other than a var / param of type any? (probably not, just wanted to ask one more time).

camh-:
> if there ever a case where we want type assert anything other than a var / param of type any?

I can't think of one. `any` is the only dynamic type in the language. Type asserting a concrete type doesn't make much sense - it is already what you are trying to make of it when you do the assertion.

juliaogris (approved):
Still great! 💚

PR #145, opened 2023-06-16
ci: Support auto-versioning of major/minor versions

camh-:
Update the `NEXTTAG_CMD` command to identify when a release note has
been added in the latest master commit in `docs/release-notes/v*.md` and
extract the version from the filename. If that version is later than
what `NEXTTAG_CMD` would have used by default, use that version number
for the tag instead.

This allows creating a commit/PR with a new release note and have that
version get auto-tagged and released when merged to master.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -148,10 +148,13 @@ nexttag:
 .PHONY: nexttag release
 
 define NEXTTAG_CMD
-{ git tag --list --merged HEAD --sort=-v:refname; echo v0.0.0; }
-| grep -E "^v?[0-9]+.[0-9]+.[0-9]+$$"
-| head -n1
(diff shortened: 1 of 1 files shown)


juliaogris (approved):
That looks really neat and simple, I like it!!

If it is not too hard, I think I would add a `v\d+.\d+.\d+.md` check, but if it is, I think this is just fine.

Thank you!

🧑‍🎤

camh-:
> If it is not too hard, I think I would add a `v\d+\.\d+\.\d+\.md` check, but if it is, I think this is just fine.

@juliaogris Done. I fixed the grep regex while I was there - the dots were not escaped. PTAL

PR #146, opened 2023-06-16
evy: Add `cls` function

juliaogris:
Add `cls` function to clear the screen. It might be useful in text based
games.

Sneak in a slight sample file re-org.


main.go
```diff
@@ -37,6 +39,17 @@ func (*cliRuntime) Print(s string) {
 	fmt.Print(s)
 }
 
+func (*cliRuntime) Cls() {
+	cmd := exec.Command("clear")
+	if runtime.GOOS == "windows" {
+		cmd = exec.Command("cmd", "/c", "cls")
```
camh-, line 45: :pineapple: So very diligent!

camh- (approved):
I'm fine with the name `cls` - as you can see from the windows command you implemented, it has some history, actually going back to BASIC. But if you do want to change it, that's fine too.

:kiwi_fruit:

PR #150, opened 2023-06-20
docs: Add builtins.md

juliaogris:
Add builtins.md with docs for I/O builtin functions. In follow up
commits we will extend builtins.md with the following sections:

- Types
- Map
- Program Control
- Conversion
- String
- Random
- Math
- Error
- UI
- Event Handlers

Add a `doctest` target to the Makefile run with `make ci` checking for
non-dirty git status of the relevant markdown files.

Update github CI environment with the environment variable `TERM=vt100`.
Without this environment variable set `cls` fails and without that we
cannot run doctest on CI.

---

Rendered markdown:
https://github.com/foxygoat/evy/blob/builtin-docs/docs/builtins.md

Current WIP:
https://github.com/foxygoat/evy/blob/builtin-types/docs/builtins.md

### DIFF
--- a/docs/builtins.md
+++ b/docs/builtins.md
@@ -1,0 +1,237 @@
+# Builtins
+
+Evy provides built-in functions and events that allow for user
+interaction, graphics, animation, mathematical operations, and more.
+
+Functions are self-contained blocks of code that perform a specific
+task. Events are notifications that are sent to a program when
+something happens, such as when a user moves the mouse or presses a
+key.
+
+## Input and Output
+
+### `print`
+
+`print` prints the arguments given to it to the output area. It separates them by
+a single space and outputs a newline character at the end.
+
+#### Example 
+
+```evy
+print "Hello"
+print 2 true "blue"
+print "array:" [1 2 3]
+print "map:" {name:"Scholl" age:21}
+```
+Output
+```evy:output
+Hello
+2 true blue
+array: [1 2 3]
+map: {name:Scholl age:21}
+```
+
+#### Reference 
+
+    print a:any...
+
+The `print` function prints its arguments to the output area, each
+separated by a single space and terminated by a newline character. If
+no arguments are provided it only prints the newline character. 
+
+The backslash character `\` can be used to represent special characters
+in strings. For example, the `\t` escape sequence represents a tab
+character, and the `\n` escape sequence represents a newline character.
+Quotes in string literals must also be escaped with backslashes,
+otherwise they will be interpreted as the end of the string literal.
+For example: 
+
+```evy
+print "Here's a tab: 👉\t👈\nShe said: \"Thank you!\""
+```
+
+Output
+```evy:output
+Here's a tab: 👉	👈
+She said: "Thank you!"
+```
+
+In a browser environment `print` outputs to the output area. When
+running Evy from the command line interface, `print` prints to standard
+out.
+
+### `read`
+ 
+`read` reads a line of input from the user and returns it as a
+string. The newline character is not included in the returned string.
+
+#### Example
+```evy
+name := read
+print "Hello, "+name+"!"
+```
+
+Input
+```evy:input
+Mary Jackson
+```
+
+Output
+```evy:output
+Hello, Mary Jackson!
+```
+
+#### Reference 
+
+    read:string
+
+The `read` function returns a string that contains the line of input
+that the user entered up until, excluding the newline character. It is
+a blocking functions, which means that it will not return until the
+user has entered a line of input and pressed the Enter key.
+
+In a browser environment `read` reads from the text input area. When
+running Evy from the command line interface, `read` reads from standard
+in.
+
+### `cls`
+
+`cls` clears the output area of all printed text.
+
+#### Example
+```evy
+print "Hello"
+sleep 1
+cls
+print "Bye"
+```
+
+Output
+```evy:output
+Bye
+```
+
+#### Reference 
+
+    cls
+
+The `cls` function clears all text output. In a browser environment
+`cls` clears the output area. When running Evy from the command line
+interface, `cls` clears the terminal, similar to the Unix `clear` or
+Windows `cls` commands.
+
+### `printf`
+
+`printf` stands for print formatted.
+
+`printf` prints its arguments to the output area according to a _format_
+string. The format string is the first argument, and it
+contains _specifiers_. Specifiers start with a percent sign `%`. They
+tell the `printf` function how and where to print the remaining
+arguments inside the format string. The rest of the format string is
+printed to the output area without changes. 
+
+Here are some valid specifiers in Evy:
+
+| Specifier | Description |
+| --------- | ------------- |
+| `%v`      | the argument in its default format  |
+| `%q`      | a double-quoted string |
+| `%%`      | a percent sign `%`  |
+
+#### Example
+
+```evy
+printf "The tank is 100%% full.\n\n"
+
+weather := "rainy"
+printf "It is %v today.\n" weather
+rainfall := 10
+printf "There will be %vmm of rainfall.\n" rainfall
+unicorns := false
+printf "There will be unicorns eating lollipops: %v.\n\n" unicorns
+
+quote := "Wow!"
+printf "They said: %q\n" quote
+printf "Array: %v\n" [1 2 3]
+printf "Map: %v\n" {a:1 b:2}
+```
+
+Output
+```evy:output
+The tank is 100% full.
+
+It is rainy today.
+There will be 10mm of rainfall.
+There will be unicorns eating lollipops: false.
+
+They said: "Wow!"
+Array: [1 2 3]
+Map: {a:1 b:2}
+```
+
+#### Reference 
+
+    printf format:string a:any...
+
+The `printf` function prints its arguments to the output area according
+to the _format_ string that is the first argument. The _specifiers_
+that start with `%` and are contained in the format string are replaced
+by the remaining arguments in the given order. For example, the
+following code `printf "first: %s, second: %s" "A" "B"` prints `first:
+A, second: B`.
+
+Full list of valid specifiers in Evy:
+
+| Specifier | Description |
+| --------- | ------------- |
+| `%v`      | the argument in a default format  |
+| `%t`      | the word `true` or `false` |
+| `%f`      | decimal point (floating-point) number, e.g. 123.456000 |
+| `%e`      | scientific notation, e.g. -1.234456e+78 |
+| `%s`      | string value |
+| `%q`      | a double-quoted string |
+| `%%`      | a literal percent sign `%`; consumes no value  |
+
+If the arguments for the `%s`, `%q`, `%f`, `%e`, and `%t` specifiers do
+not match the required type, a fatal runtime error will occur.
+
+The _width_ and _precision_ of a floating-point number can be specified
+with the `%f` and `%v` format specifiers.
+
+- Width is the number of characters that will be used to print the
+  number. If the width is not specified, it will be calculated based on
+  the size of the number. It can be useful for padding and aligned
+  output.
+- Precision is the number of decimal places that will be displayed. If
+  the precision is not specified, it will be set to 6 for `%f`. 
+
+Here is a table that shows the different ways to specify the width and
+precision of a floating-point number:
+
+| Verb    | Description |
+| ------- | ------------- |
+| `%f`    | default width, default precision |
+| `%7f`   | width 7, default precision |
+| `%.2f`  | default width, precision 2 |
+| `%7.2f` | width 7, precision 2 |
+| `%7.f`  | width 7, precision 0 |
+
+If the width/precision is preceded by a `-`, the value is padded with
+spaces on the right rather than the left. If it is preceded by a 0, the
+value is padded with leading zeros rather than spaces.
+
+The width, precision and alignment prefix (`-` or `0`) can be used with
+all valid specifiers. For example:
+
+```evy
+printf "right:  |%7.2f|\n" 1
+printf "left:   |%-7.2v|\n" "abcd"
+printf "zeropad:|%07.2f|\n" 1.2345
+```
+Output
+```evy:output
+right:  |   1.00|
+left:   |ab     |
+zeropad:|0001.23|
+```
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -13,6 +13,8 @@ jobs:
       with:
         fetch-depth: 0
     - run: ./bin/make ci
+      env:
+        TERM: vt100
     - run: ./bin/make firebase-deploy
       if: ${{ github.event_name == 'pull_request' }} # only run on PR, deploy prod/master in release job
       env:
--- a/Makefile
+++ b/Makefile
@@ -13,8 +13,8 @@ all: build test lint tiny test-tiny check-coverage sh-lint check-prettier check-
 
 ci: clean check-uptodate all ## Full clean build and up-to-date checks as run on CI
 
-check-uptodate: tidy fmt
-	test -z "$$(git status --porcelain -- go.mod go.sum $(GOFILES))" || { git status; false; }
+check-uptodate: tidy fmt doctest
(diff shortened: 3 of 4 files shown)


docs/builtins.md
```diff
@@ -0,0 +1,718 @@
...
+`print` prints its arguments to the output area, each separated by a
+single space and terminated by a newline character. If no arguments are
+provided it only prints the newline character. 
+
+A tab is printed for the `\t` control character. A newline is printed
+for the `\n` control character. Quotes in string literals are escaped
+with backslash `\`, for example `print "Here's a tab: \t.\nShe said: \"Thank you!\""`.
+
+In a browser environment `print` typically prints to a DOM Element wired
+to the evy runtime. When using evy as Command Line Interface, `print`
```
camh-, line 47: "When using evy as Command Line Interface" doesn't read very well. Perhaps "When running evy from the command line interface"

docs/builtins.md
```diff
@@ -0,0 +1,718 @@
...
+prints to Stdout.
```
camh-, line 48: s/Stdout/standard out/

docs/builtins.md
```diff
@@ -0,0 +1,718 @@
...
+
+    read:string
+
+`read` returns a string that contains the line of input that the user
+entered up until, excluding the newline character. It is a blocking
+functions, which means that it will not return until the user has
+entered a line of input and pressed the Enter key.
+
+In a browser environment `read` typically reads from a DOM Textarea
+Element wired to the evy runtime. When using evy as Command Line
```
camh-, line 81: same here: When running evy from the command line interface...

docs/builtins.md
```diff
@@ -0,0 +1,718 @@
...
+Interface, `read` reads from Stdin.
```
camh-, line 82: s/Stdin/standard in/

docs/builtins.md
```diff
@@ -0,0 +1,718 @@
...
+```
+Bye
+```
+
+#### Reference 
+
+    cls
+
+In a browser environment `cls` typically clears the DOM Element wired to
+the evy runtime which is used for printing. When using evy as Command
```
camh-: again: from the command line interface.

docs/builtins.md
```diff
@@ -0,0 +1,718 @@
...
+Line Interface, `cls` clears the terminal, similar to the Unix `clear`
+or Windows `cls` commands.
+
+### `printf`
+
+`printf` stands for print formatted.
+
+`printf` prints its arguments to the output area according to a _format_
+string. The format string is the first argument, and it
+contains _specifiers_. Specifiers are placeholder that start with a
```
camh-, line 117: I think you could remove "are placeholders that", so it reads: Specifiers start with a ...

Introducing the word "placeholder" when not used elsewhere doesn't really add anything here I think.

docs/builtins.md
```diff
@@ -0,0 +1,718 @@
...
+| Specifier | Description |
+| --------- | ------------- |
+| `%v`      | the argument in its default format  |
+| `%q`      | a double-quoted string |
+| `%%`      | a percent sign `%`  |
+
+#### Example
+
+```evy
+printf "%%v string: %v\n" "abc"
```
camh-, line 133: I think using `%%` in each example is perhaps a little confusing. I get what you are going for with the output, but I wonder if there's a better way to do these examples?

docs/builtins.md
```diff
@@ -0,0 +1,718 @@
...
+%v array: [1 2 3]
+%v map: {a:1 b:2}
+```
+
+#### Reference 
+
+    printf format:string a:any...
+
+`printf` prints its arguments to the output area according to
+the _format_ string that is represented by the first argument.
```
camh-: "that is represented by the first argument is a bit wordy". (I think "represented by" is a longer way of just saying "is". Perhaps

`printf` prints its first argument (the format string) to the output area, replacing
any specifiers in the format string with the following arguments. A _specifier_ starts
with a `%` and specifies how the value of one of the remaining arguments is
formatted. Each specifier (apart from the `%%` specifier, see below) consumes the
next unconsumed argument in order. For example, ...

camh- (commented):
Since you seem to be iterating on this as I review, here are some comments so far,...

docs/builtins.md
```diff
@@ -0,0 +1,718 @@
...
+`print` prints the arguments given to it to the output area. It separates them by
+a single space and outputs a newline character at the end.
+
+#### Example 
+
+```evy
+print "Hello"
+print 2 true "blue"
+print "array:" [1 2 3]
+print "map:" {name: "Scholl" age: 21}
```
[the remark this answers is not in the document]
juliaogris: nice, done, thanks.

docs/builtins.md
juliaogris: thanks, agreed!

docs/builtins.md
juliaogris: Done.

docs/builtins.md
```diff
@@ -0,0 +1,730 @@
...
+| Verb    | Description |
+| ------- | ------------- |
+| %f      | default width, default precision |
+| %7f     | width 7, default precision |
+| %.2f    | default width, precision 2 |
+| %7.2f   | width 7, precision 2 |
+| %7.f    | width 7, precision 0 |
+
+
+The `-` flag tells `printf` to pad the output with spaces on the right,
```
camh-, line 205: I think "flag" is the wrong word to use here. I would perhaps say:

If the width/precision is preceded by a `-`, the value is padded with spaces on the
right rather than the left. If it is preceded by a `0`, the value is padded with leading
zeros rather than spaces.

docs/builtins.md
```diff
@@ -0,0 +1,730 @@
...
+rather than the left. The `0` flag tells `printf` to pad the output
+with leading zeros, rather than spaces.
+
+The width number and the align-left flag  `-` can be used with all valid
```
camh-, line 209: nit: double space after "flag"

docs/builtins.md
juliaogris: done.

docs/builtins.md
juliaogris: done.

docs/builtins.md
juliaogris: done.

docs/builtins.md
juliaogris: ok, agreed, thank you!

docs/builtins.md
juliaogris: ok, agreed.

docs/builtins.md
camh-, line 209: To get rid of "flag" here, I'd say something like:

A specifier width and alignment prefix (`-` or `0`) can be used with all valid specifiers. For example:

And then the examples should have a non %f example I think, since that "for example" seems to be about that.

docs/builtins.md
```diff
@@ -0,0 +1,730 @@
...
+specifiers. For example:
+
+```evy
+printf "right:  |%7.2f|\n" 1
+printf "left:   |%-7.2f|\n" 1
+printf "zeropad:|%07.2f|\n" 1
+```
+Output
+```
+1 2.000000 3
```
camh-, line 219: this line of output seems to be left over from some earlier example

docs/builtins.md
```diff
@@ -0,0 +1,730 @@
...
+
+`typeof` returns the type of the argument as string value.
+
+#### Example
+
+```evy
+a:any
+a = "abcd"
+t := typeof a
+print "typof \"abcd\":" t
```
camh-, line 273: s/typof/typeof/ - and in the output section below too.

docs/builtins.md
```diff
@@ -0,0 +1,730 @@
...
+
+t = typeof {kind:true strong:true}
+print "typeof {a:3 b:4 c:5}:" t
```
camh-, line 276: it looks like `{a:3 b:4 c:5}` is wrong here and should be `{kind:true strong: true}`

docs/builtins.md
```diff
@@ -0,0 +1,730 @@
...
+
+t = typeof [[1 2] [3 4]]
+print "typeof [1 2 true]:" t
```
camh-, line 279: same here - the value in the string is not what `typeof` is being run on

docs/builtins.md
```diff
@@ -0,0 +1,730 @@
...
+exit function is called and print the status code to the output area.
+This makes it easy to identify the source of any errors that occur in
+your program.
+
+## Conversion
+
+### `str2num`
+
+`str2num` converts a string to a number. If the string is not a valid
+number, it returns `0` and sets the global `err` to true.
```
camh-, line 435: Hmmm, this is the first mention of  "global `err`". Perhaps that needs an introduction, or a section that you can link to at least?

Perhaps you meant "sets the global variable `err`" which is probably enough to clear this up.

docs/builtins.md
```diff
@@ -0,0 +1,730 @@
...
+#### Reference 
+
+    str2num:num s:string
+
+TBD
+
+### `str2bool`
+
+`str2bool` converts a string to a boolean. If the string is not a valid
+boolean, it returns `0` and sets the global `err` to true.
```
camh-, line 462: same here: global variable

docs/builtins.md
```diff
@@ -0,0 +1,730 @@
...
+
+TBD
+
+### `sprintf`
+
+`sprintf` stands for print formatted to string. 
+
+`sprintf` returns a string representation of its arguments according to
+a _format_ string. Formatting in `sprintf` and `printf` work the exact
+same way, `printf` for further details.
```
camh-, line 523: see `printf`

docs/builtins.md
juliaogris: ok, pondering.

docs/builtins.md
juliaogris: done

docs/builtins.md
juliaogris: done.

docs/builtins.md
juliaogris: done.

docs/builtins.md
juliaogris: ok, great! thank you. all fixed up (I believe).

docs/builtins.md
juliaogris: done.

docs/builtins.md
juliaogris: flag removed.

docs/builtins.md
juliaogris: oops, sorry, fixed.

docs/builtins.md
juliaogris: pre-declared, global variable ?
fixed with what you suggested.

docs/builtins.md
juliaogris: I think a section on err would be a good idea actually.
Maybe with program control?

docs/builtins.md
juliaogris: yes!

docs/builtins.md
```diff
@@ -0,0 +1,806 @@
...
+printf "There will be unicorns eating lollipops: %v.\n\n" unicorns
+
+printf "They said: %q\n" "wow"
+printf "Array: %v\n" [1 2 3]
+printf "Map: %v\n" {a:1 b:2}
+```
+
+Output
+```
+%v string: abc
```
camh-, line 149: This output needs to be updated for the examples above.

docs/builtins.md
```diff
@@ -0,0 +1,806 @@
...
+has {a:1} "X": false
+```
+
+#### Reference 
+
+    has:bool map:{} key:string
+
+The `has` function takes two arguments: a map and a key. The function
+returns true if the map has the key, and false if the map does not have
+the key. The map can have any value type, such as `{}num` or `{}
```
camh-, line 336: nit: I would not split the type across lines. it should still fit and be shorter than the previous line if joined.

docs/builtins.md
```diff
@@ -0,0 +1,806 @@
...
+exit function is called and print the status code to the output area.
+This makes it easy to identify the source of any errors that occur in
+your program.
+
+## Conversion
+
+### `str2num`
+
+`str2num` converts a string to a number. If the string is not a valid
+number, it returns `0` and sets the global variable `err` to `true`.
```
camh-, line 437: In the other places, you've use "global `err` variable". Should probably do the same here.

docs/builtins.md
camh-: Perhaps. There isn't much to actually say - "There is a pre-declared global variable `err` that gets set by some functions". Perhaps it is just enough to refer to the "global `err` variable" as you have now.

One thing that is missing is what happens to the variable when there is no error in the functions that set it. Is it the programmers responsibility to reset the var to `false`?

juliaogris: well i guess there is also `errmsg`, e.g.:
```
n := str2num "asdf"
print n err errmsg
```
outputs:
```
0 true str2num: cannot parse asdf
```
so maybe it deserves a separate section?

docs/builtins.md
juliaogris: oops, sorry! done.

docs/builtins.md
juliaogris: done.

docs/builtins.md
juliaogris: fixed, thanks.

docs/builtins.md
```diff
@@ -0,0 +1,225 @@
...
+
+#### Reference 
+
+    print a:any...
+
+The `print` function prints its arguments to the output area, each
+separated by a single space and terminated by a newline character. If
+no arguments are provided it only prints the newline character. 
+
+A tab is printed for the `\t` control character. A newline is printed
```
camh-, line 42: technically the tab is the "control character". `\t` is an escape sequence or backslash sequence or something like that

docs/builtins.md
```diff
@@ -0,0 +1,225 @@
...
+for the `\n` control character. Quotes in string literals are escaped
+with backslash `\`, for example `print "Here's a tab: \t.\nShe said: \"Thank you!\""`.
+
+In a browser environment `print` typically prints to a DOM Element wired
```
camh-, line 46: "typically prints to a DOM Element wired to the evy runtime". Why is this "typically"? Could you just say "outputs to the output area" or something? I know the code could be changed to do other things, but this is what it does right now.

docs/builtins.md
```diff
@@ -0,0 +1,225 @@
...
+#### Reference 
+
+    read:string
+
+The `read` function returns a string that contains the line of input
+that the user entered up until, excluding the newline character. It is
+a blocking functions, which means that it will not return until the
+user has entered a line of input and pressed the Enter key.
+
+In a browser environment `read` typically reads from a DOM Textarea
```
camh-, line 80: same comment about "typically..."

docs/builtins.md
```diff
@@ -0,0 +1,225 @@
...
+by the remaining arguments in the given order. For example, the
+following code `printf "first: %s, second: %s" "A" "B"` prints `first:
+A, second: B`.
+
+Full list of valid specifiers in Evy:
+
+| Specifier | Description |
+| --------- | ------------- |
+| %v        | the argument in a default format  |
+| %%        | a literal percent sign `%`; consumes no value  |
```
camh-, line 176: I would put this one first or last as it is different to the others.

camh- (approved):
🫛 Few minor comments, but all good

docs/builtins.md
juliaogris: ok. I have reworded the whole thing a little, hope that's ok.

docs/builtins.md
juliaogris: ok, done, thank you.

docs/builtins.md
juliaogris: ok, done.

docs/builtins.md
juliaogris: fair, i put it at the end like above. seems also more suited there to me.

juliaogris (commented):
@camh- so i think i have addressed all comments, but could have been a 🐾 and drink to many....
i think the doctest script fails samples in the reference section, not sure, haven't dug...

PR #151, opened 2023-06-26
builtin: Fix rand seeding

juliaogris:
Fix rand seeding for tinygo 0.28.1. Go 1.20 automatically seeds its
random functions but tinygo hasn't caught up, so manually seed the
random function, but don't use the deprecated global random.Seed for
it, use the new rand.Source.


camh- (approved):
:green_salad:

PR #152, opened 2023-06-26
doc: Add a doctest script

camh-:
Add an awk script that formats and runs evy code in markdown files. The 
output of the evy code replaces the contents of the associated output block
in the markdown file. Input can be provided to the evy code with an
associated input block.

This can form the basis of a doc testing framework, but for now it is for
ensuring the docs have properly formatted evy and that the output matches
what the program actually outputs.

Move `firebase/deploy` to `scripts/firebase-deploy` since we now have a
`scripts/` directory.


bin/hermit.hcl
```diff
@@ -1,6 +1,6 @@
 sources = ["https://github.com/cashapp/hermit-packages.git", "https://github.com/foxygoat/hermit-packages.git"]
 env = {
   GOBIN: "${HERMIT_ENV}/out/bin",
-  PATH: "${GOBIN}:${PATH}",
+  PATH: "${GOBIN}:${HERMIT_ENV}/scripts:${PATH}",
```
juliaogris, line 4: you were asking somewhere for feedback on this. i think it is good.
i run for example `servedir` regularly - as quite often i don't want to rebuild all deps (to lazy to wait hasn't stung me much yet i think). 
so i think it's fine.

Makefile
```diff
@@ -124,7 +124,7 @@ firebase-public: frontend
 .PHONY: firebase-deploy firebase-deploy-prod firebase-emulate firebase-public
 
 # --- scripts ------------------------------------------------------------------
-SCRIPTS = firebase/deploy .github/scripts/app_token
+SCRIPTS = scripts/firebase-deploy .github/scripts/app_token
```
juliaogris, line 127: oh, nice. turns out scripts has already been used for this purpose. perfect.
I think it's fine to keep the .github/scripts separate - wdyt?

scripts/doctest.awk
```diff
@@ -0,0 +1,174 @@
...
+#
+# ```evy:input
+# Fox
+# ```
+#
+# ```evy:output
+# Hello Fox
+# ```
+#
+# This script requires gawk. It can be run with "-i inplace" in which
```
juliaogris, line 30: is there something that can be done about the gawk dependency? it is the only global dependency that this repo has apart from git i think.

juliaogris (approved):
ok, so I read through that awk script again -  really neat, reads well to someone who doesn't really know awk. 🥪 

rest looks great.

I will add a make target run only with `make ci` that checks for the git status of docs/builtin.md after it's run. or is there a better/ easy way of testing that nothing's changed? maybe separate flag to script?

scripts/doctest.awk
camh-: I use a gawk feature - coprocessors - using the `|&` pipe form that allows me to send input to a program and read it back. I can probably do it with temporary files instead so it should work with the standard Mac awk (although the man page on macos is terrible and underdocumented).

camh-: Oh, I also use gawk for the `-i inplace` CLI syntax to replace the contents of a file rather than writing to stdout. I guess that's not needed too - the when the makefile runs this, it just has to do the tempfile dance.

Makefile
camh-: I would keep `.github/scripts` separate as a place for github actions only scripts.

bin/hermit.hcl
camh-: I dont understand the `servedir` comment. That is go-installed, and wouldn't be in the `scripts/` directory.

juliaogris: It is to say: personally I don't drive everything from the makefile

camh-: as discussed, i've dropped this from the path (dropped the first commit). If it turns out it might be useful based on use/experience we can add it back.

scripts/doctest.awk
camh-: Ok, I've removed the gawk-specific code and removed the comment saying to use `-i inplace`. This should work with regular/posix awk and hopefully works on a mac. You need to run `scripts/doctest.awk docs/builtins.md > new.md` and then `mv new.md docs/builtins.mk` - the makefile will need to do this (perhaps with a tmp name rather that `new.md`)

PR #153, opened 2023-06-27
cli: Add --skip-sleep flag

juliaogris:
Add --skip-sleep flag to the evy cli so that we can run tests and
formatting faster. This flag has been created in the context of the
doctest.awk script.


camh- (approved):
:kiwi_fruit:

PR #155, opened 2023-07-05
housekeeping: Reformat with new prettier, adjust UI funcs

juliaogris:
It seems that prettier 3.0.0 was just released and formats some things
differently by default, so reformat accordingly.

Also, change canvas default background from "transparent" to "white" and
add grid lines at 0 and 100 - as this has been useful in creating docs
and samples.


camh- (approved):
:green_apple:

PR #157, opened 2023-07-09
docs: Add builtin docs for conversion functions and errors

juliaogris:
Add builtin documentation for sections: conversion and errors.
The following list shows the new sections in context.

- ✅ I/O
- ✅ Types
- ✅ Map
- ✅ Program Control
- 👉 Conversion
- 👉 Error
- ⬜️ String
- ⬜️ Random
- ⬜️ Math
- ⬜️ UI
- ⬜️ Event Handlers

Fix a couple of accidentally left of over whitespace in previous PRs.

---

Rendered markdown:
https://github.com/foxygoat/evy/blob/builtin-strings/docs/builtins.md#conversion

Current WIP:
https://github.com/foxygoat/evy/blob/builtin-wip/docs/builtins.md

### DIFF
--- a/docs/builtins.md
+++ b/docs/builtins.md
@@ -438,3 +438,149 @@ number, but it is typically used to indicate whether the program
 terminated successfully or with an error. A status code of 0 means that
 the program terminated successfully, while any other status code is
 considered an error.
+
+## Conversion
+
+### `str2num`
+
+`str2num` converts a string to a number. If the string is not a valid
+number, it returns `0` and sets the global `err` variable to `true`.
+
+#### Example
+
+```evy
+n:num
+n = str2num "1"
+print "n:" n "err:" err
+n = str2num "NOT-A-NUMBER"
+print "n:" n "err:" err
+```
+
+Output
+```evy:output
+n: 1 err: false
+n: 0 err: true
+```
+
+#### Reference
+
+    str2num:num s:string
+
+The `str2num` function converts a string to a number. It takes a single
+argument, which is the string to convert. If the string is a valid
+number, the function returns the number. Otherwise, the function
+returns 0 and sets the global `err` variable to `true`. For more
+information on `err`, see the [Non-fatal Error section](#non-fatal-errors).
+
+### `str2bool`
+
+`str2bool` converts a string to a boolean. If the string is not a valid
+boolean, it returns `false` and sets the global `err` variable to
+`true`.
+
+#### Example
+
+```evy
+b:bool
+b = str2bool "true"
+print "b:" b "err:" err
+b = str2bool "NOT-A-BOOL"
+print "b:" b "err:" err
+```
+
+Output
+```evy:output
+b: true err: false
+b: false err: true
+```
+
+#### Reference
+
+    str2bool:bool s:string
+
+The `str2bool` function converts a string to a bool. It takes a single
+argument, which is the string to convert. The function returns `true`
+if the string is equal to `"true"`, `"True"`, `"TRUE"`,  or `"1"`, and
+`false` if the string is equal to `"false"`, `"False"`, `"FALSE"`, or
+`"0"`. The function returns `false` and sets the global `err` variable
+to `true` if the string is not a valid boolean. For more information
+on `err`, see the [Non-fatal Error section](#non-fatal-errors).
+
+## Error
+
+Evy programs in execution can report two types of errors:
+
+- Fatal errors
+- Non-fatal errors
+
+### Fatal Errors
+
+Fatal errors cause the program to exit immediately with an error
+message. They typically occur when the program encounters a situation
+that it cannot handle, such as trying to access an element of an array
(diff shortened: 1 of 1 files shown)


docs/builtins.md
```diff
@@ -438,3 +438,124 @@ number, but it is typically used to indicate whether the program
...
+argument, which is the string to convert. The function returns `true`
+if the string is equal to `"true"`, `"True"`, `"TRUE"`,  or `"1"`, and
+`false` if the string is equal to `"false"`, `"False"`, `"FALSE"`, or
+`"0"`. The function returns `false` and sets the global `err` variable
+to `true` if the string is not a valid boolean.  For more information
+on errors, see the [error section](#error).
+
+## Error
+
+The `err` variable is used to indicate whether an _non-fatal error_ has
```
camh-, line 510: should you say "global" here (and on the next line for `errmsg`)? You do elsewhere.

docs/builtins.md
```diff
@@ -438,3 +438,124 @@ number, but it is typically used to indicate whether the program
...
+    from running, such as converting a user input string to a number if
+    the string is not a number. Non-fatal errors will set the global
+    `err` variable to `true` and the program will continue executing.
+    If there is no error, `err` is set to `false`. It is therefore the
+    program's responsibility to check `err` after any possible
+    occurrence of an error.
+
+The global `errmsg` variable contains a detailed message about the error
+which is set alongside `err`. `errmsg` is set to the empty string `""`
+if no error has occurred.
```
camh-, line 530: somewhere in this section you should state what happens to these variables when you call another function that could set them. Do they get reset if there is no error, or do they stay set if set previously and the caller should clear them before calling the function?

docs/builtins.md
```diff
@@ -438,3 +438,124 @@ number, but it is typically used to indicate whether the program
...
+The `err` variable is used to indicate whether an _non-fatal error_ has
+occurred. The `errmsg` variable contains a detailed message about the
+error that occurred.
+
+Evy programs in execution can report two types of errors:
+
+- **Fatal errors** cause the program to exit immediately with an error
+    message. These errors can typically be prevented with guarding
+    code, such as checking the array length to avoid a fatal index out
+    of bounds error.
```
camh-, line 519: You could say something about not being able to intercept fatal errors

camh- (approved):
:herb: When I run prettier on this, it removes a few double spaces added in this PR and also indents the fatal/non-fatal list differently. Perhaps run it and keep whichever changes you like?

docs/builtins.md
juliaogris: yes, good catch, thank you.

docs/builtins.md
juliaogris: ok. i tried to say this just above, but i can see this isn't clear enough. I have added a little to this paragraph and started a new paragraph describing the reset situation. please let me know if that's better.

camh-: well, i am a bit stupid it seems. I attached the comment to exactly what i was saying was missing :facepalm: . Sorry, I just didn't see that, but also I was thinking about the `err` variable - does that get reset to false if there is no error?

docs/builtins.md
camh-, line 510: Hmmm, re-reading this, I think I would move this paragraph under the list below. Start this section with "Evy programs in execution can report two types of errors.."

Then continue with how non-fatal errors are reported. It seems this paragraph repeats some of what is said after the list, so it could be merged and have fewer words saying the same thing.

docs/builtins.md
juliaogris: done.

docs/builtins.md
juliaogris: yes, i totally agree - I had already redone this before i read your comment!
if you don't mind, could you look over the whole errors section again please?

juliaogris:
> 🌿 When I run prettier on this, it removes a few double spaces added in this PR and also indents the fatal/non-fatal list differently. Perhaps run it and keep whichever changes you like?

I will add MD prettier check and formatting at the beginning of the next PR (if that's ok).

docs/builtins.md
```diff
@@ -438,3 +438,151 @@ number, but it is typically used to indicate whether the program
...
+that is out of bounds. Fatal errors cannot be intercepted by the
+program, so it is important to take steps to prevent them from
+occurring in the first place. 
+
+One way to do this is to use _guarding code_, which is code that checks
+for potential errors and takes steps to prevent them from occurring.
+For example, guarding code could be used to check the length of an
+array before trying to access an element to avoid an out of bounds
+error. For example, guarding code could be used to check the length of
+an array before trying to access an element. If the access index is out
```
camh-, line 530: The same sentence is repeated here.

camh- (approved):
:herb: LGTM, apart from that one repeated sentence

docs/builtins.md
juliaogris: oh dang. fixed, thank you.

PR #158, opened 2023-07-09
make: Run prettier format checks on all files

juliaogris:
Run prettier format checks on all file formats that prettier can format:
.js, .html, .css, .md, .json, .yaml. Simply update the prettier check
and format commands to run on the repo root rather than the frontend
directory (Makefile changes). Update all relevant files with

 	make prettier

Ensure checks are passing again with `make check-prettier`.

---

This is a mechanical PR using prettier in more places and fixing
formatting accordingly.

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -2,48 +2,48 @@ name: ci/cd
 
 on:
   push:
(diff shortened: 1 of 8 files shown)


camh- (approved):
:green_heart: Very pretty

PR #162, opened 2023-07-13
Add builtin documentation for remaining graphics funcs

juliaogris:
Add builtin documentation for remaining graphic functions (part 2 of 2).
These are `dash`, `linecap`, `text` and `font`.

Change JS `linecap` to default to `"round"` rather than `"butt"`.
Round makes for nicer polylines.

Ensure that the `dash` pattern is reset and restored when drawing grid -
prior to this change the grid would be dashed too.

Here are graphics in the context of builtin docs:

✅ I/O
✅ Types
✅ Map
✅ Program Control
✅ Conversion
✅ Error
✅ String
✅ Random
✅ Math
👉 Graphics
⬜️ Event Handlers

---

Preview: https://github.com/foxygoat/evy/blob/builtin-graphics-part2/docs/builtins.md#dash


docs/builtins.md
```diff
@@ -1742,3 +1742,287 @@ Output
...
+
+#### Reference
+
+    linecap style:string
+
+The `linecap` sets the shape of the ends of lines to the `style` string
+argument. Valid styles are `"round"`, `"butt"` or `"square"`. An
+invalid style takes no effect.
+
+Here is a table of the different linecap shapes and their descriptions:
```
camh-, line 1833: Rather than "Here is..." perhaps just remove the line and the "Shape Description" heading below and just put the table - it is rather self-explanatory as is.

docs/builtins.md
```diff
@@ -1742,3 +1742,287 @@ Output
...
+
+Shape Description
+
+| Shape      | Description                                                                                                           |
```
camh-, line 1837: s/Shape/Style/ - the argument is called `style`

docs/builtins.md
```diff
@@ -1742,3 +1742,287 @@ Output
...
+
+<img width="300" alt="text sample" src="img/text.png">
+
+#### Reference
+
+    text s:string
+
+The `text` function prints the string argument `s` to the canvas at the
+current cursor position. The cursor position is not updated after
+writing text. Only the `fill` and `color` functions update the text
```
camh-, line 1870: Perhaps instead of "update" in this sentence, you say: "Only `fill` and `color` have an effect on the text; `stroke` has no effect."

docs/builtins.md
```diff
@@ -1742,3 +1742,287 @@ Output
...
+color, `stroke` has no effect. For more text styling, such as setting
+_font size_ or _font family_, see [`font`](#font).
+
+### `font`
+
+`font` sets font properties for text. The font properties are
```
camh-, line 1876: sets *the* font properties

docs/builtins.md
```diff
@@ -1742,3 +1742,287 @@ Output
...
+`family`,`size`, `weight`, `style`, `letterspacing`, `baseline`,
+`align`.
```
camh-, line 1878: *and* `align`.

docs/builtins.md
```diff
@@ -1742,3 +1742,287 @@ Output
...
+line 25 56
+move 25 50
+font {align:"center"}
+text "center"
+
+move 55 80
+line 90 80
+move 55 80
+font {baseline:"bottom" align:"left"}
+text "bottom-pq"
```
camh-, line 1948: I can't remember why you added `-pq` to this text and I can't figure it out either.

Ah, I remember - we wanted some descenders to show the position relative to the baseline. Hmmm, still seems a little weird but I'm not coming up with an alternative...

camh- (approved):
:broccoli: with comments

docs/builtins.md
juliaogris: Agreed, done.

docs/builtins.md
juliaogris: Yes! better.

docs/builtins.md
juliaogris: done.

docs/builtins.md
juliaogris: done.

docs/builtins.md
juliaogris: done.

docs/builtins.md
juliaogris: ok, I've just removed it. It is obvious enough.

PR #164, opened 2023-07-14
parser: Remove tinygo v0.27.0 map hack

juliaogris:
Remove tinygo v0.27.0 map hack, working around not being able to use
interfaces as map keys. tinygo v0.28.0 has fixed this bug so remove the
workaround.

### DIFF
--- a/pkg/parser/format.go
+++ b/pkg/parser/format.go
@@ -3,7 +3,6 @@ package parser
 import (
 	"fmt"
 	"io"
-	"reflect"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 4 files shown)


pkg/parser/format.go
```diff
@@ -27,22 +26,22 @@ type formatting struct {
 	//
 	// the current implementation is a hack to work around a tinygo but, details:
```
camh-, line 27: this comment can go

pkg/wasm/imports.go
```diff
@@ -45,7 +45,6 @@ func (rt *jsRuntime) Dash(segments []float64)          { dash(floatsToString(seg
 func (rt *jsRuntime) Linecap(s string)                 { linecap(s) }
 func (rt *jsRuntime) Text(s string)                    { text(s) }
 func (rt *jsRuntime) Font(props map[string]any) {
-	// encode/json not implemented in tinygo 0.27.0, so do it manually
```
camh-, line 48: Perhaps leave a comment as to why this still done manually?

camh- (approved):
:green_apple:

pkg/parser/format.go
juliaogris: done, 🥪

PR #165, opened 2023-07-16
docs: Revise spec (part 1/4)

juliaogris:
Revise first part of spec, with special attention to the grammar.

---

Preview: https://github.com/foxygoat/evy/blob/spec-revision/docs/spec.md


docs/spec.md
```diff
@@ -48,135 +86,158 @@ Here is a WSN defining itself:
...
+whitespace, which is any combination of spaces and tabs. The following
+listing contains the complete syntax grammar for Evy.
 
-    program    = { statement | func | event_handler | NL } .
-    statements = statement { statement } .
+    program    = { statement | func | event_handler | nl } eof .
     statement  = typed_decl_stmt | inferred_decl_stmt |
                  assign_stmt |
                  func_call_stmt |
-                 return_stmt | break_stmt |
```
camh-, line 80: I dont think you can remove these (return/break) from `statement`. The definition of `loop_statement` and `func_statement` which includes the corresponding `break` and `return` statements only allow those statements at the top-level of the block and not nested in an `if` for example inside a function or loop.

e.g. this is invalid according to the grammar:

```
func foo
    for i := range 5
        if i > 3
            break // invalid according to grammar - break not in `block_statements`
       else if i > 4
            return // invalid according to grammar - return not in `block_statements`
        end
    end
end
```

docs/spec.md
```diff
@@ -48,135 +86,158 @@ Here is a WSN defining itself:
...
+    binary_expr   = expr binary_op expr .
+    binary_op     = LOGICAL_OP | COMPARISON_OP | ADD_OP | MUL_OP .
     LOGICAL_OP    = "or" | "and" .
     COMPARISON_OP = "==" | "!=" | "<" | "<=" | ">" | ">=" .
     ADD_OP        = "+" | "-" .
     MUL_OP        = "*" | "/" | "%" .
 
     /* --- Slice and Literals --- */
-    slice       = assignable "[" [expr] ":" [expr] "]" .
+    slice       = <- assignable "["  slice_expr "]" -> .
```
camh-, line 218: nit: double space before `slice_expr`

camh- (approved):
:green_book: LGTM modulo a few comments - the main one being about statements and how I don't think it is possible to separate statements/block_statements/func_statements/loop_statements.

docs/spec.md
juliaogris: good point!

juliaogris: ok, I think I fixed it up. thank you for pointing this out!
I think I will wait for you to look it over one more time. this stuff is tricky.
thank you!

docs/spec.md
juliaogris: done, thank you.

docs/spec.md
```diff
@@ -48,135 +86,154 @@ Here is a WSN defining itself:
...
+The `WS` abbreviation in the grammar comments below refers to horizontal
+whitespace, which is any combination of spaces and tabs. The following
+listing contains the complete syntax grammar for Evy.
+
+    program    = { statement | func | event_handler | nl } eof .
+    statements = { nl } statement { statement | nl } .
     statement  = typed_decl_stmt | inferred_decl_stmt |
                  assign_stmt |
                  func_call_stmt |
-                 return_stmt | break_stmt |
```
camh-, line 80: You need to add this `return_stmt | break_stmt ` part back to `statement` production

camh- (approved):
:green_book: with the `statement` production fixup

docs/spec.md
juliaogris: oh sorry! missed it. Fixed!

camh- (approved):
:green_book: Still LGTM. No more comments.

PR #167, opened 2023-07-17
docs: Fix builtin typo for linecap

juliaogris:
Fix builtin typo for `linecap` to be consistent with other `Reference`
sections. This was due to a premature merge.


camh- (approved):
:green_salad:

PR #169, opened 2023-07-17
docs: Add ToC and horizontal separators to builtins

juliaogris:
Add ToC to builtins.md as an initial overview and end func docs with a
horizontal line for easier readability.

---
Preview: https://github.com/foxygoat/evy/blob/builtin-hsep/docs/builtins.md


camh- (approved):
:kiwi_fruit:

PR #172, opened 2023-07-19
format: Fix misplaced NL before func decl

juliaogris:
Fix misplaced NL before function declaration. This is issue became apparent when

	a := 1
	b := 2
	func fn
	    print "fn" a b
	end

was reformatted to

	a := 1

	b := 2
	func fn
	    print "fn" a b
	end

now fixed.


camh- (approved):
:bug:

PR #173, opened 2023-07-19
parser: Fix bad unreachable code error on comments

juliaogris:
Fix bad unreachable code error on comments by only issuing the error if
the statement is not a comment (or an empty line).


camh- (approved):
:bug:

PR #175, opened 2023-07-23
parser: Fix typing of array concatenation expressions

juliaogris:
Fix typing of array concatenation expressions so that `[]+[true]` now
returns the correct `[]bool` type.

Before:
`print (typeof []+[true])` -> `[]`

Now:
`print (typeof []+[true])` -> `[]bool`

### DIFF
--- a/pkg/evaluator/evaluator.go
+++ b/pkg/evaluator/evaluator.go
@@ -639,6 +639,9 @@ func evalBinaryArrayExpr(op parser.Operator, left, right *Array) (Value, error) 
 		return nil, fmt.Errorf("%w (array): %v", ErrOperation, op.String())
 	}
 	result := left.Copy()
+	if result.T == parser.GENERIC_ARRAY {
+		result.T = right.T
+	}
 	rightElemnts := *right.Copy().Elements
 	*result.Elements = append(*result.Elements, rightElemnts...)
 	return result, nil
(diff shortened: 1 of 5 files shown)


pkg/evaluator/evaluator.go
```diff
@@ -639,6 +639,9 @@ func evalBinaryArrayExpr(op parser.Operator, left, right *Array) (Value, error)
 		return nil, fmt.Errorf("%w (array): %v", ErrOperation, op.String())
 	}
 	result := left.Copy()
+	if result.T == parser.GENERIC_ARRAY {
```
camh-, line 642: Instead of doing this here for this one specific case, I wonder if it might be better to do this in `evalBinaryExpr` and set the type of the `Value` to that of the `*parser.BinaryExpression`, since you set that correctly in the parser. That way you don't need special knowledge of arrays in the evaluator and constrain that logic just to the parser.

camh- (approved):
:kiwi_fruit: with a thought

pkg/evaluator/evaluator.go
juliaogris: yes, conceptually I agree, this would be nicer and less land-mine-y.
however, iiuc, that'd  mean that I have extend the `Value` interface by a `SetType(t *parser.Type)` method I think which might not be worth it (especially as code size is a concern) - wdyt?

PR #176, opened 2023-07-24
parser: Fix function calls with indexed arguments

juliaogris:
Fix function call parsing for arguments with index - maps, type assertions,
grouped expressions and arrays. We're here at a finicky bit of evy. Prior
to this PR we'd get a compile error for:

	print {} []

error message:

	line 1: unexpected whitespace before "["

Remove debug print in tests from previous PR.


camh- (approved):
:bug: LGTM

PR #177, opened 2023-07-29
frontend: Select sharing URL on copy button click

juliaogris:
Select sharing URL on copy button click in frontend, so that it will
also be put into the "Primary Selection" on Linux which people commonly
use to copy text.


juliaogris:
Well, sadly this did not render the desired result on Firefox Linux. It seems that JS select doesn't put things in the "Primary Selection" on Firefox, it needs to be done manually.
Oh well.

PR #179, opened 2023-07-29
eval: Wrap []any and {}any elements

juliaogris:
Wrap `[]any` and `{}any` elements in an `*Any` value if they aren't already
such a value. Previously the following code caused a panic:

	arr := [1 true]
	arr[0] = "abc"
	print arr

### DIFF
--- a/pkg/evaluator/evaluator.go
+++ b/pkg/evaluator/evaluator.go
@@ -262,6 +262,13 @@ func (e *Evaluator) evalArrayLiteral(arr *parser.ArrayLiteral) (Value, error) {
 	if err != nil {
 		return nil, err
 	}
+	if arr.T.Sub.Name == parser.ANY {
+		for i, e := range elements {
+			if _, ok := e.(*Any); !ok {
+				elements[i] = &Any{Val: e}
+			}
+		}
+	}
 	return &Array{Elements: &elements, T: arr.T}, nil
 }
 
@@ -271,6 +278,11 @@ func (e *Evaluator) evalMapLiteral(m *parser.MapLiteral) (Value, error) {
 		val, err := e.Eval(node)
 		if err != nil {
 			return nil, err
+		}
+		if m.T.Sub.Name == parser.ANY {
+			if _, ok := val.(*Any); !ok {
+				val = &Any{Val: val}
+			}
 		}
 		pairs[key] = copyOrRef(val)
 	}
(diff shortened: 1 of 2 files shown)


camh-:
I think this wrapping should also be done with function calls:
```
func f a:any
  print (typeof a)
end
f 1
f [1 2]
```

This prints
```
num
[]num
```

but it really should print `any` in both cases I think.

camh-:
Yeah, a bug too:
```
func f a:any
  print a.(num)
end
f 1
```
```
line 2 column 10: panic: error converting any to type: not an any
```

juliaogris:
> I think this wrapping should also be done with function calls:
> 
> ```
> func f a:any
>   print (typeof a)
> end
> f 1
> f [1 2]
> ```
> 
> This prints
> 
> ```
> num
> []num
> ```
> 
> but it really should print `any` in both cases I think.

I think for this belongs with #179 I think.

camh-:
> I think for this belongs with #179 I think.

I think you meant #174, but I still think it belongs here as it needs to be done by the evaluator not parser. The evaluator will make concrete values from numeric literals but when passing to a function, the concrete value needs to be wrapped in an `evaluator.Any`. For the case of `foo 1` where `foo` takes an any, this is not what #174 is about - that is about composite literals, but this is not a composite.

juliaogris:
> > I think for this belongs with #179 I think.
> 
> I think you meant #174, but I still think it belongs here as it needs to be done by the evaluator not parser. The evaluator will make concrete values from numeric literals but when passing to a function, the concrete value needs to be wrapped in an `evaluator.Any`. For the case of `foo 1` where `foo` takes an any, this is not what #174 is about - that is about composite literals, but this is not a composite.

Yes pr 174 updated thanks.

To me THIS pr is about composite literals only and all of it should be fixed in the evaluator, but I know we disagree on this. Anyway I'll close these Prs and start a new one containing all of it.

camh-:
> To me THIS pr is about composite literals only

Fair enough. Can easily be done in a different PR

camh-:
> I think this is correct as is!
> 
> The point of `typeof` is to give us the underlying type, right?

Yes, that was dumb of me to use typeof to demonstrate the problem. However, the problem does still exist. The `Value` that is passed to a function that takes an `any` arg is not wrapped in an `Any`, but is passed as the concrete type. So if the func tries to re-assign the arg to a different concrete value, evy seems to crash:

```
$ evy run
func f a:any
    a = 2
    print a
end

f 1
f "hello"
2
panic: internal error: String.Set called with with non-String Value
[stack dump elided]
```

juliaogris:
#206 is covering this now.

PR #181, opened 2023-08-04
evaluator: Rename runtime error to panic

juliaogris:
Rename runtime error to panic so that index out of bounds and similar
runtime errors appear like a panic. Then all unrecoverable runtime
errors are panics or internal errors (which should never occur). While
at it, fix the error message for missing closing quote of string
literals and improve on the error section in builtins.md. 
Add a section on using the `panic` builtin.

This is precursory work to finish off the last part of the spec revision,
including errors.

### DIFF
--- a/docs/builtins.md
+++ b/docs/builtins.md
@@ -24,11 +24,11 @@ see [syntax by example](syntax_by_example.md).
 3. [**Map**](#map)  
    [has](#has), [del](#del)
 4. [**Program control**](#program-control)  
-   [sleep](#sleep), [exit](#exit)
+   [sleep](#sleep), [exit](#exit), [panic](#panic)
 5. [**Conversion**](#conversion)  
    [str2num](#str2num), [str2bool](#str2bool)
-6. [**Error**](#error)  
-   [Fatal Errors](#fatal-errors), [Non-fatal Errors](#non-fatal-errors)
+6. [**Errors**](#errors)  
+   [Panic](#panic-1), [Recoverable Errors](#recoverable-errors)
 7. [**String**](#string)  
    [sprint](#sprint), [sprintf](#sprintf), [join](#join), [split](#split), [upper](#upper), [lower](#lower), [index](#index), [startswith](#startswith), [endswith](#endswith), [trim](#trim), [replace](#replace)
 8. [**Random**](#random)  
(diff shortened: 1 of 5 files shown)


camh- (commented):
Another incomplete review, but I have to go now.

docs/builtins.md
```diff
@@ -492,16 +492,35 @@ Output
...
+title := "A looooooooooooooooooooooooooooooooooooooooooong title" // user input
+
+if (len title) > 50
+    panic "title too long"
+end
+```
+
+Output
+
+```evy:output
```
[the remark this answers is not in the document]
juliaogris: Thanks! #182 can now be merged. sorry about this!

docs/builtins.md
```diff
@@ -654,6 +706,12 @@ err: true
 errmsg: str2num: cannot parse "NOT A NUM"
 ```
 
+If you want your own code or function to cause a recoverable error,
+follow the convention of setting the `err` variable to `true` and the
+`errmsg` variable to a message describing the error in the error case.
+In the non-error case, make sure to set the `err` variable to `false`
+and the `errmsg` variable to the empty string.
```
camh-, line 713: The earlier paragraph on recoverable errors just says `err` is set to `false` - I think it should be the same here - just set `err` to `false` and not worry about `errmsg`. `errmsg` is only relevant when `err` is `true`.

docs/builtins.md
```diff
@@ -606,22 +652,28 @@ print arr[i] // out of bounds
...
 
 ```
-line 3: index out of bounds: 5
+line 3: panic: index out of bounds: 5
 ```
 
-### Non-fatal Errors
+If you want your own code to panic, you can use the built-in `panic`
+function. The panic exits your program immediately. It is typically
+used when something unexpected happens, such as invalid user input or a
```
camh-, line 664: I would have thought that invalid user input is something you would typically not panic on. At least I don't think we should be encouraging people to panic on bad input.

camh- (approved):
:green_book: Just a couple of little comments, but otherwise LGTM

docs/builtins.md
juliaogris: Done.

docs/builtins.md
juliaogris: fair enough... I've just removed the `such as...` part
I've removed the user input comment in the sample above too. I can't really think of a good example at the moment.

camh- (approved):
:green_apple:

camh- (approved):
:green_apple: Very good. I approve muchly. Have an apple

PR #182, opened 2023-08-06
docs: Fix and improve doctest error handling

camh-:
Fix the error handling of `doctest.awk` when running evy to output the 
error captured instead of the previous output captured. While fixing this,
capture stderr err from the command executed as that is often where error
output goes. And then close `/dev/stderr` for good measure.

Add an `expect_err` flag to go after the `evy` language marker used by 
doctest to denote that the evy program is expected to exit with an error.
In these cases, we do not report the error and instead capture the error
output to put in the `evy:output` block.

Update `docs/builtins.md` to add this `expect_err` flag to the `exit` 
builtin which exits with an error. This cleans up the output of
`make doctest` so it no longer reports an expected error.

### DIFF
--- a/docs/builtins.md
+++ b/docs/builtins.md
@@ -474,7 +474,7 @@ fraction of a second, e.g. `sleep 0.1`.
 
 #### Example
 
-```evy
+```evy expect_err
 input := "not a number"
 n := str2num input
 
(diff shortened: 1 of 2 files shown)


juliaogris (approved):
This is per usual far outside the comfort zone, but looks good to me.
I appreciate the added functionality for the docs 🥪 !

PR #183, opened 2023-08-08
docs: Use POSIX syntax for deleting array in awk

camh-:
Use the POSIX awk syntax for deleting an array in doctest.awk instead of
the gawk syntax. Whoops.


juliaogris (approved):
lgtm 💚

PR #184, opened 2023-08-08
docs: Update spec with func and var naming

juliaogris:
Update spec with valid variable and function naming. Add short
description on anonymous parameters `_`.


camh- (approved):
:green_heart:

PR #185, opened 2023-08-08
docs: Revise last section of spec contents (4/4)

juliaogris:
Revise spec contents on operators, whitespace and functions. The context of
these changes is as follows:

✅ Intro
✅ Evy syntax specification
✅ Comments
✅ Types
✅ Variables and Declarations
✅ Zero Values
✅ Assignments
✅ Copy and reference
✅ Scope
✅ Strings
✅ Arrays
✅ Maps
✅ Index and Slice
✅ Operators and Expressions
✅ Precedence
✅ Whitespace
✅ Functions
✅ Variadic functions
👉 Break and Return
👉 Typeof
👉 Type Assertion
👉 Run-time Panics and Recoverable Errors
👉 Execution Model and Event Handler
👉 Runtimes

---

Review link: https://github.com/foxygoat/evy/blob/spec-revision-4/docs/spec.md#break-and-return

After this PR I want to create one more cleanup PR replacing italics with
bold for technical terms. This leaves italics for emphasising ordinary words
and seems to better highlight the technical terms.

I also want to consistently title-case or not title case titles.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -43,9 +43,10 @@ To get an intuitive understanding of Evy, you can either look at its
     [Bare returns](#bare-returns), [Function Names](#function-names), [Anonymous Parameters](#anonymous-parameters), [Variadic functions](#variadic-functions)
 20. [**Break and Return**](#break-and-return)
 21. [**Typeof**](#typeof)
-22. [**Type assertion**](#type-assertion)
-23. [**Event Handler**](#event-handler)
-24. [**Run-time Panics and Recoverable Errors**](#run-time-panics-and-recoverable-errors)
+22. [**Type Assertion**](#type-assertion)
+23. [**Run-time Panics and Recoverable Errors**](#run-time-panics-and-recoverable-errors)
+24. [**Execution Model and Event Handlers**](#execution-model-and-event-handlers)
+25. [**Runtimes**](#runtimes)
 
 <!-- genend:toc -->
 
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -1135,85 +1136,252 @@ in Evy. The call arguments must be listed individually.
...
 
-`typeof` returns the concrete type of a value held by a variable as a
-string. It returns a string the same as the type in an evy program,
-e.g. `num`, `bool`, `string`, `[]num`, `{}[]any`, etc. For an empty
-composite literal , `typeof` returns `[]` or `{}` as it can be matched
-to any subtype. e.g. `[]` can be passed to a function that takes an
-argument of `[]num`, or `[]string`, etc.
+The `typeof` function returns the concrete type of a value held by a
+variable as a string. It returns a string that is the same as the type
+in an Evy program, such as `"num"`, `"bool"`, `"string"`, `"[]num"`, `"
```
camh-, line 1209: putting a line break within the backticks causes a space to go into the text (`" {}[]any"`). I would move the line break before the opening backtick if you can - not sure if prettier is going to just move that again though.

docs/spec.md
```diff
@@ -1135,85 +1136,253 @@ in Evy. The call arguments must be listed individually.
...
+
+typeof x: any
+typeof str: string
+```
 
 Only values of type `any` can be type asserted. That means an array of
-type any, `[]any`, _cannot_ be type assert to be an array of type `num`
-or other concrete type:
+type any, `[]any`, _cannot_ be type asserted to be an array of type
+`num` or any other concrete type. However, the elements of an array of
```
camh-, line 1308: `num` or `[]num` here? earlier in the sentence you say "an array of type any, `[]any`", then you say "an array of type `num`".

camh- (approved):
:broccoli: with some small comments

docs/spec.md
```diff
@@ -1135,85 +1136,253 @@ in Evy. The call arguments must be listed individually.
...
+no break 0
+no break 1
+break 2
+x 1 y 2
+
+```
+
+As you can see, the `break` statement causes the loop to exit when the
+value of `y` is equal to 2. The next statement after the loop is then
+executed.
```
camh-, line 1184: Since you used nested loops in the example, perhaps it deserves a mention in the text?

docs/spec.md
juliaogris: sorry, fixed.

docs/spec.md
juliaogris: yes, changed to `[]num`.

docs/spec.md
juliaogris: yeah, good point. I've mentioned it in the text now.

PR #189, opened 2023-08-20
assert: Rework assert utility for tinygo 0.28.1

juliaogris:
Rework testing utility package `assert` to make use of the tinygo's new
reflection. This means we can get rid of of most code in this
packages. As expected, the size of frontend/evy.wasm stays the same.

While at it, add a couple of godoc comments to `assert.NoError` and
`assert.Equal`.


camh- (approved):
:broccoli:

PR #190, opened 2023-08-23
godoc: Improve on existing godocs

juliaogris:
Improve godocs for evy command in main.go.
Auto-generate evy usage message and follow guidelines on:

https://go.dev/doc/comment

Docs can be viewed by pulling this branch and running:

    go run golang.org/x/pkgsite/cmd/pkgsite@latest -open .

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,10 @@ evy-fmt: ## Format evy sample code
 check-evy-fmt:
 	go run . fmt --check $(EVY_FILES)
 
-doc: doctest toc
+.PHONY: check-evy-fmt evy-fmt lint
+
+# --- Docs ---------------------------------------------------------------------
+doc: doctest godoc toc usage
 
 DOCTEST_CMD = ./scripts/doctest.awk $(md) > $(O)/out.md && mv $(O)/out.md $(md)
 DOCTESTS = docs/builtins.md docs/spec.md
@@ -101,7 +104,12 @@ USAGEFILES = docs/usage.md
 usage: install
 	$(foreach md,$(USAGEFILES),$(USAGE_CMD)$(nl))
 
-.PHONY: check-evy-fmt doc doctest evy-fmt lint toc
+GODOC_CMD = ./scripts/gengodoc.awk $(filename) > $(O)/out.go && mv $(O)/out.go $(filename)
+GODOCFILES = main.go
+godoc: install
+	$(foreach filename,$(GODOCFILES),$(GODOC_CMD)$(nl))
+
+.PHONY: doc doctest godoc toc usage
 
 # --- frontend -----------------------------------------------------------------
 NODELIB = .hermit/node/lib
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -101,7 +104,12 @@ USAGEFILES = docs/usage.md
 usage: install
 	$(foreach md,$(USAGEFILES),$(USAGE_CMD)$(nl))
 
-.PHONY: check-evy-fmt doc doctest evy-fmt lint toc
+GODOC_CMD = ./scripts/gengodoc.awk  -v cmd="evy --help" $(md) > $(O)/out.md && mv $(O)/out.md $(md)
```
camh-, line 107: nit: double space before `-v`

Makefile
```diff
@@ -101,7 +104,12 @@ USAGEFILES = docs/usage.md
...
+GODOCFILES = main.go
+godoc: install
+	$(foreach md,$(GODOCFILES),$(GODOC_CMD)$(nl))
```
camh-, line 110: This foreach doesn't really make sense as if you add more files to `GODOCFILES` it will still run `evy --help` as that is the command in `GODOC_CMD`. I'd suggest simplifying this, getting rid of the two vars and just doing:

```
godoc: install
        ./scripts/gengodoc.awk -v cmd="evy --help" main.go > $(O)/out.go && mv $(O)/out.go > main.go
```

(note: i just typed this in the browser and didn't check/test for typos)

I also note that GODOC_CMD calls the tmp file `.md` even though it is go. doesn't matter much though.


but but but, just added a comment below about extracting the command from the usage message so you don't need `-v cmd=...` here, which means you could keep this loop. I think that may be better.

scripts/gengodoc.awk
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env -S awk -f
+$0 ~ "//\tUsage:" {
+        in_usage = 1
+}
+$0 ~ "^(package |// [[])" && in_usage {
+        system(cmd "| sed -e '/./s|^|//\t|' -e 's|^$|//|'")
```
camh-, line 6: I've changed my mind about this - passing `cmd` via an arg. Instead this could be:

    system(cmd " --help | sed ...")

and put `cmd = $3` after `in_usage = 1` in the action above.

camh- (commented):
docs are good. just re-thinking gengodoc...

Makefile
```diff
@@ -101,7 +104,12 @@ USAGEFILES = docs/usage.md
 usage: install
 	$(foreach md,$(USAGEFILES),$(USAGE_CMD)$(nl))
 
-.PHONY: check-evy-fmt doc doctest evy-fmt lint toc
+GODOC_CMD = ./scripts/gengodoc.awk -v cmd="evy --help" $(md) > $(O)/out.go && mv $(O)/out.go $(md)
```
camh-, line 107: the `-v cmd=...` is now superfluous. Also the `$(md)` in this originally stood for markdown. Perhaps use `$(filename)` or something? You would also need to update the `md` in the `$(foreach ...)` line below.

juliaogris: Fixed, thank you!

camh- (approved):
:leafy_green:

PR #192, opened 2023-08-29
all: Upgrade tooling and dependencies

juliaogris:
Do the usual dance of

    hermit upgrade
    go get -u all
    go mod tidy

and then a manual update to go.mod file with most recent go version.
This time there were no new linter errors!


go.mod
```diff
@@ -1,5 +1,5 @@
 module foxygo.at/evy
 
-go 1.20
+go 1.21
```
camh-, line 3: :pineapple: Did you see the thread on #modules on gopher slack? Apparently the tooling now writes "1.21.0" here which causes parsing problems with < go 1.21.0 (older tool chains do not expect a patch version here and emit a parse error instead of an error about the version being too high).

What I'm saying is I'm not 100% sure this is correct, but I did a little bit of playing around and it seems to be ok to use this form still. So this comment is really just a heads-up in case you see any weirdness here later.

camh- (approved):
:green_salad:

go.mod
juliaogris: Nope completely missed it! Thanks

PR #195, opened 2023-09-02
evaluator: Revise godocs

juliaogris:
Privatise a whole bunch of evalutor stuff and do the usual refactors
when taking a fresh look at the code from a godocs perspective. Then
revise godocs. This is the last bit of docs revision for now.


camh- (approved):
:broccoli: This is all really really good. The docs read very well and I found no mistakes.

Just the one comment...

PR #197, opened 2023-09-05
all: Fix len func and improve errors

juliaogris:
- Fix built-in len function for strings, especially for emojis
- Improve ErrVarNotSet error if funcs use variables that have not yet
  been declared at call site.

### DIFF
--- a/pkg/evaluator/evaluator.go
+++ b/pkg/evaluator/evaluator.go
@@ -22,12 +22,12 @@ var (
 	ErrSlice         = fmt.Errorf("%w: bad slice", ErrPanic)
 	ErrBadArguments  = fmt.Errorf("%w: bad arguments", ErrPanic)
 	ErrAnyConversion = fmt.Errorf("%w: error converting any to type", ErrPanic)
+	ErrVarNotSet     = fmt.Errorf("%w: variable has not been set yet", ErrPanic)
 
 	ErrInternal         = errors.New("internal error")
 	ErrUnknownNode      = fmt.Errorf("%w: unknown AST node", ErrInternal)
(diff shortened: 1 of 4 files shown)


pkg/evaluator/evaluator.go
```diff
@@ -22,12 +22,12 @@ var (
 	ErrSlice         = fmt.Errorf("%w: bad slice", ErrPanic)
 	ErrBadArguments  = fmt.Errorf("%w: bad arguments", ErrPanic)
 	ErrAnyConversion = fmt.Errorf("%w: error converting any to type", ErrPanic)
+	ErrVarNotSet     = fmt.Errorf("%w: var has not been set yet", ErrPanic)
```
camh-, line 25: s/var/variable/ I think - in the error message that is.

pkg/evaluator/evaluator_test.go
```diff
@@ -133,6 +134,36 @@ print f2
...
+	err := eval.Run(prog)
+	assert.Equal(t, true, errors.Is(err, ErrVarNotSet))
+	evalErr := &Error{}
+	assert.Equal(t, true, errors.As(err, &evalErr))
+	assert.Equal(t, "line 7 column 13", evalErr.token.Location())
+}
+
+func TestLenString(t *testing.T) {
+	prog := `
+print (len "🌟⭐️✨🌙🪐") // 6 runes, 6 graphemes
```
camh-, line 160: This looks like 5 graphemes to me. perhaps remove the star since you do a test with that next?

camh- (approved):
🐸 with a couple of quibbling comments

PR #206, opened 2023-09-05
Composite lit fixes

juliaogris:
Enable late typing of composite literals, such that `[1]` can be `
[]num` or an `[]any` depending on what is required by the assigned
variable, parameter or return type.

This finally allows for an evy implementation of a function like `font`
which accepts and `{}any` to call as: `font {size: 4}`.

In order to get there refactor parser.Type and fix bug with composites
of any such as []any not wrapping their literal elements in an
evaluator.anyVal


juliaogris:
🥪 🍪 
closing in favour of yet another, more deliberate attempt at this.
First commit and all tests have been merged to `main`.

PR #208, opened 2023-10-05
Replace boxygoatbot with evlangbot creds

juliaogris:
Replace `boxygoatbot` with `evlangbot` credentials so that hopefully we can
release again. 🤞

This is trying to address the following failed CI release job:
https://github.com/evylang/evy/actions/runs/6419265641/job/17428830002


camh- (approved):
:frog:

PR #210, opened 2023-10-08
frontend: Turn off auto-capitalise on mobile devices

juliaogris:
Turn off auto-capitalise on mobile devices. The values were incorrectly set to
false rather than `"off"` and `"none"`.


camh- (approved):
:green_salad:

PR #212, opened 2023-10-19
frontend: Add sidemenu and dialogs

juliaogris:
There are oodles of updates here. The frontend needs to be reworked at some
point, but in the meantime, this is a visual update and refresh according to
Jason's feedback:
 
* Add About Evy dialog
* Rework sharing via dialog rather than full-screen modal
* Add sidemenu
* Span top header across full screen


camh- (commented):
I don't have much value to add for JS/HTML/CSS I'm afraid. Seems fine. Had a play with the site and I noticed the discord link did not work for me - it lead to the page https://discord.com/invite/XHHXf6y29n which is an invitation, but I am already on the site so I don't need an invitation.

Also your email address on the commits is wrong - it is `@mail.com` not `@gmail.com`

frontend/img/evy-e.svg
```diff
@@ -0,0 +1,127 @@
...
+        path.highlight.animate,
+        path:hover:not(.animate) {
+          stroke: #ffcf00;
+          stroke-width: 13;
+        }
+      }
+    </style>
+    <script>//<![CDATA[
+      function initEvySVG(e) {
+      	const svg = e.target.ownerDocument;
```
camh-, line 47: nit: This line has a tab before `const` instead of 2 spaces. the rest of the lines are all space-indented.

camh- (approved):
just one little nit

PR #214, opened 2023-10-20
frontend: Fix mobile style and UI buglets

juliaogris:
- Remove autofocus from copy input to avoid virtual keyboard popping up
- Fix size and padding of evy-e in about dialog
- Hide sidebar when showing "About" dialog, to avoid flicker on close
- Fix sticky hover state on run button on mobile


camh- (approved):
LVMLUITM - looks very much like ui to me :)

I'm not sure I can add much more than that :(

PR #217, opened 2023-10-24
parser: Implement Any AST Node

juliaogris:
Implement Any AST Node, wrapping literals, nested literals and non-any
variables if the target assignable requires it.

As a follow-up cleanup step, simplify scope.set: remove target type parameters
as the target target is now fully and correctly encoded by the value in the
scope derived from *parser.Any.

---

Only the last 2 commits are relevant, the others are covered in #215 and #216.


camh- (approved):
:green_apple: All looks great - one little typo, but I like how this has turned out.

PR #218, opened 2023-10-25
parser: Refactor UNTYPED_ARRY and UNTYPED_MAP names

juliaogris:
Rename UNTYPED_ARRY and UNTYPED_MAP to EMPTY_ARRY and EMPTY_MAP if they 
represent an empty array or map and to GENERIC_ARRAY and GENERIC_MAP if
they are used as generic parameters to builtins.

---

This PR is based on PR217. Only the last commit is relevant.


camh- (approved):
:leaves: I do like this better. Generic makes sense, empty makes sense.

PR #219, opened 2023-10-25
frontend: Improve initial load UX on slow networks

juliaogris:
Improve initial loading user experience on slow networks by pre filling the
editor area with the Welcome example, fixing a load race condition and
showing a load dialog if the run button is clicked before the tool chain is
ready.

While at it fix a regression scroll bug for the example modal and add styling
for the runbutton active state to avoid the weird blue color on tap on
mobile.


camh- (approved):
Seems to work fine. Tested with slow mode and no caching as suggested.

PR #223, opened 2023-10-30
parser: Abort parsing for invalid function signatures

juliaogris:
There is a prepare step parsing all function signatures. If an error is
encountered in this stage, abort further parsing. This addresses a bug report
as covered in the added test.

Thank you @alecthomas for the report.


camh- (approved):
:leafy_green:

PR #225, opened 2023-11-12
frontend: periodically backup code to URL hash

pentaphobe:
### Overview

- Uses the `encode()` mechanism currently used for sharing to also update the URL hash (and history) periodically.
- Avoids spamming the history by only creating a single history entry once there is a hash present - this is _technically_ the right thing to do, but also sacrifices a little extra nicety (see [caveats](#caveats))

#### How to see what this does

1. Spin up the front end locally (`make frontend-serve`)
2. Make some edits to code in the browser
3. Hit the browser's refresh button and see that your code is still there (unless you refresh less than a second after your changes)

### Why?

A couple reasons:

1. Low effort + high value way of saving editor state
2. Safeguard against a heisenbug which caused editor state to reset randomly 

#### (1) Low effort + high value

- requires no additional UI or significant code changes but provides a modicum of extra protection for the user
- reuses the same existing mechanism as share URLs
- (disastrous) scenarios helped by state stored in URL
   - accidentally navigating away from page (eg. with Chrome's "helpful" swipe motion) - just navigate back
   - accidentally closing tab - just reopen tab from history

#### (2) Safeguard against heisenbug

There's currently a minor bug in the UI which seems to cause a random refresh of the page (losing all work) - whilst this isn't a fix for the bug, the hope is that it provides a modicum of safety for such occurrences.

> But no, really _why_?

Because I got overly excited by `evy` on Friday, spent a while writing a toy game and then lost my work.  

No big deal for me, but if I were a new/young programmer experiencing the same excited flow state, that loss woulda be devastating

> [!NOTE]
> re. the heisenbug:
> It arose after maybe 45 minutes of editing - no idea how to reproduce.
>
> Did do some scrambling around in Chrome dev tools to see if I could recover my work (or work out what happened) but found no adequate clues.
>
> Opted not to create an issue in the repo as I wasn't sure that was the preferred method of tracking feedback

### Details

- Only triggers on keypress within the editor itself, and will only update history if the code has actually changed.
- The update method itself is throttled, so should only trigger at most every 1 second (configurable with the `codeBackupDelayMillis` constant)  Can be disabled entirely by setting `codeBackupDelayMillis` to `-1`

### DIFF
--- a/frontend/index.js
+++ b/frontend/index.js
@@ -15,6 +15,8 @@ let actions = "fmt,ui,eval"
 let editor
 let errors = false
 let sidemenu
+// how frequently to store code state in URL hash, set to -1 to disable entirely
+const codeBackupDelayMillis = 1000
 
 // --- Initialise ------------------------------------------------------
 
@@ -331,6 +333,10 @@ function clearOutput() {
 
 async function initUI() {
   document.addEventListener("keydown", ctrlEnterListener)
+  if (codeBackupDelayMillis >= 0) {
+    // periodic backup of code text area to the URL hash
+    document.addEventListener("keydown", codeUpdateListener)
+  }
   await fetchSamples()
   window.addEventListener("hashchange", handleHashChange)
   document.querySelector("#modal-close").onclick = hideModal
@@ -361,6 +367,14 @@ function ctrlEnterListener(e) {
     document.querySelector(".editor textarea").blur()
     handleRun()
   }
+}
+
+// on keypress within code editor, update history with current code
+function codeUpdateListener(e) {
+  // skip if not an event in the code editor
+  if (!e.target.matches(".editor, .editor *")) return
+
+  updateHistoryThrottled()
 }
 
 // --- UI: URL-hash change handling ------------------------------------
@@ -983,12 +997,19 @@ async function share() {
     return
   }
   const baseurl = window.location.origin + window.location.pathname
-  const encoded = await encode(editor.value)
   const input = document.querySelector("#dialog-share .copy input")
-  input.value = `${baseurl}#content=${encoded}`
+  const hash = await contentHash()
+  input.value = `${baseurl}${hash}`
   input.setSelectionRange(0, 0)
   input.blur()
   document.querySelector("#dialog-share").showModal()
+}
+
+// get encoded state of editor as a URL hash
+// used by share() and codeUpdateListener()/updateHistory()
+async function contentHash() {
+  const encoded = await encode(editor.value)
+  return `#content=${encoded}`
 }
 
 async function encode(input) {
@@ -1038,6 +1059,28 @@ async function bufferFromStream(stream) {
   return buffer
 }
 
+async function updateHistory() {
+  const hash = await contentHash()
+  const currentHash = window.location.hash
+  if (hash === currentHash) {
+    // no changes detected
+    return
+  }
+
+  // update history without triggering hashchange event
+  // if no hash is present yet, create a new history entry otherwise just
+  // replace the most recent one to avoid cluttering the history
+  if (currentHash === "") {
+    history.pushState({}, "", hash)
+  } else {
+    history.replaceState({}, "", hash)
+  }
+}
+
+// throttled version of the updateHistory function
+// (to avoid calling encode() too frequently)
+let updateHistoryThrottled = throttle(updateHistory, codeBackupDelayMillis)
+
 // --- Utilities -------------------------------------------------------
 
 function getElements(q) {
@@ -1058,4 +1101,16 @@ function showElements(q) {
 
 function hideElements(q) {
   getElements(q).map((el) => el.classList.add("hidden"))
+}
+
+function throttle(funcToThrottle, waitTimeMillis) {
+  let timer = null
+  return async function (...args) {
+    if (timer === null) {
+      timer = setTimeout(() => {
+        funcToThrottle(...args)
+        timer = null
+      }, waitTimeMillis)
+    }
+  }
 }


frontend/index.js
```diff
@@ -331,6 +333,10 @@ function clearOutput() {
 
 async function initUI() {
   document.addEventListener("keydown", ctrlEnterListener)
+  if (codeBackupDelayMillis >= 0) {
```
pentaphobe, line 336: Bit of a mysterious way to do feature toggling, I know - but was trying to avoid adding too many top level variables!

frontend/index.js
```diff
@@ -1038,6 +1059,28 @@ async function bufferFromStream(stream) {
   return buffer
 }
 
+async function updateHistory() {
+  const hash = await contentHash()
+  const currentHash = window.location.hash
+  if (hash === currentHash) {
+    // no changes detected
+    return
+  }
```
pentaphobe, line 1068: As mentioned in PR notes, this isn't the most efficient way to check for code changes - but seemed like a good enough first pass since alternatives would have involved adding additional state variables at the top level (something I wanted to avoid)

camh-:
Hey Keilin. Welcome and muchas gracias for the PR!

A couple of things:
1. Sorry about the extra failures on CI. It looks like there is now a newer version of `prettier` and it wants to reformat the SVGs. We'll fix that up and you can rebase on top of that. There are a couple of quoting issues in your PR that prettier wants to change - this is most easily done locally with `make prettier`.
2. Our style for commits, which we haven't written down or told anyone, is to leave the emojis off commits in a PR and put them on the merge commit. We have a [merge script](https://github.com/foxygoat/git-scripts/blob/master/git-pr-merge) that formats the merge how we like it and that's when we add the emoji. You can run `git log --oneline` to see how this looks with single-commit PRs squash merged, and multi-commit PRs with the emoji followed by the commits in the PR (`--decorate` shows this more clearly and `--first-parent` removes the noise of individual commits showing just the PRs with emojis).
3. Regarding the content of the PR, I think I'd like to have a better idea of what is happening with the reloading. I don't think we've seen any unexpected reloads so I'm curious what's going on there and not sure I want to patch around it without understanding it.
4. That said, Julia and I did talk about maintaining the playground state across reloads. Only high level stuff so far but it was a high priority. The plan was to sync the code to local storage and pull it out when the page loads without a location hash. @juliaogris may have more to add here, but I feel like we should proceed with that rather than the reload protection in this PR as it should address the same issue and provide additional benefit. Feel like seeing what might be needed for that?

Thanks again for your PR. Hope we don't put you off with our PR ways, which I know can be a bit detailed and picky. Happy to hear feedback on that as we go.

pentaphobe:
> Hey Keilin. Welcome and muchas gracias for the PR!

¡De nada!

>     1. Sorry about the extra failures on CI. `<..snip..>` a couple of quoting issues in your PR that prettier wants to change `<..snip..>`

All good - just did exactly that, though I'll fix up my commit so that it doesn't include the SVGs  👍 

>     2. Our style for commits, which we haven't written down or told anyone, is to leave the emojis off commits in a PR and put them on the merge commit. We have a [merge script](https://github.com/foxygoat/git-scripts/blob/master/git-pr-merge) that formats the merge how we like it and that's when we add the emoji. You can run `git log --oneline` to see how this looks with single-commit PRs squash merged, and multi-commit PRs with the emoji followed by the commits in the PR (`--decorate` shows this more clearly and `--first-parent` removes the noise of individual commits showing just the PRs with emojis).

Ah, got it!  I perused the log before committing so that I could venture to match style, but didn't pick up on that aspect.

Will rebase the commit to remove the clutter - is there something additional I need to do in order to ensure the merge/squash gets the right emoji?

>     3. Regarding the content of the PR, I think I'd like to have a better idea of what is happening with the reloading. I don't think we've seen any unexpected reloads so I'm curious what's going on there and not sure I want to patch around it without understanding it.

Yeah I should probably have explained myself better there, will add more detail to the "Why?" section of the description rather than cluttering up this response.

**TL;DR:** don't know root cause of the refresh issue, but seemed like automatic state retention was a useful enough feature in itself as to not be _purely_ a workaround

>     4. That said, Julia and I did talk about maintaining the playground state across reloads. Only high level stuff so far but it was a high priority. The plan was to sync the code to local storage and pull it out when the page loads without a location hash. @juliaogris may have more to add here, but I feel like we should proceed with that rather than the reload protection in this PR as it should address the same issue and provide additional benefit. Feel like seeing what might be needed for that?

Do you think they (local storage & history use) are mutually exclusive?

Seems like they could play quite well together:

```mermaid
stateDiagram-v2

visit: visit evy.dev/play
hasHash: URL has hash?
hasLocal: has local storage?
loadHash: load hash
loadLocal: load local storage
loadDefault: use default starter code

visit --> hasHash
hasHash --> loadHash: Yes
hasHash --> hasLocal: No
hasLocal --> loadLocal: Yes
hasLocal --> loadDefault: No
```

I'd assume if you were to use local storage then you'd either limit them to a single active file, or would need to provide UI for selecting from multiple - so perhaps a bit more of an involved feature?

Either way - won't be offended if this PR is culled for not fitting with other plans 👍 

> Thanks again for your PR. Hope we don't put you off with our PR ways, which I know can be a bit detailed and picky. Happy to hear feedback on that as we go.

Not at all - always enjoy learning the ins and outs of new contribution guidelines / tooling :)

pentaphobe:
- [x] Removed emoji from commit title
- [x] Ran `make prettier` over the JS changes (changes `'` to `"`)
- [x] Rebased off of #226 
- [x] Updated PR description to include some more details
- [ ] Removed emoji from PR title  (I lack the permission to edit PR titles)

juliaogris:
Dang, I really wish I had some capacity to properly dig into this - but that will have to wait until I return from the mountains.

Something about the hashing logic and its code has not been very well thought out (my bad). The idea behind it was: Let's make it possible to directly share provided samples and custom made ones via URL.

Here a questions that I haven't thought through: if you load content via a hash and modify the content - should the hash be cleared? Probably.

Currently creating the hash for a custom example is gzipping and base64 encoding the full Evy code contents (I'm also not quite clear how costly it really is). That's only done cause I wanted to stay backend free for as long as I could, but this will go - the URLs are pretty ugly.

@pentaphobe - IIUC you are suggesting that occasionally the full site reloads itself? That seems seriously broken and should never happen (haven't seen this myself yet). I'd love to get to the bottom of this. I might have to add some debug logging.

> I'd assume if you were to use local storage then you'd either limit them to a single active file, or would need to provide UI for selecting from multiple - so perhaps a bit more of an involved feature?

You mean the case where there might be a couple of browser tabs open? Hmmm, yeah - hadn't thought of that tbh. What happens when you go to the older/stale tab and reload that?
I don't think there should be any such thing as a "file selection" via the UI. So a single-file-local-storage backup with most recent changes overriding previous ones even if you have more than one tab open is likely an improvement to what is there now - ie you lose everything when you reload the page.

I think your state-diagram is sound, however imo it requires clearing the hash in the URL the moment the originally loaded evy content is modified, so that when the user hits reload they don't accidentally lose their local changes. wdyt?

pentaphobe:
> @juliaogris 
> IIUC you are suggesting that occasionally the full site reloads itself? That seems seriously broken and should never happen (haven't seen this myself yet). I'd love to get to the bottom of this.

Correct - it's possible an extension triggered it, though I don't use many and haven't experienced it anywhere else before.

Will definitely see if I can reproduce the problem, but it only happened after I'd used the session for _quite a while_ (by that point I'd reused the same tab to write a half dozen toy experiments, and then spent maybe 30-45 mins working on a whack-a-mole game / example)

If you're users of Github Issues I'm more than happy to create one so that any findings are in the one place

---

> @juliaogris:
> What happens when you go to the older/stale tab and reload that?

I'm not sure if you're asking about a hypothetical local storage solution or this PR, but I'll assume the latter:  the old/stale tab JustWorks™

With the changes in this PR you can actually dive back into any previously tabs in your browser history and it'll work.

---

> if you load content via a hash and modify the content - should the hash be cleared? Probably.
>
> `<..snip..>`
>
> I think your state-diagram is sound, however imo it requires clearing the hash in the URL the moment the originally loaded evy content is modified, so that when the user hits reload they don't accidentally lose their local changes. wdyt?

Currently the hash is updated whenever the code editor is changed, so the above scenarios seem covered (if I'm following you..) 

The hash wouldn't be _cleared_ but it would be updated to a new (accurate) value, which seems to address the main concern, ie.


1. **Alice:** Creates new cool evy program, hits "share" and sends the link to **Boo**
2. **Boo:** Opens the shared URL (which has code state in the hash)
3. **Boo:** Starts excitedly making changes
4. **This PR:** Updates the hash within 1000ms of **Boo's** new changes and stores it in the history
5. **Boo:** Gets confused and hits refresh on the browser
6. **This PR:** Reloads **Boo's** changes from the hash/history (**NOT Alice's**) 

Also notable that if **Boo** didn't realise there was a "Share" button and just copied the browser location it would include **Boo's** version as expected, not **Alice's**

juliaogris:
@pentaphobe 

Sorry for the long silence (overseas, sick). I've responded in more detail on Slack but I'll copy over here in case you missed it:

I've finally done my homework and read up on IndexDB, LocalStoreage and SessionStorage - just this one time 😉 , I think @camh- might have it right. It seems session storage is what we want:
you are working on an evy program in a session (potentially several sessions in parallel in multiple browser tabs). On every keystroke save the unencoded editor contents of the browser  sessions. When the tab is closed the session is gone. If anything reloads the tab, the session gets restored, only the undo/redo history is lost.

Are you still interested in updating your PR to reflect this? 

Something like the following 2 functions wired up in the right places should do the trick ( @camh-  hacked something up a little while back, just copy-pastering):

```
function saveEditorSession() {
  sessionStorage.setItem("evy-editor", editor.value)
}

function loadEditorSession() {
  let editorValue = sessionStorage.getItem("evy-editor")
  if (!editorContent) {
    return false
  }
  editor.update({ value: editorValue, errorLines: {} })
  return true
}
```

Totally understood if you've lost interest in this patch by now (sorry for the slow responses), in this case I might try and take over if you don't mind? Entirely up to you.

Thanks again for your input!

juliaogris:
@pentaphobe - I'm closing this now as #230 has been merged and #227 closed.
Thank you for your awesome contribution here and all your lovely Evy samples. They will be added to the frontpage. 🙏

PR #226, opened 2023-11-13
frontend: Re-run prettier over source

camh-:
Re-run prettier with:

    make prettier

to re-pretty all the front-end source code, as prettier 3.1.0 has just
been released and some formatting has changed. We need to keep
up-to-date with this, otherwise people's PRs will fail for reasons
unrelated to their changes.

Link: https://prettier.io/blog/2023/11/13/3.1.0.html


camh-:
@pentaphobe We'll get this merged and you can rebase on top of the new master. Then `make prettier` on your PR should update only the stuff you have changed.

juliaogris (approved):
I guess the change in the newer prettier version is that now formats .svg files by default which it didn't do previously.

LGTM - 🍏 Thanks you!!

camh-:
> I guess the change in the newer prettier version is that now formats .svg files by default which it didn't do previously.

There was nothing in the release notes about it that I could see. 3.0.0 said it now formatted scripts inside svg but that was released in July. I see in this PR there is some reformatting of a script in svg, so I wonder if they previously just forgot to include svgs by default? Doesn't matter though.

PR #230, opened 2023-12-30
frontend: Save editor contents to session storage

### LINKED ISSUE #227: Fix random frontend reloads
Fix random frontend reloads so that code does not get lost after a longer in browser session.

Copying @camh- notes from discord:

>I've been thinking a little about the issue of a page reloading and losing your program. This is a poor user experience whether or not the reload is due to some buggy extension, and accidental button press or even deliberate in order to unstick something that looks stuck. So my desire for understanding why the page reloaded in the first place is not particularly useful or relevant.
>
>I read up a little on session and local storage in the browser (my current source of info is this page: https://www.freecodecamp.org/news/web-storage-localstorage-vs-sessionstorage-in-javascript/) and it seems like session storage is what we want to use. Using the url hash is just a dodgy session storage it seems to me. The purpose of the hash was to share an evy program without us needing to set up a backend for it. It was always a little hacky and we expected i
t to be temporary so I think it best we dont use it for reload protection.
>
>Since session storage survives a page reload and is per-origin per-instance, each concurrent tab will have its own state which is what we want. It is only cleared when the tab/window/browser is closed.
>
>I also think we should do more to use the hash less. I think once a page is loaded from hash content, we should clear the hash and rely on session storage after that. You should only get a hash from the sharing dialog. Once cut and then pasted into a new window, after load the hash is cleared. This may even be necessary, perhaps, maybe, if the hash part of the URL is considered part of the origin. Regardless I think we should not even worry - just clear the hash and save to session storage.

juliaogris:
Save editor contents to session storage so that accidental browser reloads
to not lose the contents. We attempt to reload the session if the URL
fragment is empty and the editor saves its content to session on every
single session.

In preparatory commits fix up comments and remove unused parameters. Also
refactor onHashChange function for clarity and readability and remove URL
fragment as soon as contents get edited.

Issue: https://github.com/evylang/evy/issues/227

---

Fixes #227


frontend/index.js
```diff
@@ -379,11 +379,15 @@ async function handleHashChange() {
...
   const { source, crumbs } = await fetchSourceWithCrumbs(opts)
+
+  editor.onUpdate(null)
   editor.update({ value: source, errorLines: {} })
+
   document.querySelector(".editor-wrap").scrollTo(0, 0)
   crumbs && updateBreadcrumbs(crumbs)
   clearOutput()
   await format()
+  editor.onUpdate(clearHash)
```
camh-, line 393: I think you should also call `clearHash` here if `opt.content` is populated. The content urls are ugly and not really useful to leave in the URL bar once the content is loaded. It's annoying to have to edit it off (because its so long) when you frob with URLs. So I favor clearing it once the content has been loaded from it.

frontend/index.html
```diff
@@ -299,5 +299,12 @@ <h1>About</h1>
         />
       </symbol>
     </svg>
+    <script type="text/javascript">
+      // Avoid flicker of placeholder sample in editor if we have a saved session.
```
camh-, line 303: I still got some bad flicker the first time I loaded the per-PR app deployment. At that point, nothing was cached and the page loaded with the default content, then cleared it for about 1 sec before it came back again. I can reproduce it with the dev tools, disabling the cache and setting the throttling to fast or slow 3G. A scrollbar also appears for a brief period too and then disappears when the editor takes over.

camh- (approved):
:green_heart: looks really good. cleaner than what I came up with. i like it

frontend/index.js
juliaogris: Ok, I could got with that. If and when URLs become shorter snippet hashes that can be re-evaluated. For examples I think the URL should stay: it's meaningful and easily copyable.

frontend/index.html
juliaogris: I've added a last commit that rearranges the editorInitialisation and source code loading to avoid the flicker. Please let me know what you think.

camh-: The flickering of the code has gone, but there is still a scrollbar that appears and disappears. It seems like it shows only 11 of the 12 lines initially, hence the scrollbar, and that goes away when the editor takes over.

I don't know if this only started in this PR or not, so feel free to leave it for later.

frontend/index.js
```diff
@@ -387,7 +390,9 @@ async function handleHashChange() {
   crumbs && updateBreadcrumbs(crumbs)
   clearOutput()
   await format()
-  editor.onUpdate(clearHash)
+  // Clear hash right away for the long ugly gzip/b64 content URL fragments,
+  // keep the fragment until the first edit for the rest.
+  opts.content ? clearHash : editor.onUpdate(clearHash)
```
camh-, line 395: This didn't clear the hash. I think you needed `clearHash()` (with parens).

frontend/index.js
juliaogris: Done.

frontend/index.js
juliaogris: argh :( thank you!!

juliaogris: I'm sorry I didn't test that.

juliaogris: The one down-side I see with this is that it the history.pushState mucks up the back-forth navigation with browser button, but maybe that's ok.

The other shitty thing I only notice now is that the menu bar doesn't sync with contents anymore. this needs more thinking / UX advice.

camh-: re: history - I'm not sure what was intended exactly, but I don't think I would add to the history when clearing the hash. I think history should contain your navigation path generally and no navigation was done.

re: menu bar - sorry I don't understand this. What is the "menu bar" (hamburger?) and what is out of sync?

juliaogris: re history: there is no option not to add a URL change to the history iiuc.
with `history.replaceState` the _last_ state gets replaced with the current one (e.g. no hash) and the previous entry, e.g. `#welcome` is "removed" from the history (not what we want) and with `history.pushState` you simple add a new entry. I think empty hash should be added to history and I don't actually understand how it couldn't be.

re menu bar: the bread crumbs, the thing on the top, like `Tour > Welcome`.

frontend/index.html
juliaogris: ok, I will try to look into this. My scrollbars don't show up (only once I start scrolling). I think that's why I've missed it. I think this is new.

juliaogris: ok, so I can reproduce this on Firefox on MacOS (not Chrome). I believe it already has been a thing previously: would you mind checking https://evy.dev with throttled network & no-cash? 
It is a bit different now though:

Previously the loading sequence was like this:

1. Load welcome source code with syntax colouring (uneditable) in index.html
2. Once editor and all JS is ready create editor with come source and syntax color it.
3. Load current example (depending on URL fragment).

We have no changed this to:

1. Load welcome source code with syntax colouring (uneditable) in index.html
    if there is anything in the sessionStorage load that without syntax colouring.
3. Load current example (depending on URL fragment) and load it straight into the 
editor.

juliaogris: I have not managed to find any hints on how to remove the temporary scrollbar flicker on firefox :(

juliaogris: I found this SO post: https://stackoverflow.com/questions/77453048/how-can-i-prevent-the-scrollbar-from-fading-in-firefox

It has a super simple index.html page demonstrating the issue:

```
<!doctype html>
<html lang="en">
  <head></head>
  <body>
    <div style="overflow: auto; max-width: 100px; max-height: 100px">
      dfkdjkfdfjkdkjfdfjkkkkkkkkkkkkkkkk dfdjkfdkjf dfjkdjfdf fdjk
    </div>
  </body>
</html>
```

In chrome on MacOS (for me) no scrollbars. On firefox: scrollbar shows and disappears.
Not really sure what to do about this.

juliaogris: I can make scrollbars disappear altogether on firefox with `scrollbar-width: none;` as part of the style, but I think that's not great either.
I think the moment you try scrolling it is nice for scrollbars to show.

frontend/index.js
juliaogris: HOK.
Long story short, I might have found a workaround. I'm not sure if this makes things better... it's all added to yet another last commit. please let me know what you think.

frontend/index.html
juliaogris: Long story short, I might have found a workaround. I'm not sure if this makes things better... it's all added to yet another last commit. please let me know what you think.

juliaogris:
@camh- in case you are looking at this atm, i just realised I introduced a new bug trying to remove the flicker: session restore on reload is currently broken. Man frontend coding is hard. :(

frontend/index.js
juliaogris: maybe we need to discuss this in person. I understand your comment on the navigation path but I don't think this is an option unfortunately....

frontend/index.html
camh-: That all works better. No flickering. No scrollbar. Although I still only see up to "11" and then the "12" appears when the editor takes over. Not sure why, but I wouldn't worry about this.

frontend/index.js
camh-: I think it's all fine. I'm no longer sure what I was talking about. But having a play and the navigation seems fine.

camh- (approved):
:green_heart:

juliaogris:
This was a GitHub outage. This PR got merged not closed.

PR #231, opened 2024-01-02
github: Deploy to prod and stage on merge to main

juliaogris:
Deploy to prod and now also to stage on merge to main. We do this via
separate GitHub environments: prod and stage and different GitHub actions
job using different `FIREBASE_SERVICE_ACCOUNT_EVY_LANG` secrets wired to
the corresponding firebase prod and stage environments.

When the CI job (not the release job!) is run on `main`, rather than
pull-request, we deploy to https://evystage.dev and its live channel,
rather than the PR preview channel. We don't post preview channels to any
PR. This is done via updates to the firebase-deploy script.

While at it, add a minor, non-functional improvement to sed expression in
firebase-deploy script.

I've setup stage and prod GitHub environments `prod` and `stage` separately via
clicky-clicky-admin, hopefully using the correct Firebase Service Account secrets 
and the correct branch protection rule for prod.

---

Fixes https://github.com/evylang/todo/issues/46

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -8,6 +8,7 @@ on:
 jobs:
   ci:
     runs-on: ubuntu-latest
+    environment: stage
     steps:
       - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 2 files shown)


camh- (approved):
:cucumber:

PR #232, opened 2024-01-04
frontend: Self-host webfonts

juliaogris:
Self-host the webfonts Fira Code and Inter used in Evy projects. We want to
 do this so that Evy can be run entirely without internet access.

 In a preparatory step move `index.css` to `css/index.css` - we will splitting
 index.css into a lot more files and it seems to be common practise in
 frontend dev to groups in directories similar to `css/`, `img/`, `js/` etc.

---

Fixes: evylang/todo#42


camh- (approved):
:green_book: Interesting that apple was the last to support these fonts (woff2). Must have been quite some tension between "we love design and publishing and fonts and things" and "we don't like other peoples stuff. our way or the highway"

juliaogris:
@camh- merge emoji 🌞, because every commit towards self-hosting is less cloudy.

PR #234, opened 2024-01-05
deploy: Simplify deployment process

juliaogris:
Simplify deployment process by copying files from `frontend/**` to
`out/firebase/public/**` once only. Copy all firebase configs to out/firebase
and run deployment from there. Clean up various little make things along the
way.


camh- (approved):
:cucumber:

PR #235, opened 2024-01-07
hermit: Upgrade servedir to 0.0.3

camh-:
Upgrade servedir to 0.0.3 so I can make use of its new feature to
configure it via environment variables. Because my web browser is on a
different machine to where I do development (my server), using
`make serve` does not work for me with the defaults - I want to set the
port and specify `-a`. This has always meant keeping a dirty `Makefile`
around with my parameters on the servedir invocation.

Now I can use `direnv` and put a `.envrc` file in my evy directory that
contains my config for how I want to run servedir and not keep a dirty
Makefile around (I have `.envrc` in my git global ignore since I never
want to commit that).


juliaogris (approved):
Anything that makes frontend dev or debugging easier for you 🍭 
:peacock:

PR #236, opened 2024-01-07
frontend: Refactor confetti and Yace modules

juliaogris:
Rename YaceEditor to Editor as it is increasingly becoming customized and
tailored to evy. Pull out confetti Easter egg into its own module and remove
dependency on separate CSS file. In last commit re-work confetti Easter egg
for readability.

### DIFF
--- a/frontend/css/index.css
+++ b/frontend/css/index.css
@@ -789,23 +789,6 @@ embed {
   margin-right: auto;
 }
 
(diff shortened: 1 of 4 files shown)


camh- (approved):
🫛 This all reads pretty well I think. Perhaps the JS folks may object if some parts are not idiomatic, but I think it is idiomatic JS to find yet another way to do the same thing anyway so all good!

PR #237, opened 2024-01-08
housekeeping: Upgrade tooling and deps

juliaogris:
Upgrade tooling and go dependencies. I tried to write a make target for this,
but there seem to be too many manual steps involved - e.g. optionally
updating go.mod file when go minor version changes, or manually installing
most recent firebase and node version.

After the upgrade is done, finally pin the prettier version by providing
package.json file.


camh- (approved):
:broccoli: I have to assume the npx stuff is right and works. I could do some research, but sorry, couldn't be bothered :-|

PR #238, opened 2024-01-09
e2e: Initialize end to end testing

juliaogris:
Initialize end to end testing in top level directory e2e. 

**On CI e2 testing runs against https://evy.dev at the moment!**

Initialize `e2e` node package with

    echo '{"private": true}' > e2e/package.json
    npm install --save-dev --prefix e2e playwright
    npm install --save-dev --prefix e2e @playwright/test@latest

Add `node_modules/` to .gitignore.

Add a minimal `index.test.js` test file that loads the Evy playground `index.html`
page. It runs two tests on it:

- ensure the title is `"Evy | Playground"`
- execute `Run` and ensure the console output contains expected contents.

Setup basic playwright test config that will allow to parametrize the `baseURL`
for tests - it expects a `BASEURL` environment variable and falls back to `https://localhost:8080` by default. For local development run

    SERVEDIR_PORT=8080 make serve
 
and in other terminal

    cd e2e; npx playwright test [--ui]

Add `e2e` target to make file, triggered by `ci` target.
If `BASEURL` environment variable is not set, make will set it to `https://evy.dev` 
for now. This allows us to prototype e2e testing until we can extract the per-PR 
preview deployment URL from the firebase-deploy run. 
We will add this in a follow up PR.

Despite concerns it seems that the CI runtime has increased by only less than 30 seconds at this stage.

Part 1/2 for https://github.com/evylang/todo/issues/43

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ all: build test lint tiny test-tiny check-coverage sh-lint check-prettier check-
 	@if [ -e .git/rebase-merge ]; then git --no-pager log -1 --pretty='%h %s'; fi
 	@echo '$(COLOUR_GREEN)Success$(COLOUR_NORMAL)'
 
-ci: clean check-uptodate all ## Full clean build and up-to-date checks as run on CI
+ci: clean check-uptodate all e2e ## Full clean build and up-to-date checks as run on CI
 
 check-uptodate: tidy fmt doc
 	test -z "$$(git status --porcelain)" || { git status; false; }
@@ -127,6 +127,10 @@ godoc: install
 # --- frontend -----------------------------------------------------------------
 NODEPREFIX = .hermit/node
 NODELIB = $(NODEPREFIX)/lib
+# TODO: BASEURL should be calculated dynamically for CI/PR.
+# BASEURL needs to be in the environment so that `e2e/playwright.config.js`
+# can see it when the `e2e` target is called.
+export BASEURL = https://evy.dev
 
 ## Serve frontend on free port
 serve:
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -127,6 +127,8 @@ godoc: install
 # --- frontend -----------------------------------------------------------------
 NODEPREFIX = .hermit/node
 NODELIB = $(NODEPREFIX)/lib
+# TODO: BASEURL should be calculated dynamically for CI/PR.
+export BASEURL ?= https://evy.dev
```
camh-, line 131: Is it your intention that if BASEURL is set in the environment when `make` is invoked that the environment variable override this one? If not, there is no need for `?=`. You can use just the normal `=`.

camh-, line 131: I would add a comment as to why this is exported:

    # BASEURL needs to be in the environment so that `e2e/playwright.config.js` can see it
    # when the `e2e` target is called.

or something like that.

Makefile
```diff
@@ -140,10 +142,18 @@ prettier: | $(NODELIB)
 check-prettier: | $(NODELIB)
 	npx --prefix $(NODEPREFIX) -y prettier --check .
 
+# TODO: BASEURL should be calculated dynamically for CI/PR.
+e2e: E2EDIR=e2e
```
camh-, line 146: Is `E2EDIR` a variable just to reduce repetition? Or is there a use case for running `make e2e E2EDIR=foo` ? If the former I think I'd get rid of the variable and hard-code it:

	npm --prefix e2e ci
	npx --prefix e2e playwright install --with-deps chromium
	npx --prefix e2e playwright test --config e2e

But up to you.

Makefile
```diff
@@ -140,10 +142,18 @@ prettier: | $(NODELIB)
 check-prettier: | $(NODELIB)
 	npx --prefix $(NODEPREFIX) -y prettier --check .
 
+# TODO: BASEURL should be calculated dynamically for CI/PR.
```
camh-, line 145: This comment has moved further up and can be removed from here.

Makefile
juliaogris: yup, that was on purpose, I've been running it with:

    BASEURL=http://localhost:8080 make e2e

for local testing of the make target. Either way, I suspect it is temporary.

juliaogris: ok, done. thanks.

camh-: The risk here is that someone has `BASEURL` set in their environment for other reasons and it ends up overriding the value here. You should try to get used to setting make variables on the command line and not environment variables - i.e. `make e2e BASEURL=http://localhost:8080`. If you do that, you do not set the var in the environment before make, and make will override a var in the makefile of the form `BASEURL = ...` (i.e. a standard make assignment).

juliaogris: yeah, that makes sense, i've changed it to `=`

camh- (approved):

PR #239, opened 2024-01-09
make: Move help text to a separate line

juliaogris:
Move help text to a separate line as we have updated the awk script to
generate help. This was overlooked in a previous commit.


camh- (approved):
:green_book:

PR #242, opened 2024-01-12
deploy: Refactor deployment URLs in firebase-deploy

camh-:
Refactor the generation of the site deployment URLs for the GitHub
environment to remove the chunks of repetitive JSON at the top of the
script and generate the URLs directly without JSON for the live
deployments.

The static chunk of data at the top of the script was a little
in-your-face and seemed rather repetitive, and in hindsight, probably
unnecessary when a simple bit of code could replace it.

Remove the `environment` field from the `ci` GitHub actions job as it
is not needed in that job as deployment was moved to another job. This
reduces some of the "spam" on the PR every time you push.

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -8,7 +8,6 @@ on:
 jobs:
   ci:
     runs-on: ubuntu-latest
-    environment: stage
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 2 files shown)


juliaogris (approved):
Thanks! definitely cleaner.
I learned: `${sub^^}` to capitalize and `${!domains[*]}` to print associative array - i wonder why the `!`.
💚

camh-:
> i wonder why the `!`.

the `!` is used in other contexts as an indirect reference: `foo=1; bar=foo; echo ${!bar} # echos 1` - that doesn't make sense with array expansion, so I guess the re-used it.

juliaogris (approved):
lgtm

PR #244, opened 2024-01-15
Split the cicd workflow into stage and prod workflows

camh-:
Split up the cicd workflow so that a PR does not show a bunch of
irrelevant skipped steps on it. The new `stage` workflow is triggered on
a pull request and can be called from another workflow. The `prod`
workflow is triggered on a push to `main`, calls the `stage` workflow
and does the deployment to prod.

Docs: https://docs.github.com/en/actions/using-workflows/reusing-workflows


camh-:
I did a test of temporarily modifying `prod.yaml` to be run on this branch as well as `main` and disabled the actual prod deployment jobs just to test that the workflow calling was working ok and the secrets were accessible to the jobs. All looks good, so undrafting now.

.github/workflows/prod.yaml
```diff
@@ -54,8 +39,8 @@ jobs:
 
   howl-on-fail:
     runs-on: ubuntu-latest
-    needs: [ci, release]
-    if: ${{ always() && github.event_name == 'push' && ( needs.ci.result == 'failure' ||  needs.release.result == 'failure' ) }}
+    needs: [deploy-stage, release, deploy-prod]
+    if: failure()
```
juliaogris, line 43: Are you sure this is correct? I think there was a reason why id didn't use it in the first place, but not sure anymore.... anyway, doesn't matter i have to rework this soon.

juliaogris (approved):
I so love this! So good. Thank you Cam for sharing you GHA YAML skills. 💚

.github/workflows/prod.yaml
camh-: I'm not completely sure - I thought it had something to do with some checks being skipped due to the conditional logic, but with the removal of the conditional jobs, I figured `failure()` is simpler.

PR #246, opened 2024-01-15
ci: Fix prod workflow to checkout on deploy

camh-:
Checkout the repository on the prod/deploy job so that it can actually
work. This was split off the release job which does the checkout - the
split needed to add the checkout to the new job.

### DIFF
--- a/.github/workflows/prod.yaml
+++ b/.github/workflows/prod.yaml
@@ -30,6 +30,9 @@ jobs:
     needs: release
     environment: prod
     steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
       - run: ./bin/make deploy-prod
         env:
           FIREBASE_SERVICE_ACCOUNT: ${{ secrets.FIREBASE_SERVICE_ACCOUNT_EVY_LANG }}


.github/workflows/prod.yaml
```diff
@@ -30,6 +30,9 @@ jobs:
     needs: release
     environment: prod
     steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
camh-, line 35: we use `fetch-depth: 0` everywhere (meaning checkout all history), but I think it might only be needed for `release`? i'll leave this here for now, but perhaps we should check all of these out (pun intended).

juliaogris: It is need for deploy targets as well as it creates version.json

juliaogris: I think could do without on ci maybe. Not sure. Will try on tractor

juliaogris (approved):

.github/workflows/prod.yaml
camh-: It is needed for ci, because that does a build and it uses `$(VERSION)` in a linker flag which is based on the tags which need history.

PR #247, opened 2024-01-20
github: Re-instate howl job

juliaogris:
Re-instate howl job so we are notified on Slack and Discord when an evy
build on the `main` branch breaks. Instead of using the GitHub action we
are using the hermitised howl command so that it is more obvious on how to
run it locally.

---

I will drop the second commit as soon as I can show things are set up properly.


.github/workflows/prod.yaml
```diff
@@ -43,9 +43,14 @@ jobs:
...
+    if: always() && contains(join(needs.*.result, ','), 'failure')
     steps:
-      - uses: foxygoat/howl@v1
+      - uses: actions/checkout@v4
+      - run: ./bin/howl
         env:
-          SLACK_TOKEN: ${{ secrets.SLACK_TOKEN }}
-          SLACK_TEXT: <!here|here>
+          # Notify foxygoat Slack workspace on channel #gh-notify
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
```
camh-, line 52: I think you should only send to one place. It's annoying to get multiple notifications for the same thing. I would do discord only.

camh- (approved):
:leafy_green:

.github/workflows/prod.yaml
juliaogris: ok.

camh- (approved):
:leafy_green:

juliaogris:
Fixes https://github.com/evylang/todo/issues/4

PR #249, opened 2024-01-21
bytecode: Create bytecode package

Rossiar:
Create a bytecode package that compiles the parser AST into bytecode
and executes it in a virtual machine. Currently this initial
implementation only supports the simplest evy program; an inferred
type declaration from a constant, followed by an assignment.

### DIFF
--- a/pkg/bytecode/code.go
+++ b/pkg/bytecode/code.go
@@ -1,0 +1,98 @@
+package bytecode
+
+import (
+	"encoding/binary"
+	"errors"
+	"fmt"
+)
+
+const (
+	// OpConstant defines a constant that will be referred to by index
+	// in the bytecode.
+	OpConstant Opcode = iota
+	// OpGetGlobal retrieves a symbol from the symbol table at the
+	// specified index.
+	OpGetGlobal
+	// OpSetGlobal adds a symbol to the specified index in the symbol
+	// table.
+	OpSetGlobal
+)
+
+// ErrUnknownOp is returned when an unknown opcode is encountered.
+var ErrUnknownOp = errors.New("unknown opcode")
+
+// definitions is a mapping of OpCode to OpDefinition.
+var definitions = map[Opcode]*OpDefinition{
+	// The definition for OpConstant says that its only operand is two
+	// bytes wide, which makes it a uint16. This means that an evy
+	// bytecode program can only have 65535 constants defined.
+	OpConstant:  {"OpConstant", []int{2}},
+	OpGetGlobal: {"OpGetGlobal", []int{2}},
+	OpSetGlobal: {"OpSetGlobal", []int{2}},
+}
+
+// OpDefinition defines a name and expected operand width for each OpCode.
+type OpDefinition struct {
+	Name          string
+	OperandWidths []int
+}
+
+// Opcode defines the type of operation to be performed when reading
+// the bytecode.
+type Opcode byte
+
+// Make assembles and returns a single bytecode instruction out
+// of an Opcode and an optional list of operands. An error will
+// be returned if there is no definition for the provided Opcode.
+func Make(op Opcode, operands ...int) ([]byte, error) {
+	def, err := Lookup(op)
+	if err != nil {
+		return nil, err
+	}
+	instructionLen := 1
+	for _, w := range def.OperandWidths {
+		instructionLen += w
+	}
+	instruction := make([]byte, instructionLen)
+	instruction[0] = byte(op)
+	offset := 1
+	for i, o := range operands {
+		width := def.OperandWidths[i]
+		if width == 2 {
+			binary.BigEndian.PutUint16(instruction[offset:], uint16(o))
+		}
+		offset += width
+	}
+	return instruction, nil
+}
+
+// ReadOperands will read from the provided Instructions based on the
+// width of the provided OpDefinition. It returns the operands and the
+// read offset inside the Instructions.
+func ReadOperands(def *OpDefinition, ins Instructions) ([]int, int) {
+	operands := make([]int, len(def.OperandWidths))
+	offset := 0
+	for i, width := range def.OperandWidths {
+		if width == 2 {
+			operands[i] = int(ReadUint16(ins[offset:]))
+		}
+		offset += width
+	}
+	return operands, offset
+}
+
+// ReadUint16 is exposed to allow reading without performing a Lookup to
+// get an OpDefinition to pass to ReadOperands.
+func ReadUint16(ins Instructions) uint16 {
+	return binary.BigEndian.Uint16(ins)
+}
+
+// Lookup returns an OpDefinition for the value of an Opcode, any
+// unknown value will result in an error.
+func Lookup(op Opcode) (*OpDefinition, error) {
+	def, ok := definitions[op]
+	if !ok {
+		return nil, fmt.Errorf("%w: %d", ErrUnknownOp, op)
+	}
+	return def, nil
+}
--- a/pkg/bytecode/code_test.go
+++ b/pkg/bytecode/code_test.go
@@ -1,0 +1,61 @@
+package bytecode
+
+import (
+	"testing"
+
+	"evylang.dev/evy/pkg/assert"
+)
+
+func TestMake(t *testing.T) {
+	tests := []struct {
+		op       Opcode
+		operands []int
+		expected []byte
+	}{
+		{OpConstant, []int{65534}, []byte{byte(OpConstant), 255, 254}},
+	}
+	for _, tt := range tests {
+		instruction, err := Make(tt.op, tt.operands...)
+		assert.NoError(t, err)
+		assert.Equal(t, tt.expected, instruction)
+	}
+}
+
+func TestInstructionsString(t *testing.T) {
+	instructions := []Instructions{
+		mustMake(t, OpConstant, 2),
+		mustMake(t, OpConstant, 65535),
+	}
+	expected := "0000 OpConstant 2\n0003 OpConstant 65535\n"
+	concatted := Instructions{}
+	for _, ins := range instructions {
+		concatted = append(concatted, ins...)
+	}
+	assert.Equal(t, expected, concatted.String(), "instructions wrongly formatted.")
+}
+
+func TestReadOperands(t *testing.T) {
+	tests := []struct {
+		op        Opcode
+		operands  []int
+		bytesRead int
+	}{
+		{OpConstant, []int{65535}, 2},
+	}
+	for _, tt := range tests {
+		instruction, err := Make(tt.op, tt.operands...)
+		assert.NoError(t, err)
+		def, err := Lookup(tt.op)
+		assert.NoError(t, err)
+		operandsRead, read := ReadOperands(def, instruction[1:])
+		assert.Equal(t, read, tt.bytesRead, "wrong num bytes read")
+		assert.Equal(t, tt.operands, operandsRead, "wrong operands")
+	}
+}
+
+func mustMake(t *testing.T, op Opcode, operands ...int) []byte {
+	t.Helper()
+	ins, err := Make(op, operands...)
+	assert.NoError(t, err)
+	return ins
+}
--- a/pkg/bytecode/compiler.go
+++ b/pkg/bytecode/compiler.go
@@ -1,0 +1,120 @@
+package bytecode
+
+import (
+	"errors"
+	"fmt"
+
+	"evylang.dev/evy/pkg/parser"
+)
+
+// ErrUndefinedVar is returned when a variable name cannot
+// be resolved in the symbol table.
+var ErrUndefinedVar = errors.New("undefined variable")
+
+// Compiler is responsible for turning a parsed evy program into
+// bytecode.
+type Compiler struct {
+	constants    []value
+	instructions Instructions
+	globals      *SymbolTable
+}
+
+// Bytecode represents raw evy bytecode.
+type Bytecode struct {
+	Constants    []value
+	Instructions Instructions
+}
+
+// NewCompiler returns a new compiler.
+func NewCompiler() *Compiler {
+	return &Compiler{
+		constants:    []value{},
+		instructions: Instructions{},
+		globals:      NewSymbolTable(),
+	}
+}
+
+// Compile accepts an AST node and renders it to bytecode internally.
+func (c *Compiler) Compile(node parser.Node) error {
+	switch node := node.(type) {
+	case *parser.Program:
+		return c.compileProgram(node)
+	case *parser.InferredDeclStmt:
+		return c.compileDecl(node.Decl)
+	case *parser.AssignmentStmt:
+		return c.compileAssignment(node)
+	case *parser.Var:
+		return c.compileVar(node)
+	case *parser.NumLiteral:
+		num := &numVal{V: node.Value}
+		if err := c.emit(OpConstant, c.addConstant(num)); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Bytecode renders the compiler instructions into Bytecode.
+func (c *Compiler) Bytecode() *Bytecode {
+	return &Bytecode{
+		Instructions: c.instructions,
+		Constants:    c.constants,
+	}
+}
+
+// addConstant appends the provided value to the constants
+// and returns the index of that constant.
+func (c *Compiler) addConstant(obj value) int {
+	c.constants = append(c.constants, obj)
+	return len(c.constants) - 1
+}
+
+// addInstruction appends bytes to the instruction set and returns the
+// position of the instruction.
+func (c *Compiler) addInstruction(ins []byte) {
+	c.instructions = append(c.instructions, ins...)
+}
+
+// emit makes and writes an instruction to the bytecode and returns the
+// position of the instruction.
+func (c *Compiler) emit(op Opcode, operands ...int) error {
+	ins, err := Make(op, operands...)
+	if err != nil {
+		return err
+	}
+	c.addInstruction(ins)
+	return nil
+}
+
+func (c *Compiler) compileProgram(prog *parser.Program) error {
+	for _, s := range prog.Statements {
+		if err := c.Compile(s); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (c *Compiler) compileDecl(decl *parser.Decl) error {
+	if err := c.Compile(decl.Value); err != nil {
+		return err
+	}
+	symbol := c.globals.Define(decl.Var.Name)
+	return c.emit(OpSetGlobal, symbol.Index)
+}
+
+func (c *Compiler) compileAssignment(stmt *parser.AssignmentStmt) error {
+	if err := c.Compile(stmt.Target); err != nil {
+		return err
+	}
+	symbol := c.globals.Define(stmt.Value.String())
+	return c.emit(OpSetGlobal, symbol.Index)
+}
+
+func (c *Compiler) compileVar(variable *parser.Var) error {
+	symbol, ok := c.globals.Resolve(variable.Name)
+	if !ok {
+		return fmt.Errorf("%w %s", ErrUndefinedVar, variable.Name)
+	}
+	return c.emit(OpGetGlobal, symbol.Index)
+}
--- a/pkg/bytecode/doc.go
+++ b/pkg/bytecode/doc.go
@@ -1,0 +1,14 @@
+// Package bytecode contains a custom bytecode compiler and
+// virtual machine.
+//
+// Initially, an Evy program is turned into a sequence of tokens by the
+// [lexer]. Then, the [parser] generates an Abstract Syntax Tree (AST)
+// from the tokens. The [Compiler] of this package walks the AST and
+// writes the instructions to custom bytecode. For more on the AST
+// refer to the [parser] package documentation.
+//
+// The [VM] can read the bytecode and execute the encoded instructions,
+// providing a runtime similar to the [evaluator]. The virtual machine
+// is a straight-forward stack implementation that does not use any
+// additional registers.
+package bytecode
--- a/pkg/bytecode/instructions.go
+++ b/pkg/bytecode/instructions.go
@@ -1,0 +1,41 @@
+package bytecode
+
+import (
+	"bytes"
+	"fmt"
+)
+
+// Instructions represents raw bytecode, which is composed of opcodes
+// and an optional number of operands after each opcode.
+type Instructions []byte
+
+// String prints opcodes and their associated operands.
+func (ins Instructions) String() string {
+	var out bytes.Buffer
+	for i := 0; i < len(ins); {
+		def, err := Lookup(Opcode(ins[i]))
+		if err != nil {
+			panic(err.Error())
+		}
+		operands, read := ReadOperands(def, ins[i+1:])
+		fmt.Fprintf(&out, "%04d %s\n", i, fmtInstruction(def, operands))
+		i += 1 + read
+	}
+	return out.String()
+}
+
+func fmtInstruction(def *OpDefinition, operands []int) string {
+	operandCount := len(def.OperandWidths)
+	if len(operands) != operandCount {
+		return fmt.Sprintf("ERROR: operand len %d does not match defined %d\n",
+			len(operands), operandCount)
+	}
+	switch operandCount {
+	case 0:
+		return def.Name
+	case 1:
+		return fmt.Sprintf("%s %d", def.Name, operands[0])
+	default:
+		return fmt.Sprintf("ERROR: unhandled operandCount for %s\n", def.Name)
+	}
+}
--- a/pkg/bytecode/symbol.go
+++ b/pkg/bytecode/symbol.go
@@ -1,0 +1,46 @@
+package bytecode
+
+// SymbolScope defines a type of scope that a symbol can be defined inside.
+type SymbolScope string
+
+const (
+	// GlobalScope is the top level scope of an evy program.
+	GlobalScope SymbolScope = "GLOBAL"
+)
+
+// Symbol is a variable inside an evy program.
+type Symbol struct {
+	Name  string
+	Scope SymbolScope
+	Index int
+}
+
+// SymbolTable is a mapping of string identifiers to symbols.
+type SymbolTable struct {
+	store map[string]Symbol
+}
+
+// NewSymbolTable returns a new SymbolTable.
+func NewSymbolTable() *SymbolTable {
+	return &SymbolTable{
+		store: make(map[string]Symbol),
+	}
+}
+
+// Define adds a symbol definition to the table or returns an
+// already defined symbol with the same name.
+func (s *SymbolTable) Define(name string) Symbol {
+	if existing, found := s.store[name]; found {
+		return existing
+	}
+	symbol := Symbol{Name: name, Index: len(s.store), Scope: GlobalScope}
+	s.store[name] = symbol
+	return symbol
+}
+
+// Resolve returns the Symbol with the specified name,
+// or false if there is no such Symbol.
+func (s *SymbolTable) Resolve(name string) (Symbol, bool) {
+	obj, ok := s.store[name]
+	return obj, ok
+}
--- a/pkg/bytecode/vm.go
+++ b/pkg/bytecode/vm.go
@@ -1,0 +1,95 @@
+package bytecode
+
+import (
+	"errors"
+)
+
+const (
+	// StackSize defines an upper limit for the size of the stack.
+	StackSize = 2048
+	// GlobalsSize is the total number of globals that can be specified
+	// in an evy program.
+	GlobalsSize = 65536
+)
+
+// ErrStackOverflow is returned when the stack exceeds its size limit.
+var ErrStackOverflow = errors.New("stack overflow")
+
+// VM is responsible for executing evy programs from bytecode.
+type VM struct {
+	constants    []value
+	globals      []value
+	instructions Instructions
+	stack        []value
+	// sp is the stack pointer and always points to
+	// the next value in the stack. The top of the stack is stack[sp-1].
+	sp int
+}
+
+// NewVM returns a new VM.
+func NewVM(bytecode *Bytecode) *VM {
+	return &VM{
+		constants:    bytecode.Constants,
+		globals:      make([]value, GlobalsSize),
+		instructions: bytecode.Instructions,
+		stack:        make([]value, StackSize),
+		sp:           0,
+	}
+}
+
+// Run executes the provided bytecode instructions in order, any error
+// will stop the execution.
+func (vm *VM) Run() error {
+	for ip := 0; ip < len(vm.instructions); ip++ {
+		// This loop is the hot path of the vm, avoid unnecessary
+		// lookups or memory movement.
+		op := Opcode(vm.instructions[ip])
+		switch op {
+		case OpConstant:
+			constIndex := ReadUint16(vm.instructions[ip+1:])
+			ip += 2
+			err := vm.push(vm.constants[constIndex])
+			if err != nil {
+				return err
+			}
+		case OpGetGlobal:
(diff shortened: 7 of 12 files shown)


pkg/code/code_test.go
```diff
@@ -0,0 +1,52 @@
+package code
+
+import (
+	"testing"
+
+	"evylang.dev/evy/pkg/assert"
+)
+
+func TestMake(t *testing.T) {
```
Rossiar, line 9: Make should ideally return an `error` when:
- there is no definition for the provided op
- an invalid operand width is given

This extra return makes compiler tests much more noisy - perhaps a MustMake  test helper that fails the test if an error is thrown?

pkg/compiler/compiler.go
```diff
@@ -0,0 +1,119 @@
...
+	case *parser.Program:
+		return c.compileProgram(node)
+	case *parser.InferredDeclStmt:
+		return c.compileDecl(node.Decl)
+	case *parser.AssignmentStmt:
+		return c.compileAssignment(node)
+	case *parser.Var:
+		return c.compileVar(node)
+	case *parser.NumLiteral:
+		integer := &evaluator.NumVal{V: node.Value}
```
Rossiar, line 51: `integer` is a bad name, rename to `num` and add a test to prove that we support floating point numbers

pkg/code/code_test.go
juliaogris: Seems like a good idea.

pkg/compiler/compiler.go
juliaogris: Agreed.

pkg/compiler/symbol_table.go
```diff
@@ -0,0 +1,44 @@
...
+// NewSymbolTable returns a new SymbolTable.
+func NewSymbolTable() *SymbolTable {
+	s := make(map[string]Symbol)
+	return &SymbolTable{store: s}
+}
+
+// Define adds a symbol definition to the table, or increments the index of an already defined symbol.
+func (s *SymbolTable) Define(name string) Symbol {
+	symbol := Symbol{Name: name, Index: s.numDefinitions, Scope: GlobalScope}
+	s.store[name] = symbol
```
juliaogris, line 34: Without reading the compiler book: it is not clear to us why a symbol gets overriden and how `numDefinitions` is used. will get decremented?

pkg/compiler/symbol_table_test.go
```diff
@@ -0,0 +1,45 @@
+package compiler
+
+import "testing"
+
+func TestDefine(t *testing.T) {
+	expected := map[string]Symbol{
+		"a": {Name: "a", Scope: GlobalScope, Index: 0},
+		"b": {Name: "b", Scope: GlobalScope, Index: 1},
```
juliaogris, line 8: Would we want a symbol a with an already used name, e.g. `"a"`?

pkg/compiler/symbol_table_test.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	global := NewSymbolTable()
+
+	a := global.Define("a")
+	if a != expected["a"] {
```
juliaogris, line 14: `assert.Equals(t, want, got, messageAndArgs...)`

pkg/vm/vm_test.go
```diff
@@ -0,0 +1,64 @@
...
+		stackElem := vm.lastPoppedStackElem()
+		testExpectedObject(t, tt.expected, stackElem)
+	}
+}
+
+func testExpectedObject(
+	t *testing.T,
+	expected interface{},
+	actual evaluator.Value,
+) {
```
juliaogris, line 45: single line func signature if possible.

pkg/vm/vm_test.go
```diff
@@ -0,0 +1,64 @@
...
+		testIntegerObject(t, expected, actual)
+	default:
+		t.Errorf("unexpected object type %v", expected)
+	}
+}
+
+func testIntegerObject(t *testing.T, expected float64, actual evaluator.Value) {
+	t.Helper()
+	result, ok := actual.(*evaluator.NumVal)
+	if !ok {
```
juliaogris, line 60: assert.Equals(t, true, ok, messageAndArgs...)

pkg/compiler/compiler_test.go
```diff
@@ -0,0 +1,107 @@
...
+			expectedConstants: []interface{}{2},
+			expectedInstructions: []code.Instructions{
+				code.Make(code.OpConstant, 0),
+				code.Make(code.OpSetGlobal, 0),
+				code.Make(code.OpGetGlobal, 0),
+				code.Make(code.OpSetGlobal, 1),
+			},
+		},
+	}
+	runCompilerTests(t, tests)
```
juliaogris, line 41: maybe fold `runCompilerTests` in here.
some people pull out the testCases as global vars, especially when used in different test functions.

pkg/compiler/compiler_test.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func runCompilerTests(t *testing.T, tests []compilerTestCase) {
+	t.Helper()
```
juliaogris, line 45: should probably not be a t.Helper() - we want to know which assertion fails.

pkg/compiler/compiler_test.go
```diff
@@ -0,0 +1,107 @@
...
+		bytecode := compiler.Bytecode()
+		testInstructions(t, tt.expectedInstructions, bytecode.Instructions)
+		testConstants(t, tt.expectedConstants, bytecode.Constants)
+	}
+}
+
+func testInstructions(
+	t *testing.T,
+	expected []code.Instructions,
+	actual code.Instructions,
```
juliaogris, line 61: single line function signature.

pkg/compiler/compiler_test.go
```diff
@@ -0,0 +1,107 @@
...
+) {
+	t.Helper()
+	concatted := concatInstructions(expected)
+	assert.Equal(t, len(concatted), len(actual), "wrong instructions length")
+	for i, ins := range concatted {
+		if actual[i] != ins {
```
juliaogris, line 67: just assertEquals (t, insWant, actual[i], "someMessage %q", details )

pkg/compiler/compiler_test.go
```diff
@@ -0,0 +1,107 @@
...
+		out = append(out, ins...)
+	}
+	return out
+}
+
+func testConstants(
+	t *testing.T,
+	expected []interface{},
+	actual []evaluator.Value,
+) {
```
juliaogris, line 85: single line

juliaogris: maybe call `assertConstants` ?

pkg/compiler/compiler_test.go
```diff
@@ -0,0 +1,107 @@
...
+			testIntegerObject(t, float64(constant), actual[i])
+		case float64:
+			testIntegerObject(t, constant, actual[i])
+		default:
+			t.Errorf("unknown constant type %v", constant)
+		}
+	}
+}
+
+func testIntegerObject(t *testing.T, expected float64, actual evaluator.Value) {
```
juliaogris, line 100: `assertNumValue`

juliaogris (commented):
♥️❤️🧡💛💚💙💜💖♥️💖💜💙💚💛🧡❤️♥️❤️🧡💛💚💙💜💖♥️💖💜💙💚💛🧡❤️♥️❤️🧡💛💚💙💜💖♥️  
♥️ My first piece of feedback is a huge __THANK YOU__ for your contributions.  
♥️ So awesome what you have achieved in this short time!  
♥️❤️🧡💛💚💙💜💖♥️💖💜💙💚💛🧡❤️♥️❤️🧡💛💚💙💜💖♥️💖💜💙💚💛🧡❤️♥️❤️🧡💛💚💙💜💖♥️

You have given us a meaty, first PR - so naturally there are quite a few
things we have noted.

As mentioned initially, we are trying hard to build a codebase that aims for
high readability and learnability. Above all, it should be consistent with
its (unwritten 😅) guidelines. We understand that some of these nitpicks might
seem tedious, and there may be aspects you disagree with.

Please feel free to push back or ask for the rationale behind choices you
disagree with. (I'm not providing reasons for everything here because this
post is already quite long.)


### The English bits: Commits, PR description, godocs

Follow the [Chris Bean commit style](https://cbea.ms/git-commit/):
Some guidelines to keep in mind:
* Capitalize the subject line
* Use the imperative mood in the subject line
* Wrap the body at 72 characters
* Use the body to explain what and why vs. how 

E.g:

	vm: initial implementation

	Add an initial vm implementation that reads bytecode written by the compiler and executes it. Currently it can only support the simplest evy program which is an inferred type declaration to a constant, followed by an assignment.

=>

	vm: Create vm package

	Create a vm package that executes bytecode created by the compiler. Currently
	it can only support the simplest evy program: an inferred type declaration
	from a constant, followed by an assignment. We are putting structure in place
	here which we will expand.

(Imperative mood, captial letter on beginning of message after `label:`, line 
wrap, some kind of a why).

PR title and description become the merge commit - they matter! Therefore PR
titles should follow commit title guidelines and PR descriptions should
follow the commit body guidelines (72 char wrap), talk about why.

We have a
[merge script](https://github.com/foxygoat/git-scripts/blob/master/git-pr-merge)
that formats the merge commits and adds an emoji - this is for the merge
commit only! You can run `git log --oneline` on the evy repo to see how this
looks with single-commit PRs squash merged, and multi-commit PRs with the
emoji followed by the commits in the PR (`--graph` shows this more clearly
and `--first-parent` removes the noise of individual commits showing just the
PRs with emojis).

If you have non-permanent comments in the PR description, e.g.: "This PR is 
temporarily deployed at https://evystage.dev", post them past a `---`, those 
sections do not get copied into the merge commit.

### godoc comments in code

Please wrap at 72 char.

### Filenames
Don't use `-` or `_` in *.go file names unless used with intention, e.g. `*_test.go`, `*_unix.go`, `_ignoreme.go`.

### Code style

Don't use line breaks within function call (on commas), one func call on one
line. No need for line wrapping of go code. As a dyslexic, this makes code
more parse-able: one line is one idea; a function call is one idea. If for
whatever reasons it really does feel too long, break it up in a different
way, e.g. assign parameters to variables first. Occasional lines with 120
characters are fine. Map and array literals can be split over multiple lines,
preferrably only in assignments or inferred declarations.

### Package structure

We think it might be worthwhile pulling everything together under a `bytecode`
package:

	bytecode (with OpCodes)
	  ↳ bytecode/compiler
	  ↳ bytecode/vm

We hope to have more and different compilers in the future (directly to wasm,
llvm), which should equally be bundled together.

We probably need a package `abi` that contains the `evaluator.Value` interface
and probably the builtins and `evaluator.Runtime` interface should go in a `builtins` 
package. This is too much for the scope of this PR but should be probably done 
by us or you in the near future. It probably needs more thought.

### Testing

Ideally `TestSomething` is a self-contained function, so that tests can be read
without much jumping around - it's ok if these functions get big, they are
usually quite linear. Helpers with `t.Helper()` don't report the where the 
error happens so they should conceptually be a single assertion and also called 
that e.g. 

```go
func assertNumValue(t *testing.T, want float64, got Value) {
	t.Helper()		
	// ...
}
```

### Detailed implementation review

There were a few more detailed code review things, but it will be easier if the 
stuff mentioned here gets addressed first.

Last, one general observation: it would probably be great to somehow hack in
the `print` builtin sooner rather than later so evy programs can actually be
tested end-to-end with the compiler for expected output.

Rossiar:
@juliaogris thanks for the feedback! I have consolidated the packages into a `bytecode`  package and fixed up the commit message and some low hanging fruit (wrap godoc comments at 72 char). I will address the other pieces of feedback in further commits tomorrow using the `--fixup` strategy that was suggested in discord

pkg/compiler/compiler_test.go
Rossiar: I agree - but the `thelper` linter made me add `t.Helper()` to every private test function that used `t *testing.T` - I think this is not a good rule because it removes the ability to have complex `assertX` test functions. From what I understand an appropriate place for `t.Helper()` statement would be somewhere like the `run` function in `evaluator_test.go` if it used a testing.T to assert failures.

pkg/compiler/symbol_table.go
Rossiar: I think it's basically just there because you cannot `len` a map, all we want is an increasing index. I was also confused as to why it overwrote symbols with a new index, perhaps just to avoid a lookup? I didn't find any special justification in the book, so we should change the behaviour if it doesn't make sense. wdyt @joshcarp @pgmitche

Rossiar:
hi @juliaogris I used the fixup strategy to address all the :+1: comments (I find commenting Done to be quite spammy for the reviewer). I've left some comments on the compiler tests unaddressed because I think there is a slightly larger discussion there (see https://github.com/evylang/evy/pull/249#discussion_r1463949959) before I refactor.

pkg/compiler/symbol_table.go
juliaogris: ok, so just a question of - let's role with what the book does for now.
fwiw, you can `len` maps, unless I misunderstand: https://go.dev/play/p/JrKPpLRMUfd ?

pkg/compiler/compiler_test.go
juliaogris: I'm not sure I agree. IMO a `TestSomething` function should do _all_ the testing without calling any other functions, other than "simple" helpers that conceptually do a "single" assertion. IMO you want to know where exactly in your `*_test.go` file an assertion fails, without having to walk a call-stack within the test file. I don't recall I've ever been linter-forced to place a `t.Helper()` where it didn't make sense to me. 

WRT to the `runVMTest` function - it is currently only called in one place, I'd inline it there. If there are more tests to come that'd call the same function, I'd still inline it everywhere despite the duplication. I guess if these tests are really only about the VM you could have a `bytecode, err := compile(prog)` function that returns any error (parse or compile) over which you run assertNoError; or you do `bytecode := mustCompile(prog)` if these steps should really never fail. Not sure if this is reasonable. I guess at this stage we are in the realm of taste.
WDYT @camh- ?

pkg/compiler/symbol_table_test.go
juliaogris: I meant: should we add a test case that has two symbols with the same name, just to address the concerns raised earlier: should entries in a symbol table ever be overridden. Maybe all a bit out of context, feel free to ignore.

juliaogris:
As discussed on GitHub, we will keep on going here, even though it might be a bit tedious and have a slow turnaround. Here are a few more things to discuss:

* do we need evaluator.value (can you copy and keep private?)
* why does the symbolTable allow for overriding
* can we rework tests a little (see comments)?
* can we remove underscores in filenames please (symbol_table.go)

Rossiar:
> As discussed on GitHub, we will keep on going here, even though it might be a bit tedious and have a slow turnaround. Here are a few more things to discuss:
> 
> * do we need evaluator.value (can you copy and keep private?)

I reverted the change to evaluator and copied just the value interface and the numval type, which are all that we need for now

> * why does the symbolTable allow for overriding

I'm not sure, so I have removed this for now, if I find that it is needed at a later time I can add it back

> * can we rework tests a little (see comments)?

done

> * can we remove underscores in filenames please (symbol_table.go)

sure - renamed to symbol, what's the motivation? It might be worth adding a linter if you don't want underscores in filenames (beyond the `*_test.go` pattern)

let me know how it looks 🎉

pkg/bytecode/code.go
```diff
@@ -0,0 +1,96 @@
...
+		return nil, fmt.Errorf("%w: %d", ErrUnknownOp, op)
+	}
+	return def, nil
+}
+
+// Make assembles a single bytecode instruction out of an Opcode and
+// an optional list of operands.
+func Make(op Opcode, operands ...int) ([]byte, error) {
+	def, ok := definitions[op]
+	if !ok {
```
juliaogris, line 58: arguably you could/should use Lookup(op) for this? (keep errors in sync)

pkg/bytecode/code_test.go
```diff
@@ -0,0 +1,63 @@
...
+		assert.Equal(t, tt.expected, instruction)
+	}
+}
+
+func TestInstructionsString(t *testing.T) {
+	instructions := []Instructions{
+		mustMake(t, OpConstant, 2),
+		mustMake(t, OpConstant, 65535),
+	}
+	expected := "0000 OpConstant 2\n0003 OpConstant 65535\n"
```
juliaogris, line 29: this may read easier as:

```
expected := `
0000 OpConstant 2
0003 OpConstant 6553
`[1:]
```

just mentioning because there might be more like this coming.
up to you - don't feel strongly about this.

pkg/bytecode/doc.go
```diff
@@ -0,0 +1,15 @@
+// Package bytecode contains a custom bytecode compiler and
+// virtual machine.
+//
+// Initially, an Evy program is turned into a sequence of tokens by the
+// [lexer]. Then, the [parser] generates an Abstract Syntax Tree (AST)
+// from the tokens. The [Compiler] of this package walks the AST and
+// writes the instructions to custom bytecode. For more on the AST
+// please refer to the [parser] package documentation.
```
juliaogris, line 8: To keep tone consistent with other docs:

 For more on the AST see to the [parser] package documentation.

(I have not been very polite in the past and avoided pleases)

pkg/bytecode/doc.go
```diff
@@ -0,0 +1,15 @@
...
+//
+// The [VM] can read the bytecode and execute the encoded instructions,
+// providing a runtime similar to the [evaluator]. The virtual machine
+// is a straight-forward stack implementation that does not use any
+// additional registers. This leaves an initially simple implementation
+// that can be added to and optimized.
```
juliaogris, line 14: I'd remove the last sentence (actually don't quite get it).

pkg/bytecode/instructions.go
```diff
@@ -0,0 +1,43 @@
...
+
+// Instructions represents raw bytecode, which is composed of opcodes
+// and an optional number of operands after each opcode.
+type Instructions []byte
+
+// String prints opcodes and their associated operands.
+func (ins Instructions) String() string {
+	var out bytes.Buffer
+	i := 0
+	for i < len(ins) {
```
juliaogris, line 16: you can pull the previous 2 lines into one.
```go
for i := 0; i < len(ins ); {` 
```

pkg/bytecode/code_test.go
juliaogris: this also applies to various examples below. Maybe the single line solution for 2 things is preferable to the 4 line one, but I think for 3 or more things splitting over several lines with raw strings may be better.

juliaogris (commented):
I think this is heaps better than the first version.
I've added a couple more nits, but we're very close I think.

Could you please take the PR out of draft and squash your fixes into the single commit? I'll then go over it one more time hopefully with @camh- seeing as this is your first contribution I think its worth while not missing any conventions we have unconsciously formed.

WRT:
> sure - renamed to symbol, what's the motivation? It might be worth adding a linter if you don't want underscores in filenames (beyond the *_test.go pattern)

Here is some interesting reading on that subject: https://github.com/golang/go/issues/36060

> The Go project itself tends to prefer short names like io.go, pipe.go

It is not a documented standard, neither is using snake case filenames.
I personally like it as I am a slight dyslexic and read/skim very slowly. 
I see a filename with an underscore I know there is a special meaning to it.

I guess we could just make it an evy convention if that's acceptable?

pkg/bytecode/code.go
Rossiar: Yes good point, The book avoids this as a trade off for an "error free" Make function.

> Note that we don’t use the Lookup function to get to the definition, which gives us a much more usable function signature for Make in the tests later on. By circumventing Lookup and not having to return possible errors, we can use Make to easily build up bytecode instructions without having to check for errors after every call. The risk of producing empty byte slices by using an unknown opcode is one we’re willing to take,

But since we return an error now anyway it makes sense to use Lookup.

pkg/bytecode/code_test.go
Rossiar: In our dev branch we ended up with a mix of styles (https://github.com/evylang/evy/blob/compiler/pkg/vm/vm_test.go#L13 vs https://github.com/evylang/evy/blob/compiler/pkg/compiler/compiler_test.go#L18)

This PR is all consistently single line, which I think is fine for now and we can swap to multiline when/if needed - wdyt?

pkg/bytecode/vm.go
```diff
@@ -0,0 +1,89 @@
...
+		return ErrStackOverflow
+	}
+	vm.stack[vm.sp] = o
+	vm.sp++
+	return nil
+}
+
+func (vm *VM) pop() value {
+	o := vm.stack[vm.sp-1]
+	vm.sp--
```
camh-, line 87: This kind of leaks memory, as each stack element is an interface (`value`) which may hold a pointer to the underlying data (an evy array or map once you get to that). But this pop operation still leaves the stack slice holding the value, so the garbage collector will never clean it up. I've suggested using the stacks slice length as the stack pointer and using `slices.Delete()` in go 1.22 to pop which zeros out elements it removes. That's one alternative, but otherwise consider setting the stack element to nil when you pop it.

pkg/bytecode/code.go
```diff
@@ -0,0 +1,96 @@
...
+	OperandWidths []int
+}
+
+// Opcode defines the type of operation to be performed when reading
+// the bytecode.
+type Opcode byte
+
+// Lookup returns an OpDefinition for the value of an Opcode, any
+// unknown value will result in an error.
+func Lookup(op byte) (*OpDefinition, error) {
```
camh-, line 46: This name is a little generic package-wide - you also have a symbol table which is something you look up. Perhaps call it `LookupOpcode()`?

Also, why does it take a `byte` and not an `Opcode`? In the three uses of this function, two have an Opcode and cast it to a byte. The other arguably should cast `ins[i]` to `Opcode`. I guess there's always going to be a little tension between when  you use the raw bytes as it comes from the `Instructions` and when you use the higher level types for what those bytes represent. This may need some more thought to keep these layers clearly separated and to keep like code grouped together.

pkg/bytecode/compiler.go
```diff
@@ -0,0 +1,120 @@
...
+type Compiler struct {
+	constants    []value
+	instructions Instructions
+	symbolTable  *SymbolTable
+}
+
+// Bytecode represents raw evy bytecode.
+type Bytecode struct {
+	Constants    []value
+	Instructions Instructions
```
camh-, line 25: This Bytecode type is represented as the separate fields in two other places - `Compiler` and `VM`. It is not all that clear to me why you don't just use the `Bytecode` struct in those two places and just copy the single value rather than the individual fields.

pkg/bytecode/compiler.go
```diff
@@ -0,0 +1,120 @@
...
+// Compiler is responsible for turning a parsed evy program into
+// bytecode.
+type Compiler struct {
+	constants    []value
+	instructions Instructions
+	symbolTable  *SymbolTable
+}
+
+// Bytecode represents raw evy bytecode.
+type Bytecode struct {
```
camh-, line 23: `Bytecode` is a little stuttery given the name of the package (`bytecode`), but I admit after giving it a lot of thought, I couldn't come up with another name.

camh- (commented):
Some nice looking code, thanks. I'm excited to see the start of this - it would be nice to see evy run faster as we've already hit some limits with some of our graphic samples.

I've made a bunch of comments; please take them all as my opinion which you are free to ignore - just my thoughts as I read through the code.

I think the thing that tripped me up the most at first was the bit of fuzziness over bytes and more structured types and the lack of a type for a single `Instruction` as bytecode, and the `Instructions` type that seems to have fuzzy boundaries. But now I understand it I am struggling to come up with a better name, so :shrug: But perhaps something to keep in mind as you grow this and need to do more complex things - some slightly tighter abstractions may help.

I haven't looked over the tests yet, but Julia has so I might just leave that.

camh-:
Don't forget to update the PR description. It's now out of date with the changes you have made. This will become immortalised in the merge commit message when merged. With single-commit PRs, the PR message is usually just the commit message, and we just squash merge to avoid the useless merge commit, but since you've reworked this PR from multiple commits to one commit, that relationship has been broken. I'd just copy the commit message into the PR body.

pkg/bytecode/code.go
Rossiar: > This name is a little generic package-wide

I think it's fine as-is, symbol table uses Define/Resolve which is distinct enough from lookup.

> Also, why does it take a byte and not an Opcode?

On reflection I think it makes sense to change it to accept an Opcode.

pkg/bytecode/code.go
```diff
@@ -0,0 +1,96 @@
...
+		offset += width
+	}
+	return instruction, nil
+}
+
+// ReadOperands will read from the provided Instructions based on the
+// width of the provided OpDefinition.
+func ReadOperands(def *OpDefinition, ins Instructions) ([]int, int) {
+	operands := make([]int, len(def.OperandWidths))
+	offset := 0
```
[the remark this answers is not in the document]
Rossiar: Taking this to the discord for further discussion on `Instruction`

pkg/bytecode/compiler.go
Rossiar: I guess not using it avoids a layer on indirection in the compiler and vm, e.g. you have to go and look at bytecode to understand that it has instructions/constants and it's only really a wrapper for passing a bag of data from compiler -> vm. I don't feel particularly strongly about it so can change if you think that it is better.

pkg/bytecode/symbol.go
```diff
@@ -0,0 +1,45 @@
...
+// Symbol is a variable inside an evy program.
+type Symbol struct {
+	Name  string
+	Scope SymbolScope
+	Index int
+}
+
+// SymbolTable is a mapping of string identifiers to symbols.
+type SymbolTable struct {
+	store map[string]Symbol
```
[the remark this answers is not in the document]
Rossiar: Yes the intent is to have multiple nested symbol tables, renaming the variable is a good idea, but I don't think that it makes sense to add a comment about `SymbolTable` being nested when the code doesn't permit for that atm

pkg/bytecode/vm.go
```diff
@@ -0,0 +1,89 @@
...
+
+// Run executes the provided bytecode instructions in order, any error
+// will stop the execution.
+func (vm *VM) Run() error {
+	for ip := 0; ip < len(vm.instructions); ip++ {
+		op := Opcode(vm.instructions[ip])
+		switch op {
+		case OpConstant:
+			constIndex := ReadUint16(vm.instructions[ip+1:])
+			ip += 2
```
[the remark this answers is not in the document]
Rossiar: Again the book highlighted this:

> It’s important that we do not use code.Lookup here to get from a byte to an Opcode. That would be far too slow. It costs time to move the byte around, lookup the opcode’s definition, return it and take it apart. I know that this doesn’t sound like my usual chant of “we’re here to learn, not to build the fastest thing ever”, but we’re in the hot path here; everything we can get rid of, we should throw out. Using code.Lookup here would be like putting a sleep statement in the loop.

Perhaps we should add a comment and/or massage the code to enhance `ReadOperands` by passing a pre-allocated slice and removing the `Lookup`?

pkg/bytecode/vm.go
Rossiar: I think that `slices.Delete()` is going to reallocate the slice and have a (relatively) big performance hit on our vm. Right now the compiled programs we can write are so basic that there isn't a sensible way to benchmark this to compare, but I would be worried about performance. 

That being said leaking memory is definitely not ideal, but I am not entirely sure how gogc will behave once a pointer to a value is overwritten in the stack slice (remember that `push` is assigning values onto the stack so the index will quickly be filled with another value). Would it be worth pinning this until we can sensibly benchmark the vm?

pkg/bytecode/vm.go
```diff
@@ -0,0 +1,89 @@
...
+	if vm.sp >= StackSize {
+		return ErrStackOverflow
+	}
+	vm.stack[vm.sp] = o
+	vm.sp++
+	return nil
+}
+
+func (vm *VM) pop() value {
+	o := vm.stack[vm.sp-1]
```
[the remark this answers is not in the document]
Rossiar: Adding this would require error checking in every case that we pop. I think that it's better to add a specific panic for this case that lets the user know it was a stack underflow, rather than the generic "index out of bounds"

pkg/bytecode/vm.go
```diff
@@ -0,0 +1,89 @@
...
+
+// VM is responsible for executing evy programs from bytecode.
+type VM struct {
+	constants    []value
+	globals      []value
+	instructions Instructions
+	stack        []value
+	// sp is the stack pointer and always points to
+	// the next value in the stack. The top of the stack is stack[sp-1].
+	sp int
```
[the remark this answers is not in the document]
Rossiar: I think having `sp` aids clarity for the reader, instead of mentally decoding `len(stack)` vs the capacity - see my other comment regarding `slices.Delete()`

pkg/bytecode/vm.go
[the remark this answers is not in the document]
camh-: Fair enough. I'd add a comment along the lines:

    // This loop is the hot path of the vm, avoiding unnecessary lookups or memory movement.

That's enough I think to ensure a reader of the code has context to understand the open coding.

pkg/bytecode/vm.go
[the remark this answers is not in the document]
camh-: I wasn't thinking hot path when I wrote this. With that in mind, you could ignore the stack underflow with the comment:

> Ignore stack underflow as that indicates an error in the vm and the out-of-bounds slice panic is sufficient for that, as opposed to the stack overflow above which can occur due to a user program we're running.

pkg/bytecode/vm.go
camh-: I'd be surprised if `slices.Delete()` reallocated, but it is critical that the stack slice does not get reallocated so you're right it's a good idea to avoid that just in case.

You are right that often values on the stack will be overridden when more is pushed onto the stack, but it won't necessarily get back to that. The case I'm worried about is some nested function calls where the deepest one does something with a large map value, and after unwinding that stack, you never get back to that depth.

Don't worry about it for now, but when doing function calls, consider tracking the high water mark of the stack and when cleaning up after a function, you could clear the stack up to that high water mark. That way the clearing won't be on the hot path of vm.pop() and it only clears for the cases where it's likely to become a problem.

pkg/bytecode/code.go
camh-: `Lookup` is package scoped and exported so when I see `bytecode.Lookup()` in code, I do not immediately consider that it is looking up Opcodes in the definitions map, given that the `bytecode` package contains symbol tables, symbols, vms, compilers, instructions, etc. If this were in a package `opcode`, it would be fine IMHO.

My point about symbol table was that is a type in the package that very clearly screams "lookup" - being a table. Of all the types in the package, `SymbolTable` is the one I would most closely associate with the bare word `Lookup`, but that was more of an example of how the vagueness of the name `Lookup` could cause confusion, not something specifically about `SymbolTable`.

But now that the function takes an `Opcode` and not a `byte`, it is a bit clearer as callers to `Lookup` will either pass a var probably named `op` / `opcode` or if they have a `byte`, it would need to be `Opcode(byte)`.

So if you'd prefer to keep it `Lookup`, no worries.

pkg/bytecode/compiler.go
camh-: keep it as is. let's see where it goes.

camh- (approved):
All good. I've left some comments unresolved so you can see my response. When you've read them and agree or disagree, just resolve them ~so you can merge~ (please don't merge).

camh-:
Oh, sorry just realised. Please don't merge. We'll do that with our merge script, right @juliaogris ?

PR #251, opened 2024-01-23
docs: Fix build badge on README.md

juliaogris:
Fix the build badge in the README as it is currently orange and says "repo or
workflow not found". We have moved `cicd.yaml` to `stage.yaml` and
`prod.yaml` `prod.yaml` is the only workflow used on `main` so use it for
build status.


camh- (approved):
:green_square:

PR #252, opened 2024-01-23
make: Fix circular dependency in Makefile

camh-:
Fix the dependency that `check-coverage` and `cover` have on `test`.
This should be `test-go` as the `test` target was changed to `test-go`
in aa0139ed. The circular dependency is due to `make test` depending on
`check-coverage` which then depends on `test`, causing make to emit the
error:

    make: Circular check-coverage <- test dependency dropped.


juliaogris (approved):
🪲

PR #253, opened 2024-01-31
docs:  Add "papers" and "talks" pages

juliaogris:
Add "papers" and "talks" pages to documentation for reference and possible
later inclusion on docs.evy.dev.

In a follow up PR we will link to the pdf in a firebase re-write config using
the release tag as permalink.


camh- (approved):
:green_book:

PR #255, opened 2024-02-03
deploy: Rewrite live site generation in Go

camh-:
Rewrite the part of the `firebase-deploy` shell script that generates
the site in Go. To properly track the file renames in arbitrary
directories was going to get messy in shell, so this does it in Go.

The site generator does:
* Optionally rewrite URLs in `.html` files to some top-level paths to
  subdomains (discord, docs, learn, and play)
* Optionally rename `.js`, `.css` and `.wasm` files to include a
  cache-busting short hash of the content in the filename, and update
  any references to the renamed files in any `.html` files. These
  references include the stylesheet and javascript HTML references as
  well as an embedded javascript importmap and our custom embedded
  wasmImport map.

This adds a restriction that the site must not symlink directories -
instead it should create a directory and symlink the contents. Following
directory symlinks adds a lot of complexity for detecting loops,
recursively walking subtrees and tracking renames across that recursion
that it is much simpler to forbid directory symlinks.

The generated site can be used to deploy to a live firebase channel
(with subdomain rewriting and cache busting), to a PR channel (without
either) or to be embedded in a Go binary (all symlinks removed, no cache
busting or subdomain rewriting).

### DIFF
--- a/build-tools/firebase-deploy
+++ b/build-tools/firebase-deploy
@@ -13,18 +13,21 @@ main() {
 
 	environ=$(get_environment "${1:-stage}") || exit $?
 	channel=$(get_channel "${2:-}")
+
 	rm -rf out/firebase
 	cp -r firebase out/
-	cp -r frontend out/firebase/public
+
 	if [[ "${channel}" == "live" ]]; then
-		hash_css_js out/firebase/public
-		update_refs "${environ}"
+		local domain
+		[[ "${environ}" == "prod" ]] && domain="evy.dev" || domain="evystage.dev"
+		go run ./build-tools/site-gen --cache-bust --domain="$domain" frontend out/firebase/public
 		# `firebase deploy` must be used with live channel
 		firebase --config out/firebase/firebase.json --project "${environ}" deploy --only hosting
 		on_ci && make_env_urls "${environ}" >>"$GITHUB_ENV"
 		exit 0
 	fi
 
+	go run ./build-tools/site-gen frontend out/firebase/public
 	# `firebase hosting:channel:deploy` must be used with preview/non-live channels
 	result=$(firebase --json --config out/firebase/firebase.json --project "${environ}" hosting:channel:deploy "${channel}")
 	check_deploy_error "${result}"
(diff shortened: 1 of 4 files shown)


scripts/firebase-deploy
```diff
@@ -80,143 +82,6 @@ get_pr_num() {
...
-		hash=$(get_hash "${file}")
-		hashed_file="${file%.*}.${hash}.${file##*.}"
-		mv "${file}" "${hashed_file}"
-		ln -s "${hashed_file##*/}" "${file}"
-	done
-}
-
-get_hash() {
-	local file="$1"
-	#  Hashes with 32 bits, or 8 chars[0-9a-f] and 100 file changes in a year
```
camh-, line 96: I should probably transplant this comment to the Go code

scripts/gen-live-site/main.go
```diff
@@ -0,0 +1,271 @@
...
+}
+
+func copyHTMLFiles(src, dest, domain string, skippedFiles []string, renamedFiles map[string]string) error {
+	for _, filename := range skippedFiles {
+		in, out, err := openInOut(filepath.Join(src, filename), filepath.Join(dest, filename))
+		if err != nil {
+			return err
+		}
+		defer in.Close() //nolint:errcheck // don't care about close failing on read-only files
+		err = updateHTMLFile(out, in, filename, domain, renamedFiles)
```
juliaogris, line 144: before you return `in, out` and here you take `out, in` - on purpose?

scripts/gen-live-site/main.go
```diff
@@ -0,0 +1,271 @@
...
+}
+
+func openInOut(src, dest string) (io.ReadCloser, io.WriteCloser, error) {
+	in, err := os.Open(src)
+	if err != nil {
+		return nil, nil, err
+	}
+	out, err := os.Create(dest)
+	if err != nil {
+		in.Close() //nolint:errcheck,gosec // we're returning the more important error
```
juliaogris, line 180: I used `_ = in.Close` - removed the linter warning, is shorter and almost as obvious.
Same below.
Up to you if you prefer to keep.

scripts/gen-live-site/main.go
```diff
@@ -0,0 +1,271 @@
...
+		out.Close() //nolint:errcheck,gosec // we're returning the more important error
+		return err
+	}
+	return out.Close()
+}
+
+var (
+	subdomainRE = regexp.MustCompile(`(href|value)="/(discord|docs|learn|play)`)
+	jscssRefRE  = regexp.MustCompile(`(href|src)="(.*\.(?:css|js))"`)
+	importmapRE = regexp.MustCompile(`"(.*\.js)": "(.*\.js)"`)
```
juliaogris, line 203: seeing as this may become a slightly more robust/serious CLI now, I wonder if these REs should come via flags as well? probably not, just worthy of a thought

juliaogris (commented):
This is really nice and understandable to read. Thank you!
As discussed elsewhere: for file embedding for `evy serve` it would be great if this tool could optionally just copy from src to dest and remove all symlinks. 
Or I could just use `cp -rL frontend out/embed` I guess.
This means that hashing & all other file replacements, including subdomain replacements should ideally be optional

scripts/gen-live-site/main.go
camh-: yeah sort of. the out, in is the "normal" order it seems (although I've always preferred the other way - but look at `io.Copy`. but `openOutIn` just felt wrong, but I can change it if you like

scripts/gen-live-site/main.go
camh-: I like to be explicit when I'm silencing the linter. I dont like to alter my program just to make the linter happy, and when I do the linter thing I get to put my comment on it as to why.

build-tools/firebase-deploy
```diff
@@ -13,18 +13,21 @@ main() {
...
+
 	rm -rf out/firebase
 	cp -r firebase out/
-	cp -r frontend out/firebase/public
+
 	if [[ "${channel}" == "live" ]]; then
-		hash_css_js out/firebase/public
-		update_refs "${environ}"
+		local domain
+		[[ "${environ}" == "prod" ]] && domain="evy.dev" || domain="evystage.dev"
```
juliaogris, line 22: Ah, so this is just a little bit of extra safety saying:
it's either gotta be evystage.dev or prod&evy.dev
I like that.

scripts/gen-live-site/main.go
camh-: I'm not sure - they do not stand alone - the replacement part uses them but constructs the replacement string programmatically. There does appear to be an opportunity to refactor the replacement stuff but I have not come up with a way that is clear, so I left the duplication of the 3 replacements in place. If I could make that generic and clear, I would consider making the regexps user-pluggable but I'll leave it for now.

build-tools/site-gen/main.go
```diff
@@ -0,0 +1,320 @@
+// Command site-gen generates the website files to be deployed to firebase.
+//
+// Usage: site-gen <src-dir> <dest-dir> <domain>
+//
+// When deploying to firebase (any other hosting site), we need to make a few
+// changes to the HTML, CSS and JS files in the site:
+//   - Replace href/values with leading paths of /discord, /docs, /learn and /play
+//     with a subdomain instead, so /docs/foo with docs.<domain>/foo
+//   - Rename .css, .js and wasm files to include a short-sha of the SHA256 of the
```
juliaogris, line 9: .wasm ?

build-tools/site-gen/main.go
```diff
@@ -0,0 +1,320 @@
...
+	"path"
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"github.com/alecthomas/kong"
+)
+
+type app struct {
+	CacheBust bool   `help:"Rename .css and .js file to include short hash"`
```
juliaogris, line 43: .css, .js and .wasm files ?

build-tools/site-gen/main.go
```diff
@@ -0,0 +1,322 @@
...
+	Domain    string `help:"Rewrite top-level paths to subdomains"`
+	SrcDir    string `arg:"" required:""`
+	DestDir   string `arg:"" required:""`
+}
+
+func main() {
+	cfg := app{}
```
juliaogris, line 50: note to self: `app` is probably the better type name.
I should probably fix this up in `/main.go`

If we stick with this convention (over `config` and `cfg`)
maybe it should be `a := app{}` ?

build-tools/firebase-deploy
camh-: this was the same logic as before, just condensed. It used to live in `update_refs` in the shell script.

build-tools/site-gen/main.go
camh-: renamed `cfg` to `a`

scripts/gen-live-site/main.go
juliaogris: no, I understand now. sounds good.

scripts/gen-live-site/main.go
juliaogris: Fair enough, I will update this in `evy serve` PR

scripts/gen-live-site/main.go
juliaogris: ok, agreed.

scripts/gen-live-site/main.go
juliaogris: ok, understand. let's leave as is.

juliaogris (approved):
🦚

PR #257, opened 2024-02-04
frontend: Redirect evy.wasm filename to via wasmImports map

juliaogris:
Redirect evy.wasm filename to via wasmImports map so that we cache-bust
evy.wasm in the same way we are doing for the js files in the importmap.

This is a preparatory PR for PR #255 to also include evy.wasm cache-busting.


frontend/play/evy.wasm
juliaogris, line 1: woops, that got accidentally added a couple of PRs ago.

frontend/play/index.html
```diff
@@ -12,8 +12,14 @@
...
     <script type="importmap"></script>
-    <script src="wasm_exec.js" defer></script>
-    <script src="index.js" type="module" defer></script>
+    <script>
+      const wasmImports = {
+        // Right is replaced for cache busting, e.g. "./module/evy.wasm": "./module/evy.11738cdd.wasm"
+        "./module/evy.wasm": "./module/evy.wasm",
+      }
+    </script>
+    <script src="wasm_exec.js" type="module"></script>
```
juliaogris, line 21: I removed the `defer` here - apparently that's redundant with `type="module"`. if you want I'll move it into a separate commit.

frontend/play/evy.wasm
juliaogris: nope, this is just a "dangling" link. nothing bad happened here.

frontend/play/index.html
juliaogris: ok, I've moved this into a separate commit.

camh- (approved):
:broccoli:

PR #263, opened 2024-02-06
frontend: Move css/play.css to play/css/index.css

juliaogris:
Move css/play.css to play/css/index.css. This was a workaround when we had the
bash script doing all firebase deployment and cache busting. Now that all of
this is addressed and fixed in the beautiful build-tools/site-gen go program
we can do this fancy stuff. Yay 🎉.


camh- (approved):
:green_book:

Issue #262, opened 2024-02-06
Fix missing error on invalid eventhandler signature


juliaogris:
If an event-handler has an additional parameter browser evy crashes :( e.g.

```
on move x:num y:num z:num
    print x y z
end
```

Note that the following work:

```
on move x:num y:num
    print x y
end
```

```
on move x:num _:num
    print x
end
```

```
on move
    print "hello"
end
```

And the following give the expected error messages:

```
on move x:num y:string
    print x y
end
```
```
on move x:num
    print x
end
```

PR #268, opened 2024-02-09
parser: Return early on wrong parameters count

juliaogris:
Return early on wrong event handler parameter count. We track the correct
error but still carry on parsing the parameters in further details which
causes an index out of range exception.

Issue: https://github.com/evylang/evy/issues/262

---

Fixes https://github.com/evylang/evy/issues/262


camh- (approved):
:four_leaf_clover:

PR #270, opened 2024-02-10
build: Generate headings with anchors

juliaogris:
Generate headings with anchors with md tool, in the same way GitHub creates
anchors on .md pages. Use the same style of IDs: heading text with all
non-letter/numbers replaced by single dash.

In preparatory step rework `walk` function in the Markdown generation. We now
allow for nodes not only to be modified but also to be replaced or deleted.

For what we are trying to achieve here this might be over-kill, but it is
preparatory work for building a multiple-choice test generator.

It seems more correct to me to add a new type of node called anchoredHeading,
as it seems to be a different thing to an ordinary heading. in the markdown
page, Tasks are also explicitly modeled when used instead of unordered
lists.

I'll put up an alternative PR that is just straight-forward overriding the
markdown.Heading.PrintHTML function for comparison.

Link: https://stackoverflow.com/questions/43273842/what-are-the-rules-of-converting-one-markdown-title-into-an-html-anchor/44443100#44443100
Issue: https://github.com/evylang/todo/issues/61

---

Fixes https://github.com/evylang/todo/issues/61


build-tools/md/anchor.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"bytes"
+	"fmt"
+
+	"rsc.io/markdown"
+)
+
+// anchoredHeading produces HTML with clickable link anchor, for example:
+//
+//	<h2 class="heading">
```
camh-, line 12: should be `class="anchored"`

build-tools/md/main.go
```diff
@@ -140,18 +141,42 @@ func md2html(mdBytes []byte) (string, string) {
...
 
-func walkFn(n node) node {
+type walker struct {
+	headingIDs map[string]bool
+}
+
+func (w *walker) walk(n node) node {
 	if mdl, ok := n.(*markdown.Link); ok {
 		return updateLink(mdl)
 	}
```
camh-, line 157: This can be removed now you have it in the type switch

build-tools/md/main.go
```diff
@@ -140,33 +141,62 @@ func md2html(mdBytes []byte) (string, string) {
...
-	if !ok {
-		return
+type walker struct {
+	headingIDs map[string]bool
+}
+
+func (w *walker) walk(n node) node {
+	if mdl, ok := n.(*markdown.Link); ok {
+		return updateLink(mdl)
+	}
```
juliaogris, line 157: needs to be removed. on it.

frontend/preview/development/releasing.html
```diff
@@ -3,7 +3,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
-    <title>evy docs · Releasing</title>
```
camh-, line 6: I think `extractHeading()` is now broken. It type asserts blocks to `(*markdown.Heading)` and I guess they are not that any more, even though your new heading embeds this.

build-tools/md/main.go
```diff
@@ -140,33 +141,62 @@ func md2html(mdBytes []byte) (string, string) {
...
+func (w *walker) updateHeading(h *markdown.Heading) node {
+	return newAnchoredHeading(h, w.headingID(h))
+}
+
+var reHeadingID = regexp.MustCompile(`[^\pL0-9]+`)
+
+func (*walker) headingID(h *markdown.Heading) string {
+	id := inlineText(h.Text.Inline)
+	id = strings.ToLower(id)
+	id = reHeadingID.ReplaceAllString(id, "-")
```
juliaogris, line 176: ok, and this needs to increment trailing number if ID found.

build-tools/md/main.go
juliaogris: yup already made a note to that extend but I guess that's not visible on the commit?

frontend/preview/development/releasing.html
juliaogris: yeah, need to fix this, 💰

build-tools/md/main.go
camh-: The note was not there when I reviewed it / made my comment.

build-tools/md/main.go
juliaogris: done.

juliaogris:
Closing this as it seems too complicated for what we are trying to achieve here in favor of #272 .

PR #273, opened 2024-02-12
frontend: Fix ios rendering bug

juliaogris:
Fix ios rendering bug where the height of an `<embed>` was different on
different browsers - it was higher on Safari and lower on FF and Chrome. The
result was a landing page with just a big evy-e and nothing else visible.
Thank you so much @starkcoffe for pointing this out.

---

I think I need to talk to Jas about the landing page heading thing.


e2e/test.js-snapshots/playground-chromium-linux.png
juliaogris, line 1: No idea what's different here 🤷‍♀️

e2e/test.js-snapshots/landing-top-chromium-linux.png
juliaogris, line 1: Heading moved a little to the left.

e2e/test.js-snapshots/landing-bottom-chromium-linux.png
juliaogris, line 1: This test became flaky as part of this PRs change, so I updated the test.

e2e/test.js-snapshots/landing-top-ios-linux.png
juliaogris, line 1: And this is what it's really all about.

e2e/test.js
```diff
@@ -12,9 +12,14 @@ test("landing", async ({ page, baseURL }, testInfo) => {
   await expect(page).toHaveScreenshot("landing-top.png")
 
   if (testInfo.project.name != "ios") {
+    // let's make sure we don't take snapshots of the gif
```
camh-, line 15: I don't understand this comment (what gif)?. But perhaps it is better to say what you *are* testing rather than what you are not.

camh-: ok, i didn't read very clearly. I see this test is testing multiple images on the page and that this part is the playground displayed on the landing page.

camh- (approved):

PR #274, opened 2024-02-13
bytecode: Implement addition and subtraction

Rossiar:
These simple operations for integer mathematics will
be expanded upon in the next change, which will add
the multiplication, division and modulo operations.
This change also standardises the compiler and vm test
layouts to increase their readability.


pkg/bytecode/vm.go
```diff
@@ -63,11 +71,50 @@ func (vm *VM) Run() error {
...
+	rightType := right.Type()
+	leftType := left.Type()
+
+	// more cases to come (string, bool).
+	if leftType == parser.NUM_TYPE && rightType == parser.NUM_TYPE {
+		leftValue := left.(*numVal).V
+		rightValue := right.(*numVal).V
+		return vm.executeBinaryNumOperation(op, leftValue, rightValue)
+	}
+	return fmt.Errorf("%w %s %s", ErrUnsupportedType, leftType, rightType)
```
juliaogris, line 98: hmm, this should never be let through by the parser - maybe even a panic is justified here if that makes your code easier? Up to you.

juliaogris (commented):
Looks good to me just a couple of comments, mostly food for thought. 
I'll wait for @camh- if he's got anything to add again.

pkg/bytecode/vm.go
```diff
@@ -63,11 +71,50 @@ func (vm *VM) Run() error {
 			globalIndex := ReadUint16(vm.instructions[ip+1:])
 			ip += 2
 			vm.globals[globalIndex] = vm.pop()
+		case OpAdd, OpSubtract:
```
camh-, line 74: Since, as @juliaogris said, the type checking of the args should have been done by the parser, you could optimise all this as:
```
case OpAdd:
    right := vm.pop().(*numVal).V
    left := vm.pop().(*numVal).V
    vm.push(&numVal{V: left+right})
case OpSubtract:
    right := vm.pop().(*numVal).V
    left := vm.pop().(*numVal).V
    vm.push(&numVal{V: left-right})
```
For extra safety, you could define `vm.popNumVal()` that panics with a better error than the type assertion would (and probably good to make it panic with an `error` that `Is` some sort of `ErrInternal`).

Another comment I was going to make last PR that is relevant here now is `numVal` could just be a `float64` and not a struct with one. Get rid of the pointer too and it is likely that the value itself will fit inside the interface pair. Get rid of all the `.V` stuff, perhaps add a `newNumVal` to at least abstract the representation, and it would make the code look like:
```
case OpAdd:
    right := vm.popNum()
    left := vm.popNum()
    vm.push(newNum(left+right))
case OpSubtract:
    right := vm.popNum()
    left := vm.popNum()
    vm.push(newNum(left-right))
```

And since binary ops are rather common, you could define `popBinaryNums()` that is optimised as a single function call so you can do:
```
case OpAdd:
    right, left := vm.popBinaryNums()
    vm.push(newNum(left+right))
case OpSubtract:
    right, left := vm.popBinaryNums()
    vm.push(newNum(left-right))
```

Now we're getting tight on the hot path.

camh- (commented):
mostly comments around tests - I'm sorry I didn't get around to reviewing the tests last time, some of this I could have picked up then.

pkg/bytecode/vm.go
camh-: > Get rid of the pointer too and it is likely that the value itself will fit inside the interface pair.

I think I'm wrong about this. I think earlier versions of Go might have stored values in the `data` field of an `iface`, if they fit, but I don't think they do any more (in fact I can't find any information on that at all historically or not, so now I think I'm imagining it. Looks like a non-pointer won't save anything anyway. That said, I still think `float64` is a better underlying type for `numVal`, but if it has a constructor (and I've locally added an `asFloat64()` method) then that can isolate the representation, which might give you some flexibility in optimizing later..

Rossiar: It would be great if you could elaborate on the advantages that these changes would bring to the implementation, I see some downsides:

- drift with evaluator value (I still think it's a good idea to unify these in an abi package)
- vm is not protected against changes to the parser, e.g. if we ever decide that you can concatenate a num and a string this code will panic in a confusing manner instead of returning a proper error
- I think the code above becomes more complicated than the existing when string concatenation and boolean expressions are added

pkg/bytecode/vm.go
Rossiar: If the parser changes the vm will not behave correctly and will panic in unusual ways that could be harder to debug than just reading a clear error message.

I discussed panic vs error in our implementation with josh and patrick yesterday and we feel that error reads better and there is not a material difference to the user (they just see an error message either way).

pkg/bytecode/vm_test.go
```diff
@@ -7,10 +7,57 @@ import (
...
+		{
+			input: `
+			x := 1 + 2
+			x = x
+			`,
+			expected: 3,
+		},
+		{
+			input: `
+			x := 2 + 1
```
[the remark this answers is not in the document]
Rossiar: removed

pkg/bytecode/vm_test.go
```diff
@@ -7,10 +7,57 @@ import (
...
+		{
+			input: `
+			x := 2 - 1
+			x = x
+			`,
+			expected: 1,
+		},
+		{
+			input: `
+			x := 1 - 2
```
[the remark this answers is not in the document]
Rossiar: This tests minus numbers in the virtual machine and that left/right are not interchangeable for minus operations, which is clearer now that we have added names.

pkg/bytecode/vm_test.go
```diff
@@ -7,10 +7,57 @@ import (
...
 )
 
-func TestIntegerArithmetic(t *testing.T) {
+func TestVMArithmetic(t *testing.T) {
 	tests := []vmTestCase{
-		{"x := 1\nx = x", 1},
-		{"y := 2\ny = y", 2},
+		{
+			input: `
+			x := 1
```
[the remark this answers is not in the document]
Rossiar: moved to its own test function

pkg/bytecode/vm_test.go
```diff
@@ -7,10 +7,57 @@ import (
 	"evylang.dev/evy/pkg/parser"
 )
 
-func TestIntegerArithmetic(t *testing.T) {
+func TestVMArithmetic(t *testing.T) {
 	tests := []vmTestCase{
-		{"x := 1\nx = x", 1},
-		{"y := 2\ny = y", 2},
+		{
+			input: `
```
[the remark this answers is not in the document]
Rossiar: added

Rossiar (commented):
fixed most issues, some outstanding conversations to follow up on.

What do we think of unifying vm_test and compiler_test? The inputs and test cases are the same and vm_test runs the compiler - would this simplify or complicate things?

pkg/bytecode/vm_test.go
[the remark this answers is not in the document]
camh-: It is still the same test. If the previous test passed then you already know it is working in the right order. If the test were `1 - 1` then yeah, that would not tell you that you had the left/right order correct, but `2 - 1` evaluating to `1` means you have left/right correct, and this test does not add any new validation.

pkg/bytecode/vm.go
juliaogris: ok. Could you please add a base error `ErrInternal` similar to the [evaluator package](https://github.com/evylang/evy/blob/1937ad097b260891e106c93f3d2a0f78610d68d8/pkg/evaluator/evaluator.go#L28) so that we separate internal errors more easily from other unavoidable user caused runtime errors?

pkg/bytecode/vm.go
juliaogris: Let's talk briefly about error checking / validation.

In general, in my view, errors should only be checked once at the boundaries of a "component". To me the "component" is the Evy toolchain.

If you feel strongly that parser Errors should be confirmed again in the `bytecode`
package then they should only be checked at (bytecode) compile time and not at (bytecode) execution time, because:

* Execution is as you called it the "hot path", don't do avoidable steps there.
* Don't duplicate checks for readability (less code is more readable especially to beginners).
* Don't duplicate checks because of `evy.wasm` size concerns. In general we want to avoid sending evy toolchain bytes over the network to the evy.dev site if they can be avoided.

To respond to your points in specific:

* **drift with evaluator value**: I will wear all consolidation efforts. TBH the evaluator.numVal etc. should just be a float 64 and not a struct wrapping a float64, my bad, this was an oversight. I have raise an [evaluator cleanup issue](https://github.com/evylang/todo/issues/66).
* **vm is not protected against changes to the parser**: The parser and vm are delivered together, if the language changes everything needs to be update. We are writing to the evy [spec](https://github.com/evylang/evy/blob/main/docs/spec.md#operators-and-expressions), let's not worry about possible future changes. 
If you insist on type checks in the bytecode package they should happen in the compile step not in the hotloop of the execution.
* **I think the code above becomes more complicated than the existing when string concatenation and boolean expressions are added**

So I've followed along and written it _all_ out as Cam suggested and to me this reads direct and obvious - I don't need to hop to helper functions to understand what's going on.

```go
case OpAdd:
    right, left := vm.popBinaryNums()
    vm.push(newNum(left+right))
case OpSubtract:
    right, left := vm.popBinaryNums()
    vm.push(newNum(left-right))
case OpMulitply: //...
case OpDivide: //...
case OpModulo: //...
case OpNumEQ:
    right, left := vm.popBinaryNums()
    vm.push(newBool(left == right))
case OpNumNEQ:
    right, left := vm.popBinaryNums()
    vm.push(newBool(left != right))
case OpNumLT:
    right, left := vm.popBinaryNums()
    vm.push(newBool(left < right))
case OpNumLTE:
    right, left := vm.popBinaryNums()
    vm.push(newBool(left <= right))
case OpNumGT:
    right, left := vm.popBinaryNums()
    vm.push(newBool(left > right))
case OpNumGTE:
    right, left := vm.popBinaryNums()
    vm.push(newBool(left >= right))
case OpStringConcat:
    right, left := vm.popBinaryString()
    vm.push(newString(left+right))
case OpStringEQ:
    right, left := vm.popBinaryStrings()
    vm.push(newBool(left == right))
case OpStringNEQ:
    right, left := vm.popBinaryStrings()
    vm.push(newBool(left != right))
case OpStringLT:
    right, left := vm.popBinaryStrings()
    vm.push(newBool(left < right))
case OpStringLTE:
    right, left := vm.popBinaryStrings()
    vm.push(newBool(left <= right))
case OpStringGT:
    right, left := vm.popBinaryStrings()
    vm.push(newBool(left > right))
case OpStringGTE:
    right, left := vm.popBinaryStrings()
    vm.push(newBool(left >= right))
case OpArrayConcat:
    right, left := vm.popBinaryArray()
    vm.push(newArray(left+right))
case OpArrayEQ: 
    right, left := vm.popBinaryArray()
    vm.push(newBool(left.Equals(right)))
case OpArrayNEQ: 
    right, left := vm.popBinaryArray()
    vm.push(newBool(!left.Equals(right)))
case OpMapEQ: 
    right, left := vm.popBinaryArray()
    vm.push(newBool(left.Equals(right)))
case OpMapNEQ: 
    right, left := vm.popBinaryArray()
    vm.push(newBool(!left.Equals(right)))
case OpAnyEQ: // ...
case OpAnyNEQ: // ...
```

camh-:
> What do we think of unifying vm_test and compiler_test? The inputs and test cases are the same and vm_test runs the compiler - would this simplify or complicate things?

Agree 100%. I was going to comment on this but felt I had said enough already. I'm glad you brought it up.

This is what we've done elsewhere - more of end-to-end testing that unit testing. As you point out, the test cases are the same and they will surface the same errors. If you plan to keep the compiler and vm in the same package, this is easy since it is logically valid to have a single test that covers both. I originally had a mental model of `package bytecode` containing the representation of the bytecode, `bytecode/compiler` that generates a stream of bytecode and `bytecode/vm` that executes a stream of bytecode. With that split it becomes a little weird that tests in the `bytecode/vm` pacakge would catch errors in the compiler. I'm not sure if you're considering a structure like this at all, but I guess we'll deal with that when the time comes if the time comes.

pkg/bytecode/vm_test.go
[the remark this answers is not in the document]
camh-: Perhaps I misunderstood - did you mean with this test that you are testing negative results? If so, that also appears to be unnecessary as it is really testing that the Go compiler produces code that subtracts properly - nothing in the VM so far is specifically about negative numbers that I can see.

pkg/bytecode/vm.go
camh-: Yeah, pretty much what Julia said. The advantages are largely about a shorter hot path, but also simpler code, something we've been continuously striving for with evy. We would like evy itself to be quite readable and approachable so that evy becomes an accessible code base to learners who are ready to move past the evy language itself.

I felt that:
```
case OpAddition:
    right, left := vm.popBinaryNums()
    vm.push(newNum(left + right))
```

explains the operation quite succinctly.

I don't think you need to be doing type validation in the VM - the parser has already validated that, but if you do feel it still needs to be done, I'd do it in the compiler. I'd front-load as much as I could into the compiler to keep the VM lean and fast. I'm also making the assumption that there are no plans to load external bytecode (i.e. it's not the JVM for example where source is compiled to bytecode and later executed). Instead we'll be compiling and executing in one hit, so you should be able to trust the validation of the parser or compiler. If my assumption is wrong, then perhaps we should talk about the far plans sometime, but even then I'd be considering other approaches to run-time type validation.

I would also not be using the `OpAddition` opcode/operation to be doing string concatenation or array concatenation - I would have the compiler identify those operations from the AST and have separate ops for those two operations. Keep the opcodes simple like CPU instructions are. Don't decide at runtime one of many meanings of an opcode, instead decide at compile time and have many opcodes.

With regard to the ABI, the point I ended up at was that we don't yet know enough what that should look like. I would very much prefer you to write your bytecode code for itself and when that is done, we can see what might apply to an ABI. I think we're all learning here, so I prefer not to look too far ahead because its so easy to get that wrong without experience.

pkg/bytecode/code.go
```diff
@@ -29,6 +33,11 @@ var definitions = map[Opcode]*OpDefinition{
 	OpConstant:  {"OpConstant", []int{2}},
 	OpGetGlobal: {"OpGetGlobal", []int{2}},
 	OpSetGlobal: {"OpSetGlobal", []int{2}},
+	// Operations like OpAdd have no operand width because the virtual
+	// machine is expected to pop the values from the stack when reading
+	// this instruction.
+	OpAdd:      {"OpAdd", []int{}},
```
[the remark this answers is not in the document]
Rossiar: I had a brainfart here - will swap to nil slices

pkg/bytecode/vm_test.go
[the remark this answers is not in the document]
Rossiar: ok I can remove it

pkg/bytecode/vm.go
Rossiar: Thanks for all the feedback, I have incorporated these changes and will bear the style in mind as I continue 🙏

Rossiar (commented):
I have:
- merged the tests into vm_test 
- set operand widths to nil instead of empty slice
- refactored the vm to use panics instead of errors and added ErrInternal
- changed the binary expression handling to avoid helper methods
- changed numVal to be a float64 (in doing so I had to remove Set because you can't have a pointer receiver on a float64, I think this is ok because you can now just write `a = numVal(12)` to reset the value)

pkg/bytecode/code.go
```diff
@@ -16,10 +16,14 @@ const (
...
+	// OpAdd instructs the virtual machine to perform an addition.
+	OpAdd
+	// OpSubtract instructs the virtual machine to perform a subtraction.
+	OpSubtract
 )
 
-// ErrUnknownOp is returned when an unknown opcode is encountered.
-var ErrUnknownOp = errors.New("unknown opcode")
+// ErrUnknownOpcode is returned when an unknown opcode is encountered.
+var ErrUnknownOpcode = errors.New("unknown opcode")
```
juliaogris, line 26: This should wrap `ErrInternal`.
```go
var ErrUnknownOpcode = fmt.Errorf("%w: unknown opcode", ErrInternal)
```

pkg/bytecode/compiler.go
```diff
@@ -7,9 +7,14 @@ import (
...
-// ErrUndefinedVar is returned when a variable name cannot
-// be resolved in the symbol table.
-var ErrUndefinedVar = errors.New("undefined variable")
+var (
+	// ErrUndefinedVar is returned when a variable name cannot
+	// be resolved in the symbol table.
+	ErrUndefinedVar = errors.New("undefined variable")
+	// ErrUnknownOperator is returned when an operator cannot
+	// be resolved.
+	ErrUnknownOperator = errors.New("unknown operator")
```
juliaogris, line 16: `ErrUnknownOperator` should be an err `ErrInternal` IMO.
```go
var ErrUnknownOperator  = fmt.Errorf("%w: unknown operator", ErrInternal)
```

pkg/bytecode/vm.go
```diff
@@ -12,8 +13,14 @@ const (
 	GlobalsSize = 65536
 )
 
-// ErrStackOverflow is returned when the stack exceeds its size limit.
-var ErrStackOverflow = errors.New("stack overflow")
+var (
+	// ErrStackOverflow is returned when the stack exceeds its size limit.
+	ErrStackOverflow = errors.New("stack overflow")
```
juliaogris, line 18: IMO this should wrap `ErrPanic`, `ErrRuntime` or similar.

In the evaluator I called it [`ErrPanic`](https://github.com/evylang/evy/blob/main/pkg/evaluator/evaluator.go#L19), things that can go wrong during runtime
such as `ErrBounds` for index out of bounds.

Then you have 2 top level error types:

- `ErrInternal` - should never happen, we mucked something up.
- `ErrPanic` or similar - "user error".

pkg/bytecode/compiler.go
```diff
@@ -7,9 +7,14 @@ import (
 	"evylang.dev/evy/pkg/parser"
 )
 
-// ErrUndefinedVar is returned when a variable name cannot
-// be resolved in the symbol table.
-var ErrUndefinedVar = errors.New("undefined variable")
+var (
+	// ErrUndefinedVar is returned when a variable name cannot
+	// be resolved in the symbol table.
+	ErrUndefinedVar = errors.New("undefined variable")
```
juliaogris, line 13: `ErrUndefinedVar` should wrap a `ErrRuntime` or `ErrPanic` or similar:
```go
ErrUndefinedVar = fmt.Errorf("%w: undefined variable", ErrRuntime)
```

This specific error can actually happen in parsed evy code (it is not an internal error), e.g.:

```
b := makeNum
print "b" b
a := 3

func makeNum:num
    return a
end
```

A better parser (like Go's) should find this at parse time, but ours does not at the moment.

pkg/bytecode/vm.go
juliaogris: I'm looking at the evaluator again and I'm wondering why I called this sentinel
`ErrPanic`, especially as I also have a `PanicError` that is returned by Evy's
builtin `panic "boom"` function.

The problem with the name `ErrRuntime` is that while these errors happen at "execution time" and are not preventable (ie not internal), they don't align with what Runtime means in a different Evy execution context: 
In this codebase the Runtime means "environment providing hooks" -  browser or the terminal - one provides graphics functions and means to write to an HTML `<div>` the other one is purely text based using Stdin/Stdout. 

So if you have a good idea wrt to top level error grouping / naming - go ahead and I'll adjust inside the evaluator.

pkg/bytecode/vm.go
```diff
@@ -75,13 +82,12 @@ func (vm *VM) lastPoppedStackElem() value {
 	return vm.stack[vm.sp]
 }
 
-func (vm *VM) push(o value) error {
+func (vm *VM) push(o value) {
 	if vm.sp >= StackSize {
-		return ErrStackOverflow
+		panic(ErrStackOverflow)
```
juliaogris, line 87: This is a "user" error (should wrap `ErrPanic` or similar) and should therefore be returned imo.

juliaogris (commented):
looks pretty solid to me. Just a little bit around error grouping
sentinel errors.
All of this is foundational work for the `bytecode` package and this back and forth will hopefully soon drop away.

pkg/bytecode/vm.go
juliaogris: @Rossiar , I think this is unresolved -`ErrStackOverflow` is a "user error" and should be returned rather than panic'ed with - agreed?

juliaogris (commented):
ok, i'm happy with this - Thank You - there is only one last unresolved comment:
https://github.com/evylang/evy/pull/274#discussion_r1491705035

pkg/bytecode/vm.go
Rossiar: Sorry I forgot to comment here - how about a recover in `Run`? This keeps the code less cluttered and also collects index out of bounds and other unforeseen panics, and avoids the user seeing a wall of stack trace as an error

juliaogris: Yes, personally I really like this approach within a package, however unfortunately tinygo-wasm does not support recover. It's super annoying.
So recover is not an option at this stage I'm afraid :(

https://github.com/tinygo-org/tinygo/issues/2914

Rossiar: Ah what a shame, in that case I will put the error handling back

juliaogris: tbh - panics are not handled at all on evy web: https://github.com/evylang/todo/issues/68 I really need to sort this out.

juliaogris (approved):
ok, this is all good from my end now, but maybe @camh- has still something else to add? I'll leave it up to him.

camh- (approved):

PR #279, opened 2024-02-19
bytecode: Implement unary expressions

Rossiar:
This change adds the 'not' and 'minus' unary operators
to the compiler and vm.

Co-authored-by: joshcarp joshcarp@users.noreply.github.com
Co-authored-by: pgmitche pgmitche@users.noreply.github.com

### DIFF
--- a/pkg/bytecode/code.go
+++ b/pkg/bytecode/code.go
@@ -32,6 +32,12 @@ const (
 	OpTrue
 	// OpFalse represents the boolean literal false.
 	OpFalse
+	// OpNot represents the not unary operator, which performs logical
+	// negation on a boolean operand.
+	OpNot
+	// OpMinus represents the minus unary operator, which negates the
+	// value of a numeric operand.
+	OpMinus
 )
 
 var (
@@ -64,6 +70,8 @@ var definitions = map[Opcode]*OpDefinition{
 	OpModulo:   {"OpModulo", nil},
 	OpTrue:     {"OpTrue", nil},
 	OpFalse:    {"OpFalse", nil},
+	OpNot:      {"OpNot", nil},
+	OpMinus:    {"OpMinus", nil},
 }
 
 // OpDefinition defines a name and expected operand width for each OpCode.
--- a/pkg/bytecode/vm_test.go
+++ b/pkg/bytecode/vm_test.go
@@ -118,6 +118,22 @@ func TestBoolExpressions(t *testing.T) {
 				mustMake(t, OpSetGlobal, 0),
 			},
 		},
+		{
+			name: "not operator",
+			input: `
+			x := !true
+			x = x
+			`,
+			expectedStackTop:  false,
+			expectedConstants: []any{},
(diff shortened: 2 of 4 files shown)


pkg/bytecode/code.go
```diff
@@ -32,6 +32,12 @@ const (
 	OpTrue
 	// OpFalse represents the boolean literal false.
 	OpFalse
+	// OpBang represents the bang unary operator, which performs logical
+	// negation on a boolean operand.
+	OpBang
```
juliaogris, line 37: maybe this should be called `OpNot` ? no strong feelings, just and idea, seems closer to the actual operation than bang.

pkg/bytecode/vm_test.go
```diff
@@ -135,6 +135,66 @@ func TestBoolExpressions(t *testing.T) {
...
+			program, err := parser.Parse(tt.input, parser.Builtins{})
+			assert.NoError(t, err, "parser error")
+			comp := NewCompiler()
+			err = comp.Compile(program)
+			assert.NoError(t, err, "compiler error")
+			vm := NewVM(comp.Bytecode())
+			err = vm.Run()
+			assert.NoError(t, err, "runtime error")
+			stackElem := vm.lastPoppedStackElem()
+			switch expected := tt.expectedStackTop.(type) {
```
juliaogris, line 184: I think at this point I'd split this up into `TestBoolUnary` and `TestNumUnary` top level tests.

juliaogris (commented):
looks pretty good to me too, just a couple of thoughts.
waiting for @camh- here too.

pkg/bytecode/code.go
Rossiar: Good idea, I like that name better

pkg/bytecode/vm_test.go
Rossiar: Taking another look at this, I don't think it makes sense to highlight unary expression tests in their own methods. So I have just included them in the relevant bool or num tests

juliaogris (commented):
@Rossiar - if you rebase and take this out of draft, i think we are good.

pkg/bytecode/vm_test.go
```diff
@@ -118,6 +118,22 @@ func TestBoolExpressions(t *testing.T) {
 				mustMake(t, OpSetGlobal, 0),
 			},
 		},
+		{
+			name: "not operator",
+			input: `
+			x := !true
+			x = x
```
camh-, line 125: Consider a clean-up follow-up PR that removes all these `x = x` from the test cases as well as the `GetGlobal` / `SetGlobal` at the end of the `expectedInstructions` and instead have the test runner add these. This is how the evaluator works: https://github.com/evylang/evy/blob/e9e8840bba3957d00fd405af7b94929fe12eddd9/pkg/evaluator/evaluator_test.go#L53 in many places. Then the test input is a simple one-liner string in many cases and is generally cleaner.

camh- (approved):

PR #280, opened 2024-02-20
frontend: Update docs look

juliaogris:
Update docs look for https://docs.evy.dev to move a little closer to the
design. We are still not done so we haven't moved things to /docs yet, but we
got a little closer:

* header added, though
  - search not yet working
  - active page not yet highlighted
* heading styles closer to design
* ul styles closer to design

In order to add the header we pulled out a shared header component from
the playground into frontend/css/header.css.

Issue: https://github.com/evylang/todo/issues/65
Link: https://www.figma.com/file/4Wf4Mbq8I2Yxh2bptOnSE9/evy

---

preview at: https://evy-lang-stage--280-qnq572di.web.app/preview/


Makefile
```diff
@@ -230,6 +230,12 @@ run-playwright:
 ## Run end-to-end tests with playwright (see run-playwright)
 e2e: run-playwright
 
+## Run end-to-end tests and list failed snapshot test image files.
+e2e-diff: PLAYWRIGHT_ARGS = --reporter json
+e2e-diff:
+	$(PLAYWRIGHT_CMD) | \
```
camh-, line 236: annoyingly, when this runs in docker (and perhaps locally in some circumstances), the `npm ci` command produces output that is not json and that breaks.

You will need to add `> /dev/null` at the end of the `npm --prefix e2e ci` command.

I also added `--env NPM_CONFIG_UPDATE_NOTIFIER=false` on the `PLAYWRIGHT_CMD_DOCKER` command after the other `--env` to silence the silly update notification - silly because we cannot update npm in the docker image. This output did not interfere with the json output, but it is annoying to see so nice to be rid of. Can you add these?

camh- (approved):
:green_book:

PR #281, opened 2024-02-20
bytecode: Implement equality and inequality operators

Rossiar:
The equality operators are the only operators that can
operate across all evy types. Therefore it makes sense
to utilise the value interface and its Equals function
to generically handle this expression for all evy types.
The alternative is to create separate equality and
inequality opcodes for each evy type, which would
introduce many new opcodes and repeated code in this
package.


juliaogris (approved):
lgtm - forgot to message earlier.

camh- (approved):

PR #282, opened 2024-02-21
make: Generalise running playwright in docker

camh-:
Change the way playwright is invoked in the Makefile so that specifying
whether to run it in docker is orthogonal to the Makefile target being
invoked.

`make e2e-docker` becomes `make USE_DOCKER=1 e2e`, and `make snaps`
becomes `make snaps USE_DOCKER=1` (note it does not matter where you put
`USE_DOCKER=1` as shown in those two examples).

Now, when a new Makefile target is added for a playwright command, a
second docker version does not need to be added.

Update the CI workflows to use this new method of specifying that
playwright should be run in docker.


juliaogris (approved):
💚 🐳

PR #283, opened 2024-02-22
bytecode: Refactor tests

Rossiar:
Add assumption that x = x at the end of every test,
this cuts down on duplicated assertions and makes
tests easier to read. Add some missing assertions
to TestVMGlobals.

Remove initialization to empty list as all tests
pass with nil rather than the empty slice. Rework
vm tests to not use custom asserts and avoid
linter error on duplicated lines. We can now do all
essential testing in vm_test.go via assert.Equal
and assert.NoError.

### DIFF
--- a/pkg/bytecode/vm_test.go
+++ b/pkg/bytecode/vm_test.go
@@ -8,80 +8,69 @@ import (
 	"evylang.dev/evy/pkg/parser"
 )
 
+// testCase covers both the compiler and the VM.
+type testCase struct {
+	name string
+	// input is an evy program
+	input string
+	// wantStackTop is the expected last popped element of the stack in the vm.
+	wantStackTop value
+	// wantBytecode is the bytecode expected from the compiler.
+	wantBytecode *Bytecode
+}
+
 func TestVMGlobals(t *testing.T) {
 	tests := []testCase{
 		{
-			name: "global assignment",
-			input: `
-			x := 1
-			x = x
-			`,
-			expectedStackTop:  1,
-			expectedConstants: []any{1},
-			expectedInstructions: []Instructions{
-				mustMake(t, OpConstant, 0),
-				mustMake(t, OpSetGlobal, 0),
-				mustMake(t, OpGetGlobal, 0),
-				mustMake(t, OpSetGlobal, 0),
+			name:         "global assignment",
+			input:        "x := 1",
+			wantStackTop: makeValue(t, 1),
+			wantBytecode: &Bytecode{
+				Constants: makeValues(t, 1),
+				Instructions: makeInstructions(
+					mustMake(t, OpConstant, 0),
+					mustMake(t, OpSetGlobal, 0),
+				),
 			},
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			program, err := parser.Parse(tt.input, parser.Builtins{})
-			assert.NoError(t, err, "parser error")
-			comp := NewCompiler()
-			err = comp.Compile(program)
-			assert.NoError(t, err, "compiler error")
-			vm := NewVM(comp.Bytecode())
-			err = vm.Run()
+			bytecode := compileBytecode(t, tt.input)
+			assertBytecode(t, tt.wantBytecode, bytecode)
+			vm := NewVM(bytecode)
+			err := vm.Run()
 			assert.NoError(t, err, "runtime error")
-			stackElem := vm.lastPoppedStackElem()
-			switch expected := tt.expectedStackTop.(type) {
-			case int:
-				assertNumValue(t, float64(expected), stackElem)
-			case float64:
-				assertNumValue(t, expected, stackElem)
-			default:
-				t.Errorf("unexpected object type %v", expected)
-			}
+
+			got := vm.lastPoppedStackElem()
+			assert.Equal(t, tt.wantStackTop, got)
 		})
 	}
 }
 
 func TestUserError(t *testing.T) {
 	tests := []struct {
-		name        string
-		input       string
-		expectedErr error
+		name    string
+		input   string
+		wantErr error
 	}{
 		{
-			name: "divide by zero",
-			input: `
-			x := 2 / 0
-			x = x
-			`,
-			expectedErr: ErrDivideByZero,
-		},
-		{
-			name: "modulo by zero",
-			input: `
-			x := 2 % 0
-			x = x
-			`,
-			expectedErr: ErrDivideByZero,
+			name:    "divide by zero",
+			input:   "x := 2 / 0",
+			wantErr: ErrDivideByZero,
+		}, {
+			name:    "modulo by zero",
+			input:   "x := 2 % 0",
+			wantErr: ErrDivideByZero,
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			program, err := parser.Parse(tt.input, parser.Builtins{})
-			assert.NoError(t, err, "parser error")
-			comp := NewCompiler()
-			err = comp.Compile(program)
-			assert.NoError(t, err, "compiler error")
-			vm := NewVM(comp.Bytecode())
-			err = vm.Run()
-			assert.Equal(t, tt.expectedErr, err)
+			bytecode := compileBytecode(t, tt.input)
+
+			vm := NewVM(bytecode)
+			err := vm.Run()
+			assert.Equal(t, tt.wantErr, err)
 		})
 	}
 }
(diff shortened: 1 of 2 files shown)


camh-:
I've added a `do-not-merge` label because this PR has the commit of the last PR in it and will need to be rebased. It's out of date anyway so needs rebasing for that, but I just want to make sure the other commit gets removed.

pkg/bytecode/vm_test.go
```diff
@@ -11,11 +11,8 @@ import (
...
-			name: "global assignment",
-			input: `
-			x := 1
-			x = x
-			`,
+			name:              "global assignment",
+			input:             "x := 1",
 			expectedStackTop:  1,
 			expectedConstants: []any{1},
 			expectedInstructions: []Instructions{
```
camh-, line 18: This value still has the trailing `GetGlobal`/`SetGlobal` which should be removed here and managed in the test itself.

pkg/bytecode/vm_test.go
```diff
@@ -28,12 +25,16 @@ func TestVMGlobals(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			tt.input += "\nx = x"
```
camh-, line 28: I don't think you should be mutating the test cases. Instead create a local var and append to that. The risk is later the same test cases get used for some other tests and now the test case is wrong.

So just `input := tt.input + "\nx = x"` I think is fine.

This applies to all the tests changed in this way.

pkg/bytecode/vm_test.go
```diff
@@ -28,12 +25,16 @@ func TestVMGlobals(t *testing.T) {
...
 			program, err := parser.Parse(tt.input, parser.Builtins{})
 			assert.NoError(t, err, "parser error")
 			comp := NewCompiler()
 			err = comp.Compile(program)
 			assert.NoError(t, err, "compiler error")
-			vm := NewVM(comp.Bytecode())
+			bytecode := comp.Bytecode()
+			assertInstructions(t, tt.expectedInstructions, bytecode.Instructions)
```
camh-, line 35: This looks like it was missing before and is not part of the refactor. Could you note in the commit message that some missing assertions were added too?

pkg/bytecode/vm_test.go
```diff
@@ -89,94 +85,78 @@ func TestUserError(t *testing.T) {
...
+			tt.input += "\nx = x"
 			program, err := parser.Parse(tt.input, parser.Builtins{})
 			assert.NoError(t, err, "parser error")
 			comp := NewCompiler()
 			err = comp.Compile(program)
 			assert.NoError(t, err, "compiler error")
-			vm := NewVM(comp.Bytecode())
+			bytecode := comp.Bytecode()
+			// add expectation of x = x as the last instructions
+			tt.expectedInstructions = append(tt.expectedInstructions,
```
camh-, line 153: Similar to before - best to keep you test cases constant. In this case I think I'd just subslice the result:

```
bci := bytecode.Instructions
assertInstructions(t, tt.expectedInstructions, bci[:len(bci)-2])
```

The reason I prefer this is if something breaks such that `x=x` generates different opcodes (perhaps the opcode operand is a different global and not number 0), the test wont fail for reasons unrelated to the specific test. Hopefully the test for the assignment will fail so it points at the buggy code, but it should mean all the tests don't start failing. It's not foolproof as the code generated for `x = x` may emit more or less than 2 instructions, but that's a narrower set of failures.

Rossiar:
I've addressed the feedback but now the linter is failing because the test code is duplicating too much. I wrote up a patch to fix this but reading back on @juliaogris [previous comment](https://github.com/evylang/evy/pull/249#discussion_r1464201814) on testing methodology I wanted to discuss before applying it to this PR.

juliaogris:
> [test.patch](https://github.com/evylang/evy/files/14561496/test.patch)
> 
> I've addressed the feedback but now the linter is failing because the test code is duplicating too much. I wrote up a patch to fix this but reading back on @juliaogris [previous comment](https://github.com/evylang/evy/pull/249#discussion_r1464201814) on testing methodology I wanted to discuss before applying it to this PR.

Ok, sorry it took so long to respond.
I've reworked the tests the way I'd approach it:
https://github.com/evylang/evy/commit/1dfb1fa42067dbbfa328c240500dfe45587fdd47?diff=unified&w=1

Feel free to take this as is without committer reference. However, if this doesn't work for
you at all, maybe just go with whatever _you_ want.

Rationale for the changes made:
I still stand by my [previous comment](https://github.com/evylang/evy/pull/249#discussion_r1464201814) though I understand that others might disagree.
What's tested in another package should not require detailed testing here (parser errors), however, what's essential to this package should be tested in a linear way
so it's easy to spot the root cause.

I've tried all kinds of setup and other breakages and tried to make sure that it is still reasonably easy to point to the root cause with the changes made here.
Let me know what you think. @camh- I hope this represents well enough what we discussed, too.

Rossiar:
> I've reworked the tests the way I'd approach it:

This lgtm apart from the `assert.Equal` on the bytecode which is not very readable when instructions are incorrect (which happens often):
```
        vm_test.go:293: want != got
            &bytecode.Bytecode{Constants:[]bytecode.value{1, 2, 3, 4, 5, 6}, Instructions:bytecode.Instructions{0x0, 0x0, 0x0, 0x0, 0x0, 0x1, 0x3, 0x0, 0x0, 0x2, 0x0, 0x0, 0x3, 0x5, 0x0, 0x0, 0x4, 0x6, 0x7, 0x4, 0x2, 0x0, 0x0}}
            &bytecode.Bytecode{Constants:[]bytecode.value{1, 2, 3, 4, 5, 6}, Instructions:bytecode.Instructions{0x0, 0x0, 0x0, 0x0, 0x0, 0x1, 0x3, 0x0, 0x0, 0x2, 0x0, 0x0, 0x3, 0x5, 0x0, 0x0, 0x4, 0x6, 0x0, 0x0, 0x5, 0x7, 0x4, 0x2, 0x0, 0x0}}
```
Adding an extra message to the assert:
```go
assert.Equal(t, tt.wantBytecode, bytecode, "\nwant instructions:\n%s\ngot instructions:\n%s", tt.wantBytecode.Instructions, bytecode.Instructions)
```
creates a more readable error message:
```
vm_test.go:293: want != got
            &bytecode.Bytecode{Constants:[]bytecode.value{1, 2, 3, 4, 5, 6}, Instructions:bytecode.Instructions{0x0, 0x0, 0x0, 0x0, 0x0, 0x1, 0x3, 0x0, 0x0, 0x2, 0x0, 0x0, 0x3, 0x5, 0x0, 0x0, 0x4, 0x6, 0x7, 0x4, 0x2, 0x0, 0x0}}
            &bytecode.Bytecode{Constants:[]bytecode.value{1, 2, 3, 4, 5, 6}, Instructions:bytecode.Instructions{0x0, 0x0, 0x0, 0x0, 0x0, 0x1, 0x3, 0x0, 0x0, 0x2, 0x0, 0x0, 0x3, 0x5, 0x0, 0x0, 0x4, 0x6, 0x0, 0x0, 0x5, 0x7, 0x4, 0x2, 0x0, 0x0}}
            
            want instructions:
            0000 OpConstant 0
            0003 OpConstant 1
            0006 OpAdd
            0007 OpConstant 2
            0010 OpConstant 3
            0013 OpMultiply
            0014 OpConstant 4
            0017 OpDivide
            0018 OpModulo
            0019 OpSubtract
            0020 OpSetGlobal 0
            
            got instructions:
            0000 OpConstant 0
            0003 OpConstant 1
            0006 OpAdd
            0007 OpConstant 2
            0010 OpConstant 3
            0013 OpMultiply
            0014 OpConstant 4
            0017 OpDivide
            0018 OpConstant 5
            0021 OpModulo
            0022 OpSubtract
            0023 OpSetGlobal 0
```

Can you fixup and then we merge your branch into this one and then merge down to main?

juliaogris:
Ok, cool you could take this. I have changed the branch as suggested, but I pulled the 
change into a separate `assertBytecode` - I hope that works for you (made it a bit more readable for me).

BTW, this also explains why you had `wantConstants` and `wantInstructions` separately before,
@camh- and I debated this a fair bit: for the reader who comes to it with less prior
knowledge we feel it is easier to understand what's happening if you pull it together
into the `wantBytecode` field.

🙏  

> ```go
> assert.Equal(t, tt.wantBytecode, bytecode, "\nwant instructions:\n%s\ngot instructions:\n%s", tt.wantBytecode.Instructions, bytecode.Instructions)
> ```

juliaogris (approved):

PR #288, opened 2024-03-07
frontend: Prepare for sidebar in docs

juliaogris:
Prepare for sidebar in docs with various little cleanups.

- add module attribute to wasm_exec.js loading
- refactor sidebar JS in play with newly gained understanding
- remove table of content from output as it will be included in sidebar
- improve styling / CSS for tables, inline code and spacing

We are also trialing
`Commit-group` and `Gen-command` commit trailers here, to separate hand-written
from generated code / HTML files.


camh- (approved):
:green_book: I think the new trailers work fine.

PR #291, opened 2024-03-11
docs: Specify required types array concatenation

juliaogris:
Add note to language specification that only arrays of the same type can be
concatenated.


camh- (approved):
:four_leaf_clover:

Issue #292, opened 2024-03-12
Allow for start index to equal length for slicing


joshcarp:
Encountered this when doing some evy leetcode:
```
print [1 2 3][1:] // [2 3]
print [1 2][1:] // [2]
print [1 2 3][2:2] // []
print [1][1:] // line 5: panic: index out of bounds: 1
```

All other programming languages that I've checked seem to support this and return an empty slice:
```go
package main

import "fmt"

func main() {
	fmt.Println([]int{1, 2, 3}[1:])  // [2 3]
	fmt.Println([]int{1, 2}[1:])     // [2]
	fmt.Println([]int{1, 2, 3}[2:2]) // []
	fmt.Println([]int{1}[1:])        // []
}

```

PR #293, opened 2024-03-12
evaluator: Allow slice start index to be equal to length

joshcarp:
Fixes: #292 

Before:
```
print [1 2 3][1:] // [2 3]
print [1 2][1:] // [2]
print [1 2 3][2:2] // []
print [1][1:] // line 5: panic: index out of bounds: 1
```

After:
```
print [1 2 3][1:] // [2 3]
print [1 2][1:] // [2]
print [1 2 3][2:2] // []
print [1][1:] // []
```

### DIFF
--- a/pkg/evaluator/value.go
+++ b/pkg/evaluator/value.go
@@ -96,7 +96,7 @@ func (s *stringVal) runes() []rune {
 
 func (s *stringVal) Index(idx value) (value, error) {
 	runes := s.runes()
-	i, err := normalizeIndex(idx, len(runes))
+	i, err := normalizeIndex(idx, len(runes), indexExpression)
 	if err != nil {
 		return nil, err
 	}
@@ -206,7 +206,7 @@ func (a *arrayVal) Set(v value) {
 }
 
 func (a *arrayVal) Index(idx value) (value, error) {
-	i, err := normalizeIndex(idx, len(*a.Elements))
+	i, err := normalizeIndex(idx, len(*a.Elements), indexExpression)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


pkg/evaluator/value.go
```diff
@@ -334,22 +334,25 @@ func isBreak(val value) bool {
 func normalizeSliceIndices(start, end value, length int) (int, int, error) {
 	startIdx := 0
 	var err error
-	if start != nil {
```
joshcarp, line 337: from https://go.dev/ref/spec#Types on type assertions:
> asserts that x is not nil and that the value stored in x is of type T.

therefore the nil checks on start and end are redundant

camh-: The nil checks are there to handle when the value is not supplied. We still want to call `normalizeIndex` which does the type assertion as it handles the case of the wrong type for the index, as opposed to a missing index. A single type assertion cannot tell those cases apart.

This evy code produces a result with the code in this PR but should be an error:
```
a := [1 2 3]
b := "x"
print a[b:b]
```

(although the current code generates an `InternalError` for this which is not correct. It seems perhaps what's needed here is for the parser to type check this properly and then perhaps the type assertion approach you're taking here would work. But it still does remove that internal error check (which really should not happen, but that's the case for all the internal errors).

pkg/evaluator/value.go
```diff
@@ -334,22 +334,25 @@ func isBreak(val value) bool {
...
-		}
-	}
-	endIdx := length
-	if end != nil {
-		// length is a valid end slice index, but not a valid ordinary index (out of bounds)
-		if endNum, ok := end.(*numVal); ok && int(endNum.V) != length {
-			endIdx, err = normalizeIndex(end, length)
+	if startNum, ok := start.(*numVal); ok {
+		// length is a valid start slice index, but not a valid ordinary index (out of bounds)
+		if startInt := int(startNum.V); startInt == length {
```
camh-, line 339: At this point, once I see my 4th "start" variable, I start to wonder if there's a better way, and I think there is.

Given that we want to maintain the internal error for the wrong type (notwithstanding fixing the parser), I think we should fix `normalizeIndex` to take a third parameter - `isSlice bool` and change the check in it from:

    if i < -length || i >= length

to

    // A slice expression (isSlice) allows the index to equal the length.
    if i < -length || (isSlice && i > length) || (!isSlice && i >= length)

Then this `normalizeSliceIndicies` becomes a bit more regular (with fewer start variables):

```
        ...
	if start != nil {
		startIdx, err = normalizeIndex(start, length, true /* isSlice */)
		if err != nil {
			return 0, 0, err
		}
	}
	endIdx := length
	if end != nil {
		endIdx, err = normalizeIndex(end, length, true /* isSlice */)
		if err != nil {
			return 0, 0, err
		}
	}
        ...
```

What do you think about that? You identified that both start and end should be able to be equal to the length in a slice expression, and now the code for both is the same, presenting the same symmetry.

camh- (commented):
This PR should also update the spec as now the spec and the code do not match.

pkg/evaluator/value.go
camh-: An alternative to `isSlice bool` is `limit int` where you pass `length` for `limit` for normal indexing and `length + 1` for `limit` for slice expression indexing and use the appropriate value in the various computations in `normalizeIndex`.

camh-: > `i < -length || (isSlice && i > length) || (!isSlice && i >= length)`

Thinking some more, that expression is really too ugly, so either we pass in the limit or calculate it locally in `normalizeIndex()`. I'm not a fan of boolean arguments - they pretty much always need a `/* isSlice */` comment in the argument list to make it readable at the calling site - so I think passing the `limit` makes the most sense.

What do you think?

camh-: Shower thought: I don't like bool args, and using a limit sets the domain of the function `normalizeIndex()` to be larger than what it needs to be (there are really only two limit values we would ever sensibly set), so I think an "enum" might be the better way:

```go
type indexType int

const (
	indexExpression indexType = iota
	sliceExpression
)

func normalizeIndex(idx value, length int, indexType indexType) (int, error) {
	limit := length - 1
	if indexType == sliceExpression {
		limit++ // slice expression indices can index one past the end
	}

	index, ok := idx.(*numVal)
	if !ok {
		return 0, fmt.Errorf("%w: expected num, found %v", ErrType, idx.Type())
	}
	i := int(index.V)
	if i < -length || i > limit {
		return 0, fmt.Errorf("%w: %d", ErrBounds, i)
	}
	if i < 0 {
		return length + i, nil // -1 references len-1 i.e. last element
	}
	return i, nil
}
```

and at the calling sites of `normalizeIndex()`, pass `indexExpression` or `sliceExpression`. That reads much better than the other approaches I mentioned I think.

I'll stop going on about this now. I think I've said all I can (until the next shower).

juliaogris (approved):

Issue #297, opened 2024-03-17
Map key '.' identifier crashes when using evy keywords


joshcarp:
keywords, which are _technically_ valid identifiers aren't allowed to be accessed or set via dot field assignment `map.fieldname`. This leads to this behaviour:
```
m:{}any
m.foobar = true
m["end"] = true
print m // {foobar:true end:true}
m.end = true // line 5: expected map key, found END
```

juliaogris:
Add `lexer.AsIdent(tok token) (token, ok)` to lexer package for this.

juliaogris:
Syntax highlighting will need to be aware of this special case.

PR #299, opened 2024-03-17
parser: Allow keywords as map keys

juliaogris:
Allow keywords as map keys after dot expression and in literal initialization
so that the following now parses without error:

    m := { "for": "me" "end": "you" }
    print m.for m.end

Fixes: https://github.com/evylang/evy/issues/297


camh- (approved):
:four_leaf_clover:

Issue #298, opened 2024-03-17
Panic when comparing two different `any` types


joshcarp:
When comparing different underlying types in an `any` a panic occurs:

```
a:any
a = 1
b: any
b = "hello"
print (a == b)
```

yields:
```
evy run foo.evy
panic: internal error: Num.Equals called with non-Num value

goroutine 1 [running]:
evylang.dev/evy/pkg/evaluator.(*numVal).Equals(0x1400019af48?, {0x105616680?, 0x1400021f500?})
        /home/runner/work/evy/evy/pkg/evaluator/value.go:59 +0x60
evylang.dev/evy/pkg/evaluator.(*anyVal).Equals(0x1400021f1d0?, {0x1056166f0?, 0x140001a7210?})
        /home/runner/work/evy/evy/pkg/evaluator/value.go:147 +0x48
evylang.dev/evy/pkg/evaluator.(*Evaluator).evalBinaryExpr(0x18?, 0x1400018c600)
        /home/runner/work/evy/evy/pkg/evaluator/evaluator.go:633 +0x2ec
evylang.dev/evy/pkg/evaluator.(*Evaluator).eval(0x140001fad58?, {0x105615ba8?, 0x1400018c600})
        /home/runner/work/evy/evy/pkg/evaluator/evaluator.go:223 +0x450
evylang.dev/evy/pkg/evaluator.(*Evaluator).eval(0x140001fadc8?, {0x105615fc8?, 0x1400019af48})
        /home/runner/work/evy/evy/pkg/evaluator/evaluator.go:231 +0x4ec
evylang.dev/evy/pkg/evaluator.(*Evaluator).evalAny(0x10?, 0x1400019af60)
        /home/runner/work/evy/evy/pkg/evaluator/evaluator.go:328 +0x2c
evylang.dev/evy/pkg/evaluator.(*Evaluator).eval(0x1400019af78?, {0x105615e18?, 0x1400019af60})
        /home/runner/work/evy/evy/pkg/evaluator/evaluator.go:199 +0x284
evylang.dev/evy/pkg/evaluator.(*Evaluator).evalExprList(0x1400021f500?, {0x140001a71a0, 0x1, 0x1050b17e4?})
        /home/runner/work/evy/evy/pkg/evaluator/evaluator.go:586 +0x7c
evylang.dev/evy/pkg/evaluator.(*Evaluator).evalFunccall(0x140001d3a40, 0x1400018c5c0)
[...]

PR #300, opened 2024-03-18
evaluator: Fix `any` comparison

juliaogris:
Fix `any` comparison, ensuring that the underlying types are equal before
comparing the values.

Previously when comparing different underlying types in an `any` a panic occured:

    a:any
    a = 1
    b: any
    b = "hello"
    print (a == b)

yielded:

    evy run foo.evy
    panic: internal error: Num.Equals called with non-Num value

Fixes: https://github.com/evylang/evy/issues/298


camh- (approved):
:green_salad:

PR #303, opened 2024-03-18
bytecode: Implement numeric comparison operators

Rossiar:
Adds comparison operators for numic operands, these will have
distinct opcodes from string comparison operands and avoids
verbose type checking code in the vm. Instead, the compiler
will perform type checks on operands to ensure that unique
opcodes are output depending on the types involved.

### DIFF
--- a/pkg/bytecode/code.go
+++ b/pkg/bytecode/code.go
@@ -42,6 +42,18 @@ const (
 	OpEqual
 	// OpNotEqual represents the inequality operator.
 	OpNotEqual
+	// OpNumLessThan represents a less than operator for numeric
+	// operands.
+	OpNumLessThan
+	// OpNumLessThanEqual represents a less than or equal operator for
+	// numeric operands.
+	OpNumLessThanEqual
+	// OpNumGreaterThan represents a greater than operator for numeric
+	// operands.
+	OpNumGreaterThan
+	// OpNumGreaterThanEqual represents a greater than or equal operator
+	// for numeric operands.
+	OpNumGreaterThanEqual
 )
 
 var (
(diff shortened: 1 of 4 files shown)


pkg/bytecode/compiler.go
```diff
@@ -107,6 +111,14 @@ func (c *Compiler) compileBinaryExpression(expr *parser.BinaryExpression) error
 	if err := c.Compile(expr.Right); err != nil {
 		return err
 	}
+	if expr.Left.Type() == parser.NUM_TYPE && expr.Right.Type() == parser.NUM_TYPE {
+		return c.compileNumBinaryExpression(expr)
+	}
```
Rossiar, line 116: This is not necessary for this PR, but I wanted to think ahead about how we would be outputting different opcodes based on type (e.g. OpStringConcat instead of OpAdd when two strings are involved)

pkg/bytecode/vm_test.go
```diff
@@ -165,7 +165,8 @@ func TestVMArithmetic(t *testing.T) {
 					mustMake(t, OpSetGlobal, 0),
 				),
 			},
-		}, {
+		},
```
Rossiar, line 168: I'm not sure how formatting managed to be off from the last change, but I ran `make fmt` in this PR  to resolve.

juliaogris: I think I did that manually - i kinda dislike it when there are so many curly only lines spreading the data out so much and i've seen others do this too, but I'm not sure about this myself. I wonder though why `make fmt` didn't pick this up previously.

pkg/bytecode/compiler.go
```diff
@@ -13,6 +13,10 @@ var (
 	// ErrUnknownOperator is returned when an operator cannot
 	// be resolved.
 	ErrUnknownOperator = fmt.Errorf("%w: unknown operator", ErrInternal)
+	// ErrUnsupportedExpression is returned when an expression is not
+	// supported by the compiler, this indicates an error in the compiler
+	// itself, as all parseable evy expressions should be supported.
+	ErrUnsupportedExpression = fmt.Errorf("%w: unsupported expression", ErrInternal)
```
juliaogris, line 19: I guess in theory it could also mean that the parser has made a mistake - either way and internal error.

pkg/bytecode/code.go
```diff
@@ -42,6 +42,18 @@ const (
 	OpEqual
 	// OpNotEqual represents the inequality operator.
 	OpNotEqual
+	// OpNumLessThan represents a less than operator for numeric
+	// operands.
+	OpNumLessThan
```
juliaogris, line 47: Interesting that you have `OpEqual` for all types but `OpNumLessThan`, when there is also going to be an `OpStringLessThan` I assume? To me it seems either all operators should be bound to their type or none of them, but then I guess we also have `OpAdd` and we will likely have `OpStringConcat` and `OpArrayConcat`... just musing.

juliaogris (approved):
lgtm - just a couple of musings that don't really require any change. Curious to see what @camh- thinks.

pkg/bytecode/code.go
camh-: For the string and array concat, I was expecting them to be different opcodes as they are quite different operations. But comparisons are really the same thing just on different types. But that should not make any difference here - if the VM works better with specific opcodes for specific types, then it should do that. I imagine that equality/inequality was easy as `value` has an `Equals` method so there was no need for different code paths for different types.

camh- (approved):
LGTM

PR #304, opened 2024-03-18
bytecode: Implement string literals and operations

Rossiar:
Add an internal representation for strings to the package and support
assigning strings. Add opcodes for concatenating and comparing strings.
The compiler takes the burden of type checking the arguments so that
the vm can run an optimised hot path.


juliaogris (approved):
aah, so now the `OpStringLessThan` etc come together, yes makes sense to me to have them is seprate operators and not do the typecheck during runtime.

Again, lgtm, curious to see what @camh- will find.

camh- (approved):
LGTM

PR #305, opened 2024-03-19
bytecode: Implement arrays

Rossiar:
Adds support for array literals and array concatenation. Fixed an
issue with assignment where the target and value were swapped, which
would cause the result of any non-trivial assignment to fail. This had
gone unnoticed in the test suite due to the basic (e.g. x = x) usage of
assignments up until this point.

### DIFF
--- a/pkg/bytecode/compiler.go
+++ b/pkg/bytecode/compiler.go
@@ -73,6 +73,15 @@ func (c *Compiler) Compile(node parser.Node) error {
 		if err := c.emit(OpConstant, c.addConstant(num)); err != nil {
 			return err
 		}
+	case *parser.ArrayLiteral:
+		for _, elem := range node.Elements {
+			if err := c.Compile(elem); err != nil {
+				return err
+			}
+		}
+		if err := c.emit(OpArray, len(node.Elements)); err != nil {
+			return err
+		}
 	}
 	return nil
 }
@@ -129,6 +138,9 @@ func (c *Compiler) compileBinaryExpression(expr *parser.BinaryExpression) error 
 	}
 	if expr.Left.Type() == parser.STRING_TYPE && expr.Right.Type() == parser.STRING_TYPE {
 		return c.compileStringBinaryExpression(expr)
+	}
+	if expr.Left.Type().Name == parser.ARRAY && expr.Right.Type().Name == parser.ARRAY && expr.Op == parser.OP_PLUS {
+		return c.emit(OpArrayConcatenate)
 	}
 	return fmt.Errorf("%w: %s with types %s %s", ErrUnsupportedExpression,
 		expr, expr.Left.Type(), expr.Right.Type())
--- a/pkg/bytecode/compiler.go
+++ b/pkg/bytecode/compiler.go
@@ -219,10 +219,10 @@ func (c *Compiler) compileDecl(decl *parser.Decl) error {
 }
 
 func (c *Compiler) compileAssignment(stmt *parser.AssignmentStmt) error {
-	if err := c.Compile(stmt.Target); err != nil {
+	if err := c.Compile(stmt.Value); err != nil {
 		return err
 	}
-	symbol := c.globals.Define(stmt.Value.String())
+	symbol := c.globals.Define(stmt.Target.String())
 	return c.emit(OpSetGlobal, symbol.Index)
 }
 
--- a/pkg/bytecode/value.go
+++ b/pkg/bytecode/value.go
@@ -2,6 +2,7 @@ package bytecode
 
 import (
 	"strconv"
+	"strings"
 
 	"evylang.dev/evy/pkg/parser"
 )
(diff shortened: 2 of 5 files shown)


pkg/bytecode/compiler.go
```diff
@@ -130,6 +143,9 @@ func (c *Compiler) compileBinaryExpression(expr *parser.BinaryExpression) error
 	if expr.Left.Type() == parser.STRING_TYPE && expr.Right.Type() == parser.STRING_TYPE {
 		return c.compileStringBinaryExpression(expr)
 	}
+	if expr.Left.Type().Name == parser.ARRAY && expr.Right.Type().Name == parser.ARRAY && expr.Op == parser.OP_PLUS {
```
Rossiar, line 146: I am unsure if this is the best way to type check arrays, for the compiler it doesn't matter what the sub-type is. It just needs to make sure that the arguments are of type `ARRAY` to know that it should output the correct opcode (and not `OpAdd` or `OpStringConcatenate`)

pkg/bytecode/value.go
```diff
@@ -59,3 +60,37 @@ func (s stringVal) Equals(v value) bool {
...
 	return s == s2
 }
+
+type arrayVal struct {
+	Elements []value
+}
+
+func (a *arrayVal) Type() *parser.Type {
+	return parser.GENERIC_ARRAY
+}
```
Rossiar, line 70: This doesn't feel correct, and in fact we only use Value#Type() in panic error messages (where we could just print the name of the go type). Is this usage acceptable, or would it be better to drop Type entirely and avoid the issue?

pkg/bytecode/vm_test.go
```diff
@@ -479,6 +555,8 @@ func compileBytecode(t *testing.T, input string) *Bytecode {
 func makeValue(t *testing.T, a any) value {
 	t.Helper()
 	switch v := a.(type) {
+	case []any:
```
camh-, line 558: This is ok for now, but you'll probably need to be more rigid in your type handling here. If the value should be `[]num` then you should be accepting `[]int` or `[]float64`. Once you get proper `any` wrapping, you'll treat an input Go type of `[]any` differently to `[]int` for example.

It will also make it catch errors in your test data at compile time rather at test run time.

pkg/bytecode/value.go
juliaogris: FWIW, types were the trickiest bit for me so far, in parser and evaluator. Maybe its because I didn't model them well, maybe they just are hard, but I think Cam is right you will probably not get past tracking types.
Comment sounds fine for now to me.

pkg/bytecode/vm.go
```diff
@@ -210,7 +228,7 @@ func (vm *VM) popBoolVal() boolVal {
 	return val
 }
 
-// popNumVal pops an element from the stack and casts it to a string
+// popStringVal pops an element from the stack and casts it to a string
```
juliaogris, line 231: woops - we all read past this...

juliaogris (commented):
This LGTM now, what do you think @camh- ?

pkg/bytecode/vm_test.go
```diff
@@ -461,6 +461,94 @@ func TestStringExpressions(t *testing.T) {
...
+					mustMake(t, OpSetGlobal, 0),
+				),
+			},
+		},
+		{
+			name: "concatenate preserve original",
+			input: `x := [1 2]
+			y := [3 4]
+			y = x + y
+			x = x`,
```
camh-, line 516: `compileBytecode()` already puts `x = x` on the end. Is this needed?

pkg/bytecode/value.go
```diff
@@ -59,3 +60,38 @@ func (s stringVal) Equals(v value) bool {
...
+type arrayVal struct {
+	Elements []value
+}
+
+func (a arrayVal) Type() *parser.Type {
+	// Revisit this when adding the typeof builtin: https://github.com/evylang/evy/pull/305/files#r1531149977
+	return parser.GENERIC_ARRAY
+}
+
+func (a *arrayVal) String() string {
```
camh-, line 73: Sorry, my previous comment on the pointer receiver applied to all the methods, and passing this around as a pointer to satisfy the interface, when a value type would work just fine too. I don't see the need for this to be used as a pointer at all.

pkg/bytecode/compiler.go
```diff
@@ -207,10 +219,10 @@ func (c *Compiler) compileDecl(decl *parser.Decl) error {
 }
 
 func (c *Compiler) compileAssignment(stmt *parser.AssignmentStmt) error {
-	if err := c.Compile(stmt.Target); err != nil {
```
camh-, line 210: This is an unrelated bug fix is it? If so, please put it in a separate commit so it gets its own commit message rather than it being hidden and undescribed in an unrelated commit. This is useful when looking back on history to understand why a change was made, particularly bug fixes when you find only half a bug was fixed. You see this change in history and it has a completely unrelated description on it ("Implement arrays" ???).

This seems like a pretty fundamental bug. How come no test cases picked this up? Is there a test case added for this?

Or am I misunderstanding this because I'm not sure how assignment even worked if the target and value were backwards.

pkg/bytecode/value.go
```diff
@@ -59,3 +60,38 @@ func (s stringVal) Equals(v value) bool {
 	}
 	return s == s2
 }
+
+type arrayVal struct {
+	Elements []value
+}
+
+func (a arrayVal) Type() *parser.Type {
+	// Revisit this when adding the typeof builtin: https://github.com/evylang/evy/pull/305/files#r1531149977
```
camh-, line 69: What is that link? It only takes me back to this pull request. I assume it is meant to link to the comment, but that should be `https://github.com/evylang/evy/pull/305#discussion_r1531149977` I think, which is what I get when I "Copy Link" on my comment.

pkg/bytecode/compiler.go
Rossiar: Spoke offline, I have split the fix into a separate commit, and then a follow up commit to add the test to cover array concatenation not manipulating the original arrays, which doesn't pass without the fix.

camh- (approved):

PR #306, opened 2024-03-19
bytecode: Implement maps

Rossiar:
Add support for maps to the compiler, the implementation is very
similar to arrays in https://github.com/evylang/evy/pull/305 but works in key/value pairs.

### DIFF
--- a/pkg/bytecode/code.go
+++ b/pkg/bytecode/code.go
@@ -75,6 +75,11 @@ const (
 	// OpArrayConcatenate represents a + operator used to concatenate two
 	// arrays.
 	OpArrayConcatenate
+	// OpMap represents a map literal, the operand N is the length of
+	// the map multiplied by 2. This length N represents the total length
+	// of the flattened map in the stack, where keys and values have been
+	// pushed sequentially (k1, v1, k2, v2...).
+	OpMap
 )
 
 var (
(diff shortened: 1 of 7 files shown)


pkg/bytecode/value.go
```diff
@@ -94,3 +95,34 @@ func (a *arrayVal) Equals(v value) bool {
 	}
 	return true
 }
+
+type mapVal map[string]value
+
+func (m mapVal) Type() *parser.Type {
+	return parser.GENERIC_MAP
```
Rossiar, line 102: Similar to https://github.com/evylang/evy/pull/305#discussion_r1530713436

pkg/bytecode/code.go
```diff
@@ -75,6 +75,11 @@ const (
 	// OpArrayConcatenate represents a + operator used to concatenate two
 	// arrays.
 	OpArrayConcatenate
+	// OpMap represents a map literal, the operand N is the length of
+	// the map multiplied by 2. This length N represents the total length
```
camh-, line 79: Can the operand be defined as the number of pairs instead? Then you could double the size of a map literal. It seems nothing else cares about the value of this operand except where the VM executes the op, and it just increments by 2 each time. Or is there some other general code I've forgotten that would use this value as a size?

camh- (approved):
LGTM.

pkg/bytecode/code.go
Rossiar: Yes it can be defined as the number of pairs, we discussed this between ourselves as Thorsten also outlines in the book:
> It’s equally feasible to use the number of pairs, but then we’d have to double it in the VM to get the number of values sitting on the stack. If we can pre-calculate that in the compiler, why not?

I don't see much merit to his argument, but if we're keeping to the theme of "compiler complicated, vm simple" then I think it follows to pre-calculate in the compiler, I don't see a strong reason to use one over the other tbh.

PR #310, opened 2024-03-26
docsite: Move from preview to docs.evy.dev

juliaogris:
Now that the docsite is reasonably complete, let's move it from
frontend/preview to its proper place in frontend/docs, which means it will
be visible from https://docs.evy.dev. This required a directory move and
manual Makefile update of the `docs` target and a couple of comments.

Add end-to-end tests for the docsite, and also ensure that all external URLs in
.md files return HTTP 200 status with `make test-urls`, which is not include
 in CI to avoid flakiness and should only be occasionally run, e.g. on tools
 update.

---

separate firebase docs deployment at:
https://evy-lang-stage-docs--310-9g2yufjo.web.app/


camh- (approved):
:green_book:

PR #312, opened 2024-03-27
e2e: Attempt to fix flaky tests

juliaogris:
Attempt to fix flaky tests, which is really hard as they are not reproducible
on either local run or PR deployments. The await for `networkidle` timed out
a couple of times on the docs tests, so I've attempted to just remove it. It
was added in the first place because the dialog loads an image
asynchronously, but maybe there is no need for an extra wait here.

We've also had a failure in the playground tests where that I saw in the same
way before I added the delay for the animation which should only take 200ms,
but maybe 300ms as enough extra little wiggle room.

---

just deployed manually against evytest.dev and it ran e2e against it which worked as expected, but who knows....

### DIFF
--- a/e2e/docs/test.js
+++ b/e2e/docs/test.js
@@ -17,14 +17,12 @@ for (const baseURL of baseURLs) {
 
     test("docs about", async ({ page, baseURL }, testInfo) => {
       await page.goto(baseURL)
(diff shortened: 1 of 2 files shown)


camh- (approved):
:test_tube:

PR #314, opened 2024-03-27
docsite: Add syntax coloring

juliaogris:
Add syntax coloring to docsite. This PR required factoring out all syntax
coloring related bits from the playground CSS and module/editor.js.

---
https://evy-lang-stage-docs--314-elz1ivtw.web.app/syntax_by_example.html


camh- (approved):
:green_book:

PR #315, opened 2024-03-28
deploy: Fix per PR deploys

juliaogris:
Fix and simplify deployment which got broken when we introduced evytest.dev.
We now just have a single Makefile target `deploy` that takes the make
variables `CHANNEL` and `ENV`.


camh- (approved):
:deciduous_tree:

PR #316, opened 2024-03-28
docs: Update syntax_by_example.md with evy codeblock

juliaogris:
Update syntax_by_example.md with evy codeblock so that we can use it with evy
syntax coloring in a follow-up commit, which meant we had to update and
revise syntax by example altogether a little bit, to make code snippets
compile-able.

We have also updated some headings, which means all docs including the
sidebar have to be updated in the follow-up generated commit.


camh- (approved):
:green_book:

PR #319, opened 2024-03-28
playsite: Add theme toggle and tweak styles

juliaogris:
Add theme toggle and tweak styles a little. We are moving them together into
the the first commit to avoid multiple e2e snapshot test update commits.


camh- (approved):
:broccoli:

PR #320, opened 2024-03-28
playsite: Add theme.js for preferred theme

juliaogris:
Add theme.js for preferred theme to play.evy.dev so that user doesn't have to
reset preferred theme on each visit. This has already been done on
docs.evy.dev, it was just not moved across to the playsite when themes were
added.


camh- (approved):
:broccoli: :broccoli:

PR #318, opened 2024-03-28
bytecode: Implement index operator

Rossiar:
Adds indexing of maps, arrays and strings. Negative indexing
is supported for arrays and strings. Adds OpSetIndex to
represent an index expression used on the left hand side
of an assignment.

Co-authored-by: joshcarp <joshcarp@users.noreply.github.com>
Co-authored-by: pgmitche <pgmitche@users.noreply.github.com>

PR #321, opened 2024-03-31
bytecode: Implement slice expressions 

Rossiar:
Support positive and negative slices in the compiler, introduce
OpNone to allow passing empty values for the start or end of a
slice to signal the vm to default the value. The vm will return
a new user error named ErrSlice if the start index of a slice
is greater than the end index.

### DIFF
--- a/pkg/bytecode/code.go
+++ b/pkg/bytecode/code.go
@@ -86,6 +86,15 @@ const (
 	// OpSetIndex represents an index operator on the left hand side of
 	// an assignment.
 	OpSetIndex
+	// OpSlice represents a slice operator used on an array or string. The top
+	// three elements are read from the stack and are the end index, start index
+	// and the data structure being operated on. OpNone is used as a default
+	// when the start or end are not specified by the user program.
+	OpSlice
+	// OpNone represents an unspecified value used where values are
+	// optional and unspecified such as the start and end value of a
+	// slice index, or the values in a step range.
+	OpNone
 )
 
 var (
@@ -139,6 +148,8 @@ var definitions = map[Opcode]*OpDefinition{
 	OpMap:      {"OpMap", []int{2}},
 	OpIndex:    {"OpIndex", nil},
 	OpSetIndex: {"OpSetIndex", nil},
+	OpSlice:    {"OpSlice", nil},
+	OpNone:     {"OpNone", nil},
 }
 
 // OpDefinition defines a name and expected operand width for each OpCode.
--- a/pkg/bytecode/compiler.go
+++ b/pkg/bytecode/compiler.go
@@ -51,6 +51,8 @@ func (c *Compiler) Compile(node parser.Node) error {
 		return c.compileAssignment(node)
 	case *parser.BinaryExpression:
 		return c.compileBinaryExpression(node)
+	case *parser.SliceExpression:
+		return c.compileSliceExpression(node)
 	case *parser.UnaryExpression:
 		return c.compileUnaryExpression(node)
 	case *parser.GroupExpression:
@@ -201,6 +203,29 @@ func (c *Compiler) compileStringBinaryExpression(expr *parser.BinaryExpression) 
 	default:
 		return fmt.Errorf("%w %s", ErrUnknownOperator, expr.Op)
 	}
+}
+
+func (c *Compiler) compileSliceExpression(expr *parser.SliceExpression) error {
+	var err error
+	if err = c.Compile(expr.Left); err != nil {
+		return err
+	}
+	if err = c.compileOrEmitNone(expr.Start); err != nil {
+		return err
+	}
+	if err = c.compileOrEmitNone(expr.End); err != nil {
+		return err
+	}
+	return c.emit(OpSlice)
+}
+
+// compilerOrEmitNone will emit OpNone if the provided parser node is
+// nil. If the node is not nil then it will be compiled as normal.
+func (c *Compiler) compileOrEmitNone(node parser.Node) error {
+	if node != nil {
+		return c.Compile(node)
+	}
+	return c.emit(OpNone)
 }
 
 func (c *Compiler) compileUnaryExpression(expr *parser.UnaryExpression) error {
--- a/pkg/bytecode/value.go
+++ b/pkg/bytecode/value.go
@@ -14,6 +14,8 @@ var (
 	// ErrMapKey reports that no value was found for a specific key
 	// used in a map index.
 	ErrMapKey = fmt.Errorf("%w: no value for map key", ErrPanic)
+	// ErrSlice reports an invalid slice where start index > end index.
+	ErrSlice = fmt.Errorf("%w: invalid slice", ErrPanic)
 )
 
 type value interface {
@@ -26,6 +28,13 @@ type indexable interface {
 	// Index returns the value at the specified key. A user error will
 	// be returned if the key is not found inside the structure.
 	Index(key value) (value, error)
+}
+
+type sliceable interface {
+	// Slice returns a subset of elements between the start (inclusive)
+	// and end (exclusive) index. A user error will be returned if start
+	// or end index are outside the bounds of the structure.
+	Slice(start, end value) (value, error)
 }
 
 type numVal float64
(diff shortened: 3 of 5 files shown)


pkg/bytecode/code.go
```diff
@@ -83,6 +83,14 @@ const (
...
 	// string variable.
 	OpIndex
+	// OpSlice represents a slice operator used on an array or string. The top
+	// three elements are read from the stack and are the end index, start index
+	// and the data structure being operated on. OpNull is used as a default
+	// when the start or end are not specified by the user program.
+	OpSlice
+	// OpNull represents a null value. It is used as a sensible default for
+	// the start or end of an OpSlice.
+	OpNull
```
camh-, line 93: I don't think this should be called `OpNull`. Evy does not have null values so I think this can be confusing down the track. I would perhaps call it `OpNoneVal` just like the type and document it as:
> OpNoneVal represents an unspecified value used where values are optional and unspecified such as
> the start and end value of a slice index or the values in a step range.

And then update any reference to "null" values in naming to refer to "none" or similar.

pkg/bytecode/compiler.go
```diff
@@ -259,6 +259,9 @@ func (c *Compiler) compileDecl(decl *parser.Decl) error {
 }
 
 func (c *Compiler) compileAssignment(stmt *parser.AssignmentStmt) error {
+	if err := c.Compile(stmt.Target); err != nil {
```
camh-, line 262: I didn't realise it was a bug to not compile the target of an assignment - I just assumed it was unimplemented because there were no more complex targets than simple variables at the time. So I don't see this as a bug fix but just extending functionality.

Since array and map values are implemented as well as the index operator, I see this change as a separate PR that allows array and map indexing on the left-hand side of an assignment. This slicing PR would follow.

The reason for this split is that 1) this functionality is not really related to slicing, and 2) it really needs more tests to test those forms of assignment. The test added in this commit does not actually test that assigning to an index works - it tests that the assignment does not affect the original value after slicing, but that test can also pass if the index assignment does not work at all. That test can be then added in the slicing PR once index assignment has been merged.

And I don't believe it does work, because the current assignment logic after this commit always creates a new global and that should not happen for array and map index assignment. It also appears that the code in the previous commit does not copy the slice so it would cause mutations in the original slice when modifying a subslice.

pkg/bytecode/value.go
```diff
@@ -134,6 +151,14 @@ func (a arrayVal) Index(idx value) (value, error) {
 	return a.Elements[index], nil
 }
 
+func (a arrayVal) Slice(start, end value) (value, error) {
+	startIdx, endIdx, err := normalizeSliceIndices(start, end, len(a.Elements))
+	if err != nil {
+		return nil, err
+	}
+	return arrayVal{Elements: a.Elements[startIdx:endIdx]}, nil
```
camh-, line 159: This is not copying the slice so mutating the original slice will mutate the slice var and vice versa:

From https://docs.evy.dev/spec.html#index-and-slice : 
> The slice expression `arr[start:end]` copies a substring or subarray [...]

Evaluator implementation: https://github.com/evylang/evy/blob/main/pkg/evaluator/value.go#L232

As commented on the next commit, the test there for testing this mutation isolation does not look like it works due to index assignments not being correctly implemented.

pkg/bytecode/compiler.go
Rossiar: > I see this change as a separate PR that allows array and map indexing on the left-hand side of an assignment

This would feel incomplete to me without being able to handle dot expressions, when I open the new PR is it reasonable to implement those in it as well?

pkg/bytecode/value.go
Rossiar: My bad, I guess the assignment problems obfuscated the real problem for me. I will revisit once the new PR is pushed to support assigning to index expressions.

pkg/bytecode/value.go
```diff
@@ -173,3 +198,45 @@ func (m mapVal) Index(idx value) (value, error) {
...
+	}
+	// end index is exclusive so can be equal to length
+	if endIdx > length || endIdx < -length {
+		return 0, 0, fmt.Errorf("%w: %d", ErrBounds, end)
+	}
+	if endIdx < 0 {
+		endIdx += length
+	}
+	if startIdx > endIdx {
+		return 0, 0, fmt.Errorf("%w: %d > %d", ErrSlice, startIdx, endIdx)
```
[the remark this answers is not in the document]
camh-: Yeah, that's also not good. I just think if you saw an error `Invalid slice: 5 > 3` where start and end are variables and you were expecting those to hold different values (hence invalid slicing), that error message would be strange - what is `5` and what is `3`? I'm not fussed about the format of the error message WRT parens, but there should be something saying that "start is greater than end" in some way so the error is clear.

pkg/bytecode/compiler.go
camh-: I think that should be fine.

Rossiar: @camh- I went back and found that we'd need a new opcode for assigning to an index expression, so I made it part of #318 - I think that dot expressions can follow in another PR

pkg/bytecode/value.go
```diff
@@ -149,6 +166,18 @@ func (a arrayVal) Set(idx, val value) error {
...
 }
 
+func (a arrayVal) Slice(start, end value) (value, error) {
+	startIdx, endIdx, err := normalizeSliceIndices(start, end, len(a.Elements))
+	if err != nil {
+		return nil, err
+	}
+	elements := make([]value, endIdx-startIdx)
+	for i := startIdx; i < endIdx; i++ {
+		elements[i-startIdx] = a.Elements[i]
```
Rossiar, line 176: From my understanding we don't need to use [copyOrRef](https://github.com/evylang/evy/blob/main/pkg/evaluator/value.go#L242) here because the composite types will pass by ref and the primitives in this file are aliased onto go primitives, so will be copied by value automatically.

pkg/bytecode/vm.go
```diff
@@ -160,9 +160,7 @@ func (vm *VM) Run() error {
 				key := vm.popStringVal()
 				m[string(key)] = val
 			}
-			if err := vm.push(m); err != nil {
-				return err
-			}
+			err = vm.push(m)
```
Rossiar, line 163: reducing cyclomatic complexity of `vm.Run()` - this will need a rethink and a fixup PR very soon as it is at 30/30 branches

Rossiar:
@camh- I have rebased onto main and fixed array slicing so it now preserves the original value and passes the test that wrote originally (without the extra erroneous OpGetGlobals) . Please take a look!

pkg/bytecode/vm.go
camh-: Just put in //golint:ignore whatever line. The `Run()` method is special in that it is written to be optimal and will likely need to break the rules to do so.

pkg/bytecode/value.go
camh-: That sounds about right. The evaluator could probably be refactored to do values the same way as you're doing here and then copyOrRef would not be needed there.

pkg/bytecode/value.go
```diff
@@ -189,3 +218,46 @@ func (m mapVal) Index(idx value) (value, error) {
...
+
+func normalizeSliceIndices(start, end value, length int) (int, int, error) {
+	startIdx := 0
+	if _, ok := start.(numVal); ok {
+		startIdx = int(start.(numVal))
+	}
+	endIdx := length
+	if _, ok := end.(numVal); ok {
+		endIdx = int(end.(numVal))
+	}
```
camh-, line 244: nit: I'd move this block of code to after the stuff with startIdx - keep all the startIdx and the endIdx stuff together.

pkg/bytecode/value.go
```diff
@@ -149,6 +166,18 @@ func (a arrayVal) Set(idx, val value) error {
...
 
+func (a arrayVal) Slice(start, end value) (value, error) {
+	startIdx, endIdx, err := normalizeSliceIndices(start, end, len(a.Elements))
+	if err != nil {
+		return nil, err
+	}
+	elements := make([]value, endIdx-startIdx)
+	for i := startIdx; i < endIdx; i++ {
+		elements[i-startIdx] = a.Elements[i]
+	}
```
camh-, line 177: You can use `copy` instead of the loop:
```go
copy(elements, a.Elements[startIdx:endIdx])
```

camh- (approved):

PR #322, opened 2024-04-03
Update all tools with

juliaogris:
Update hermitised tools with tools

    hermit upgrade

Ensure that node, openjre and firebase don't have new major versions.
Upgrade npm packages with ncu.

    (cd .hermit/node/; ncu -u; npm install)

ncu is a node upgrading tool, previously installed globally with `npm
install --global npm-check-updates`.

We are not upgrading playwright because screenshot tests fail with newer
version in un-reproducible ways.

Add manual fixes in follow up commit.


camh- (approved):
:green_salad:

PR #323, opened 2024-04-03
bytecode: Implement if statements

Rossiar:
Add two new opcodes (`OpJumpOnFalse`, `OpJump`) that give the vm the
ability to skip instructions. Add the `ChangeOperand` function to
`Instructions`, which allows the jump operand to be rewritten after
further instructions have been emitted to the bytecode (this process is
known as backfilling). The alternative to backfilling is a two-pass
compiler, which we discarded as too complex for now.

### DIFF
--- a/pkg/bytecode/vm_test.go
+++ b/pkg/bytecode/vm_test.go
@@ -1041,6 +1041,186 @@ func TestErrMapKey(t *testing.T) {
 	assert.Error(t, ErrMapKey, err)
 }
 
+func TestIf(t *testing.T) {
+	tests := []testCase{
+		{
+			name: "single if",
+			input: `x := 1
+			if x == 1
+				x = 2
+			end`,
+			wantStackTop: makeValue(t, 2),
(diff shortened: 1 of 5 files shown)


pkg/bytecode/vm_test.go
```diff
@@ -1041,6 +1041,149 @@ func TestErrMapKey(t *testing.T) {
...
+					mustMake(t, OpEqual),
+					mustMake(t, OpJumpNotTruthy, 63),
+					mustMake(t, OpConstant, 6),
+					mustMake(t, OpSetGlobal, 0),
+					mustMake(t, OpJump, 69),
+					mustMake(t, OpConstant, 7),
+					mustMake(t, OpSetGlobal, 0),
+				),
+			},
+		},
```
camh-, line 1172: perhaps one more test where no block is executed? i.e. no else and no conditions are true.

camh- (approved):

PR #324, opened 2024-04-04
bytecode: Implement while and break statements

Rossiar:
The bytecode for a while statement is similar to that produced for
an `if`, except that the OpJump points to the start of the
statement and not the end. Break statements are also jumps to
the end of their enclosing loop.

### DIFF
--- a/pkg/bytecode/compiler.go
+++ b/pkg/bytecode/compiler.go
@@ -31,6 +31,8 @@ type Compiler struct {
 	constants    []value
 	instructions Instructions
 	globals      *SymbolTable
+	// breaks tracks the positions of break statements in the inner-most loop.
+	breaks []int
 }
 
 // Bytecode represents raw evy bytecode.
@@ -57,10 +59,14 @@ func (c *Compiler) Compile(node parser.Node) error {
 		return c.compileAssignment(node)
 	case *parser.BinaryExpression:
 		return c.compileBinaryExpression(node)
+	case *parser.BreakStmt:
+		return c.compileBreakStatement(node)
 	case *parser.BlockStatement:
 		return c.compileBlockStatement(node)
 	case *parser.IfStmt:
 		return c.compileIfStatement(node)
+	case *parser.WhileStmt:
+		return c.compileWhileStatement(node)
 	case *parser.SliceExpression:
 		return c.compileSliceExpression(node)
 	case *parser.UnaryExpression:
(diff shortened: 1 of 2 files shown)


pkg/bytecode/compiler.go
```diff
@@ -31,6 +31,9 @@ type Compiler struct {
 	constants    []value
 	instructions Instructions
 	globals      *SymbolTable
+	// breakStack tracks the positions of all break statements
+	// in the bytecode
+	breakStack []int
```
Rossiar, line 36: I'm interested if there are any suggestions on how to better handle this. A break can be nested an arbitrary number of nodes deep inside a loop so some sort of compiler-level state is needed to track where to backfill.

pkg/bytecode/compiler.go
```diff
@@ -57,10 +60,14 @@ func (c *Compiler) Compile(node parser.Node) error {
...
 	case *parser.BinaryExpression:
 		return c.compileBinaryExpression(node)
+	case *parser.BreakStmt:
+		return c.compileBreakStatement(node)
 	case *parser.BlockStatement:
 		return c.compileBlockStatement(node)
 	case *parser.IfStmt:
 		return c.compileIfStatement(node)
+	case *parser.WhileStmt:
+		return c.compileWhileStatement(node)
```
Rossiar, line 70: worth mentioning that Thorsten's Monkey language doesn't support loops so we're off-book for this

pkg/bytecode/compiler.go
camh-: I would not do this as a stack, but just a list. At the start of a function that compiles a loop (currently just `compileWhileStatement()`, but I guess the upcoming `compileForStatement()`), you save the current "break list", empty the one in the compiler state and then compile the block that is the statement. After that, range over the break list and patch all the break statements, and then restore the break list you saved at the top.

This is a bit simpler than the current method of having all ancestor loops in the stack and identifying where the nesting ends.

Alternatively, make this stack a stack of slices. Push a new empty slice onto the stack, have `break` add to that list on the top of the stack, and at the end of the loop compile function, pop that slice off and iterate it patching all of the breaks in that list.

You'll need something similar for `continue` - you don't need a slice, just a single continue IP for generating the jump, but you will either need a stack for that too or save it like my original suggestion.

Perhaps combine these into a loop state stack:
```
type loopState struct {
  breakInstructions []int
  continueIP int
}
```

and push a fresh one onto the stack each loop compilation.

Rossiar: Good thinking about the break list, that's much simpler.

> You'll need something similar for continue

The [evy spec](https://github.com/evylang/evy/blob/main/docs/spec.md) doesn't mention `continue` and with a quick test on the website I confirmed that it isn't implemented.

camh-: Ooops. No `continue`. My mistake.

pkg/bytecode/compiler.go
```diff
@@ -286,6 +293,50 @@ func (c *Compiler) compileConditionalBlock(block *parser.ConditionalBlock) (int,
...
+		return err
+	}
+	// Prepare end position of while block, jump to end if
+	jumpOnFalsePos, err := c.emitPos(OpJumpOnFalse, JumpPlaceholder)
+	if err != nil {
+		return err
+	}
+	// take a snapshot of the break list before compiling the body of the
+	// loop
+	outOfScopeBreaks := c.flushBreaks()
```
camh-, line 308: Since flushBreaks does not need to copy (other comment), I think  you can get rid of it and make this:
```
outOfScopeBreaks := c.breaks
c.breaks = []int{}
```

pkg/bytecode/compiler.go
```diff
@@ -382,3 +433,12 @@ func (c *Compiler) compileIndexExpression(expr *parser.IndexExpression) error {
 	}
 	return nil
 }
+
+// flushBreaks will copy and return the list of break positions
+// and reset the value to an empty array.
+func (c *Compiler) flushBreaks() []int {
+	currentBreaks := make([]int, len(c.breaks))
+	copy(currentBreaks, c.breaks)
```
camh-, line 441: You don't need to copy the breaks, as you are replacing it with a fresh slice. The new break list will not share the backing array with the old.

pkg/bytecode/compiler.go
```diff
@@ -31,6 +31,9 @@ type Compiler struct {
 	constants    []value
 	instructions Instructions
 	globals      *SymbolTable
+	// breaks tracks the positions of all break statements
```
camh-, line 34: This comment needs to be updated - it just tracks the inner-most loop breaks.

pkg/bytecode/vm_test.go
```diff
@@ -1221,6 +1221,164 @@ func TestIf(t *testing.T) {
...
+			},
+		},
+		{
+			name: "break loop",
+			input: `x := 0
+			while x < 5
+				x = x + 1
+				if x == 3 
+                    break
+                end
```
camh-, line 1287: odd formatting. tabs vs spaces?

camh- (approved):

PR #327, opened 2024-04-10
Let fmt package call String() on arguments

camh-:
Remove the call to the `String()` method on arguments to the printf
family of functions where those functions would call `String()` anyway,
according to the rules documented in the overview of the `fmt` package.

This makes for less verbose and slightly cleaner code.


juliaogris (approved):
Better, I agree! 🟢

PR #328, opened 2024-04-11
docs: Change join builtin description for []any

camh-:
Update the builtin docs for `join` to say it takes an array of `any`
instead of an array of `string`, as this is what the test case expects.
It makes `join` a bit more useful with less boilerplate to make strings
of non-strings.

Closes: https://github.com/evylang/todo/issues/94


juliaogris (commented):
lgtm 💚

juliaogris (approved):
💚

PR #329, opened 2024-04-11
Define and implement an array repetition operator

camh-:
Define the `array * num` operator to repeat `array` `num` number of
times. This allows for easy initialisation of variable sized arrays
without needing to use a loop to do so.

Update the spec with a definition of the operator. Update
`syntax_by_example.md` to give some basic examples.

Implement this in the evaluator and the bytecode compiler/vm.

Issue: https://github.com/evylang/todo/issues/101

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -33,7 +33,7 @@ To get an intuitive understanding of Evy, you can either look at its
 12. [**Maps**](#maps)
 13. [**Index and Slice**](#index-and-slice)
 14. [**Operators and Expressions**](#operators-and-expressions)  
-    [Arithmetic and Concatenation Operators](#arithmetic-and-concatenation-operators), [Logical Operators](#logical-operators), [Comparison Operators](#comparison-operators), [Unary Operators](#unary-operators)
+    [Arithmetic, Concatenation and Repetition Operators](#arithmetic,-concatenation-and-repetition-operators), [Logical Operators](#logical-operators), [Comparison Operators](#comparison-operators), [Unary Operators](#unary-operators)
 15. [**Precedence**](#precedence)
 16. [**Statements**](#statements)
 17. [**Whitespace**](#whitespace)  
@@ -588,6 +588,13 @@ If you try to concatenate two literals of different types such as
 `arr := [1 2] + ["a" "b"]`, you will get a parse error. Use
 `arr := [1 2 "a" "b"]` instead.
 
+Arrays can be repeated with the `*` operator, for example `[0] * 5`
+results in `[0 0 0 0 0]`. `["hello" "world"] * 2` results in `["hello"
+"world" "hello" "world"]`. An array repeated zero times results in an
+empty array of the same type as the array. The number on the right hand
+side of the `*` operator must be an non-negative integer value. Using a
+non-integer or negative value will result in an error.
+
 The elements of an array can be accessed via index starting at `0`. In
 the example `arr := ["abc" 123]` the first element in the array
 `arr[0]` is `"abc"`.
@@ -787,21 +794,23 @@ no automated type conversion of operands.
 There are a variety of binary operators: arithmetic, concatenation,
 logical, and comparison operators.
 
-| Operator            | Operands  | Result   | Description   |
-| ------------------- | --------- | -------- | ------------- |
-| `+` `-` `*` `/` `%` | `num`     | `num`    | arithmetic    |
-| `+`                 | `string`  | `string` | concatenation |
-| `+`                 | array     | array    | concatenation |
-| `and` `or`          | `bool`    | `bool`   | logical       |
-| `<` `<=` `>` `>=`   | `num`     | `bool`   | comparison    |
-| `<` `<=` `>` `>=`   | `string`  | `bool`   | comparison    |
-| `==` `!=`           | all types | `bool`   | comparison    |
+| Operator            | Operands        | Result   | Description   |
+| ------------------- | --------------- | -------- | ------------- |
+| `+` `-` `*` `/` `%` | `num`           | `num`    | arithmetic    |
+| `+`                 | `string`        | `string` | concatenation |
+| `+`                 | array           | array    | concatenation |
+| `*`                 | array `*` `num` | array    | repetition    |
+| `and` `or`          | `bool`          | `bool`   | logical       |
+| `<` `<=` `>` `>=`   | `num`           | `bool`   | comparison    |
+| `<` `<=` `>` `>=`   | `string`        | `bool`   | comparison    |
+| `==` `!=`           | all types       | `bool`   | comparison    |
 
-### Arithmetic and Concatenation Operators
+### Arithmetic, Concatenation and Repetition Operators
 
 The **arithmetic operators** `+`, `-`, `*`, `/`, and `%` stand for addition,
 subtraction, multiplication, division, and the [modulo operator]. The
-symbol `+` can also be used to concatenate strings and arrays.
+symbol `+` can also be used to concatenate strings and arrays. The `*` symbol
+can be used to repeat an array a number of times.
 
 The **modulo operator** `%`, also known as the remainder operator, returns the
 remainder of a division operation. For example, `10 % 3` results in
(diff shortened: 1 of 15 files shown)


docs/spec.md
```diff
@@ -588,6 +588,13 @@ If you try to concatenate two literals of different types such as
 `arr := [1 2] + ["a" "b"]`, you will get a parse error. Use
 `arr := [1 2 "a" "b"]` instead.
 
+Arrays can be repeated with the `*` operator, for example `[0] * 5`
+results in `[0 0 0 0 0]`. `["hello" "world"] * 2` results in `["hello"
+"world" "hello "world"]`. An array repeated zero times results in an
```
juliaogris, line 593: quote missing after second `"hello`

docs/spec.md
```diff
@@ -588,6 +588,13 @@ If you try to concatenate two literals of different types such as
...
+empty array of the same type as the array. The number on the right hand
+side of the `*` operator must be an integer value. Using a decimal value
+will result in an error.
```
juliaogris, line 596: how about negative integers?

pkg/bytecode/vm.go
```diff
@@ -20,6 +20,9 @@ var (
 	// produce an invalid result. In Golang, floating point division
 	// by zero produces +Inf, and modulo by zero produces NaN.
 	ErrDivideByZero = fmt.Errorf("%w: division by zero", ErrPanic)
+	// ErrBadRepetition is returned when the right-hand side of the array
+	// repetition operator is not an integral value.
+	ErrBadRepetition = fmt.Errorf("%w: repetition count not a while number", ErrPanic)
```
camh-, line 25: oops. w/while/whole/

pkg/bytecode/vm.go
```diff
@@ -152,6 +155,18 @@ func (vm *VM) Run() error {
...
 			concatenated.Elements = append(concatenated.Elements, right.Elements...)
 			err = vm.push(concatenated)
+		case OpArrayRepeat:
+			right := vm.popNumVal()
+			left := vm.popArrayVal()
+			repetitions := int(right)
+			if float64(repetitions) != float64(right) {
+				return fmt.Errorf("%w: %s", ErrBadRepetition, right)
+			}
+			elements := make([]value, 0, repetitions)
```
camh-, line 165: s/repetitions/len(left.Elements) * repetitions/

pkg/bytecode/vm.go
```diff
@@ -20,6 +20,9 @@ var (
 	// produce an invalid result. In Golang, floating point division
 	// by zero produces +Inf, and modulo by zero produces NaN.
 	ErrDivideByZero = fmt.Errorf("%w: division by zero", ErrPanic)
+	// ErrBadRepetition is returned when the right-hand side of the array
+	// repetition operator is not an integral value.
```
juliaogris, line 24: s/integral value/integer value/

juliaogris (approved):
awesome, thank you so much. a couple of doc nits, otherwise really clear.

camh-:
All fixed and pushed. PTAL

pkg/evaluator/evaluator.go
```diff
@@ -718,14 +719,29 @@ func evalBinaryBoolExpr(op parser.Operator, left, right *boolVal) (value, error)
...
+		rightArray := right.(*arrayVal)
+		result := left.Copy()
+		rightElements := *rightArray.Copy().Elements
+		*result.Elements = append(*result.Elements, rightElements...)
+		return result, nil
+	case parser.OP_ASTERISK:
+		rightNum := right.(*numVal)
+		repetitions := int(rightNum.V)
+		if float64(repetitions) != float64(rightNum.V) {
+			return nil, fmt.Errorf("%w: %v", ErrBadRepetition, right)
```
juliaogris, line 734: Maybe this should also include an error case for negative integers? (and maybe deserves a test?)

docs/spec.md
```diff
@@ -588,6 +588,13 @@ If you try to concatenate two literals of different types such as
 `arr := [1 2] + ["a" "b"]`, you will get a parse error. Use
 `arr := [1 2 "a" "b"]` instead.
 
+Arrays can be repeated with the `*` operator, for example `[0] * 5`
+results in `[0 0 0 0 0]`. `["hello" "world"] * 2` results in `["hello"
+"world" "hello" "world"]`. An array repeated zero times results in an
+empty array of the same type as the array. The number on the right hand
+side of the `*` operator must be an non-negative integer value. Using a
+decimal value will result in an error.
```
juliaogris, line 596: Using a decimal or non-integer value will result in an error (see below)?

juliaogris (approved):
looks still great to me. just a question around negative repetition counts.

pkg/evaluator/evaluator.go
camh-: yes it should. good catch. a negative value will crash things I expect

docs/spec.md
camh-: what's below? I will change to "non-integer or negative"

pkg/evaluator/evaluator.go
camh-: done for both evaluator and vm

docs/spec.md
camh-: done. I guess "see below" refers to the comment above? Let me know if you think that reads ok now.

pkg/bytecode/compiler.go
```diff
@@ -183,6 +183,9 @@ func (c *Compiler) compileBinaryExpression(expr *parser.BinaryExpression) error
 	if expr.Left.Type().Name == parser.ARRAY && expr.Right.Type().Name == parser.ARRAY && expr.Op == parser.OP_PLUS {
 		return c.emit(OpArrayConcatenate)
 	}
+	if expr.Left.Type().Name == parser.ARRAY && expr.Right.Type() == parser.NUM_TYPE {
```
Rossiar, line 186:
```suggestion
	if expr.Left.Type().Name == parser.ARRAY && expr.Right.Type() == parser.NUM_TYPE && expr.Op == parser.OP_ASTERISK {
```
being specific about the op doesn't hurt

pkg/bytecode/vm.go
```diff
@@ -20,6 +20,9 @@ var (
 	// produce an invalid result. In Golang, floating point division
 	// by zero produces +Inf, and modulo by zero produces NaN.
 	ErrDivideByZero = fmt.Errorf("%w: division by zero", ErrPanic)
+	// ErrBadRepetition is returned when the right-hand side of the array
+	// repetition operator is not an non-negative integer value.
```
Rossiar, line 24:
```suggestion
	// repetition operator is not a positive integer value.
```
double negatives are hard to read

pkg/bytecode/vm.go
```diff
@@ -20,6 +20,9 @@ var (
...
+	ErrBadRepetition = fmt.Errorf("%w: bad repetition count", ErrPanic)
```
Rossiar, line 25: this error message is a bit vague, perhaps something like this?
```suggestion
	ErrBadRepetition = fmt.Errorf("%w: array repetition must be a positive integer", ErrPanic)
```

Rossiar (commented):
some small things, overall lgtm

pkg/bytecode/vm.go
camh-: Unfortunately the term "positive" excludes zero. "non-negative integer" is the mathematical term to refer to  the integer range 0.. : https://en.wikipedia.org/wiki/Natural_number. I'll try to reword it a different way.

juliaogris: yup, agree, positive is not correct because it excludes zero. The go stdlib uses that term too, e.g.: https://pkg.go.dev/math/rand#Int

pkg/bytecode/vm.go
juliaogris: i was also thinking of alternatives for the word count here, but couldn't come up with anything. I also like `array repetition must be a non-negative integer`

camh-: I initially had it more verbose when I had forgotten to check for negative numbers, but simplified it when it got too long. I'll keep this base error the same, but add details where the error is returned.

docs/spec.md
juliaogris: yeah, sorry, the below was below when I made the comment.... you read it correctly.
🍍

juliaogris: yeah, sorry, the below was below when I made the comment.... you read it correctly.
🍍

juliaogris (approved):
still 💚

docs/spec.md
camh-: Understood. See above.

pkg/bytecode/compiler.go
camh-: Thanks. This is arguably a bug. Fixed.

camh-:
do not merge label because I need to squash in the fixup commits before merging. I've forgotten to do that before, so this may help.

Rossiar (approved):
lgtm

Issue #317, opened 2024-03-28
Disallow array and string index with decimal places


juliaogris:
```
x := "abc"
print "x:" x[1.62] // x: b
```

This should cause a runtime error.

More specifically, the index must be an integer, whether a constant, variable or expression. In Go, this would be `float64(int(i)) == i` where `i` is the index.

Currently, the value is truncated via conversion to `int`: https://github.com/evylang/evy/blob/main/pkg/evaluator/value.go#L374

PR #331, opened 2024-04-13
eval: Error on non-integer index and slice expressions

juliaogris:
Error on non-integer index and slice expressions as it doesn't make much sense
to use it, e.g. in arr[1.1]. We are reusing the nice error message wording
that was introduced in with the ErrBadRepetion. We have also copied the way
we do normalizeSlices indices from evaluator, basing it on normalizeIndex.

Fixes https://github.com/evylang/evy/issues/317


camh- (approved):
:green_apple:

PR #332, opened 2024-04-22
bytecode: Implement for loops

Rossiar:
Add OpStepRange and OpIterRange which push the result of evaluating 
the condition of the loop onto the stack to be evaluated by OpJumpOnFalse. 
These operations update the iteration state on the stack and also push the
current value if there is a loop variable so the bytecode can assign that value
to the variable.

Co-authored-by: Cam Hutchison <camh@xdna.net>

### DIFF
--- a/pkg/bytecode/compiler.go
+++ b/pkg/bytecode/compiler.go
@@ -63,6 +63,8 @@ func (c *Compiler) Compile(node parser.Node) error {
 		return c.compileBreakStatement(node)
 	case *parser.BlockStatement:
 		return c.compileBlockStatement(node)
+	case *parser.ForStmt:
+		return c.compileForStatement(node)
 	case *parser.IfStmt:
 		return c.compileIfStatement(node)
 	case *parser.WhileStmt:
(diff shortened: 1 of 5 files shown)


pkg/bytecode/value.go
```diff
@@ -197,26 +200,33 @@ func (m mapVal) Equals(v value) bool {
...
+		val, ok := m.m[idx]
+		if !ok {
+			return nil, fmt.Errorf("%w %q", ErrMapKey, idx)
+		}
+		return val, nil
+	case numVal:
+		// special case where ranging over a map uses a numVal to
+		// access the order
+		key := m.order[int(idx)]
+		return key, nil
```
Rossiar, line 227: We are a bit leery of this, but didn't see a better way to implement this - suggestions welcome!

pkg/bytecode/compiler.go
```diff
@@ -241,6 +243,175 @@ func (c *Compiler) compileBlockStatement(block *parser.BlockStatement) error {
 	return nil
 }
 
+func (c *Compiler) compileForStatement(stmt *parser.ForStmt) error {
```
Rossiar, line 246: I tried quite hard to unify compileIterRange and compilerStepRange, but the two functions are just different enough that any code sharing begins to look really awkward.

camh-:
I have pushed my changes that rework this PR that addresses a few issues raised on discord:
1) It no longer re-evaluates the expressions used in the loop, as per the spec.
2) It does not create a var with an empty name when there is no loop variable (allowing nested loops with the current scopeles implementation).
3) Does the loop condition computation and comparison as part of the operations which should be faster than doing it in bytecode.
4) Ranges over strings.

This also unifies the two implementations (iter range and step range) in the compiler and also removes the map indexing by integer.

I've kept the commit that adds the `Order` field to the map value.

pkg/bytecode/compiler.go
camh-: I managed to unify these largely (with a separate preamble for the two range types) by pushing a lot of the loop comparison logic into the operations.

Rossiar (commented):
I can't approve as this is my PR - but lgtm

@juliaogris can you unblock us with a tick?

juliaogris (approved):
💚 - looks great to me, thanks @camh- for taking the time to give me personal walk through as well!

juliaogris (approved):
💚

PR #334, opened 2024-04-27
parser/eval: Move StepRange defaults to parser

camh-:
Add methods to `StepRange` to return the start, stop and step values,
returning the default value (as a `NumLiteral`) if the value wanted is a
default value (0 for start and 1 for step). Update the evaluator to use
these methods instead of deciding the defaults for itself.

This puts the defaults in one place for when there are multiple
consumers of the AST; the evaluator and the bytecode compiler.


juliaogris (approved):
right, definitely better, i like it.
🥦

PR #335, opened 2024-04-28
bytecode: Remove bytecode test assertions

Rossiar:
These assertions are proving laborious to write as the tests in
this package grow more mature and complex. The team has agreed
that since they had add little value for the effort, it is safe
to remove them.


juliaogris (approved):
lgtm - I agree, a bit too cumbersome at this stage for testing.

PR #337, opened 2024-05-03
make: Rework full and slim builds

juliaogris:
Move the build default slim for faster builds with `go build` and also to re-enable

	go install evylang.dev/evy@latest

Embeds related to evy.wasm and full build dependency have been moved to
main-full.go which is included via the `-tags=full` flag. Update Makefile and
goreleaser config.


camh- (approved):
:broccoli:

Issue #338, opened 2024-05-09
Evy fmt converts string key map initialisations to string type


joshcarp:
Before formatting:
<img width="251" alt="image" src="https://github.com/evylang/evy/assets/32605850/e2402d5d-e131-4715-a3ce-947e9232e96d">

After formatting:

<img width="251" alt="image" src="https://github.com/evylang/evy/assets/32605850/3624751d-94ee-4a70-8431-c51bd00a02d8">

joshcarp:
https://play.evy.dev/#content=H4sIAAAAAAAAE0tUsLJVqC4uKcrMS7cqKSpNreXiKijKzCtR0EjU5AIAzCy3Kx4AAAA=

juliaogris:
```evy
a := { "a" : true }
print a
```

should be a syntax error. the correct version would be

```evy
a := { a : true }
print a
```

PR #339, opened 2024-05-09
parser: Report error for quoted map keys

juliaogris:
Report parse error for quoted map keys in map literals.
Before the following code did NOT report an error:

	m := {"a": true}
	print m

Now it does. The issue was that we accidentally printed `string` for string
literals in the parser which doubled up with the `string` keyword and made
AsIdent() work incorrectly for string literals which it now just passes
through.

Fixes: https://github.com/evylang/evy/issues/338

### DIFF
--- a/pkg/parser/parser_test.go
+++ b/pkg/parser/parser_test.go
@@ -1473,11 +1473,9 @@ end
 
 func TestMapLitErr(t *testing.T) {
 	inputs := map[string]string{
-		`
-func fnm:{}any
-    return {a:1{b:2}}
-end
-`: `line 3 column 16: expected map key, found "{"`,
+		`print {a:1{b:2}}`:    `line 1 column 11: expected map key, found "{"`,
+		`print {"x": true}`:   `line 1 column 8: expected map key, found "x"`,
+		`print {"end": true}`: `line 1 column 8: expected map key, found "end"`,
 	}
 	for input, wantErr := range inputs {
 		parser := newParser(input, testBuiltins())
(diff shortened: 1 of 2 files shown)


pkg/parser/parser_test.go
```diff
@@ -1478,6 +1478,12 @@ func fnm:{}any
     return {a:1{b:2}}
 end
 `: `line 3 column 16: expected map key, found "{"`,
+		`a := {"x": true}
+print a
+`: `line 1 column 7: expected map key, found "x"`,
+		`a := {"end": true}
+print a
+`: `line 1 column 7: expected map key, found "end"`,
```
camh-, line 1486: These tests all seem needlessly verbose for what they are testing. The following would seem to work just as well and are otherwise valid evy program (you could just use `a :=` instead of `print` since you are testing a parse error, so there is no need for `a` to actually be used with a followup `print`, but you also don't need a var anyway and can just have a valid program apart from the map literal):
```
        inputs := map[string]string{
                `print {a:1{b:2}}`:    `line 1 column 11: expected map key, found "{"`,
                `print {"x": true}`:   `line 1 column 8: expected map key, found "x"`,
                `print {"end": true}`: `line 1 column 8: expected map key, found "end"`,
        }
 ```

camh- (approved):
:frog: LGTM, although I think the test programs could be a bit simpler - I always have trouble seeing the boundaries and structure of these multiline programs in the mdidle of maps.

pkg/parser/parser_test.go
juliaogris: yeah, way nicer. Thanks for looking so carefully. fixed.

PR #341, opened 2024-05-10
Add evy test command

joshcarp:
Opening as draft PR incase there's any interest. I'm not looking for feedback on this PR because it's just an instrumental feature that will speed up some of the work i'm doing at the moment

- `evy test` will run any functions that start with `test` 
- adds builtin `assert` and `assertEqual` functions
- status of the asserts will be printed out on completion of the test function
- Changes the error print out to include the filename so that IDEs can link to the failing test

Motivation: I'm doing a lot of translation from other programming languages at the moment and being able to quickly test without needing to add custom `assert` functions every time is nice. 

```
$ evy test test/test.evy                         
4 of 4 asserts passed
```

```
$ evy test test/test.evy
assert failed: [() (()) (()()) ] != [() (()) (()())]
test/test.evy:24:1: 1 of 4 asserts FAILED
```
image

### DIFF
--- a/main.go
+++ b/main.go
@@ -81,6 +81,7 @@ evy is a tool for managing evy source code.
 type app struct {
 	Version kong.VersionFlag `short:"V" help:"Print version information"`
 	Run     runCmd           `cmd:"" help:"Run Evy program."`
+	Test    testCmd          `cmd:"" help:"Test Evy program."`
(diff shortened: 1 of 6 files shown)


juliaogris:
I'd only use `assert`.
With one parameter we expect a bool value to be true.
With two values we assert equality.

Also:
mayb worthwhile for the CLI:
- `--fail-fast`
- `--json`

PR #346, opened 2024-05-17
all: Upgrade hermitised tools and go packages

juliaogris:
Upgrade tooling and dependencies with

    hermit upgrade
    go get -u
    go mod tidy

We don't update node tools here because they need more care manual update to
the e2e test.


camh- (approved):
:green_circle:

PR #347, opened 2024-05-18
examples: Add top level Evy source examples directory

juliaogris:
Add top level Evy source examples directory linking to the frontend examples
and the human-eval training set in Evy.


camh- (approved):
:green_apple:  Fantastic. I reviewed every line in detail! LGTM

PR #348, opened 2024-05-18
node: Upgrade node tooling

juliaogris:
Upgrade node tooling

    (cd .hermit/node/; ncu -u; npm install)
    (cd e2e; ncu -u; npm install)

Manually upgrade playwright docker image use in Makefile to v1.44.0 align with
CLI tooling.

Several new rendering issues showed in the new version, which could eventually
be traced down to an incorrect CSS main wrapper height value of 10% combined
with display flex. The available space is 100% - topnav-height. After this
CSS fix all tests started passing again without updates. Phew.

### DIFF
--- a/.hermit/node/package-lock.json
+++ b/.hermit/node/package-lock.json
@@ -6,7 +6,7 @@
     "": {
       "devDependencies": {
         "prettier": "^3.2.5",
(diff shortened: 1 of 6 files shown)


camh- (approved):
:green_salad:

PR #350, opened 2024-05-19
docs: Fix max reference

juliaogris:
Fix `max` reference example which had a copy-paste error showing the use of
`min`.


camh- (approved):
:green_book:

PR #351, opened 2024-05-20
builtins: Add `abs` function

juliaogris:
Add `abs` function for the absolute value of a number. When converting the
human-eval samples I had to manually implement this function one too many
times.


camh- (approved):
:four_leaf_clover:

PR #352, opened 2024-05-20
parser: Improve error message for variadic

juliaogris:
Improve error message for invalid variadic parameters, where a non-variadic
and a variadic parameters are combined. Previously it said

	invalid variadic parameter, must be used with single type

now it says:

	variadic parameter cannot be used with other parameters


camh- (approved):
:herb:

PR #353, opened 2024-05-20
builtins: Add `hsl` function

juliaogris:
Add the function `hsl` that sets the drawing pen color to the given `hue`,
`saturation`, `lightness` and `alpha` values.

It is useful for mathematical manipulation of colors such as creating trails
in animations, gradients and finding complimentary colors.


camh- (approved):
:rainbow: LGTM

PR #357, opened 2024-05-22
evaluator: Deep copy array repetitions

juliaogris:
Deep copy array repetitions so that the repetition operator can be used to
initialize 2 or larger dimensional arrays. Previously a reference to inner
arrays was copied which did not allow for array elements to be updated
independently.

Co-authored-by: Cam Hutchison <camh@xdna.net>


camh- (approved):
:green_heart:

PR #358, opened 2024-05-22
levy: Add --with-marked flag

juliaogris:
Add --with-marked flag in levy tool that exports HTML files with already
marked or checked answers. This is another useful debug output while
creating Evy course materials. Add all necessary wiring to the
question.Model and new golden tests.

In the run-up to the actual change we fixed a couple of minor issues in the
HTML output (value instead of id for checkbox and radio button) and added a
difficulty class to the question form, so we can style questions according to
difficulty at least in debug output.

Also restructure golden files into sub-directories as they were getting a
little busy.


learn/pkg/question/question.go
```diff
@@ -189,32 +191,43 @@ func (m *Model) PrintHTML(buf *bytes.Buffer) {
...
 }
 
-func (m *Model) printAnswerChoicesHTML(list *markdown.List, buf *bytes.Buffer) {
+func (m *Model) printAnswerChoicesHTML(list *markdown.List, buf *bytes.Buffer, withMarked bool) error {
 	buf.WriteString("<fieldset>\n")
 	inputType := "radio"
 	if m.Frontmatter.AnswerType == "multiple-choice" {
 		inputType = "checkbox"
 	}
+	checkedByIndex := map[int]string{}
```
camh-, line 218: Rather that building up a map of just indices to a constant string, perhaps consider getting the answer and asking it if if the index is a correct answer:

```
        var answer Answer
        if withMarked {
                var err error
                answer, err = m.Frontmatter.getAnswer(m.privateKey)
                if err != nil {
                        return err
                }
        }
        for i, item := range list.Items {
                checked := ""
                if answer.correctAnswerIndices()[i] {
                        checked = "checked" // note: would put space at end here. this passes current tests though
                }
 ```

Just a suggestion. I spent a little time understanding this `checkedByIndex` map so my thoughts went down this path. Up to you.

Also, this works because the zero value `Answer` has no correct answers so if `answer` is not set, it will never result in a checked input.

Comment (camh-) [reply learn/pkg/question/question.go:218]: Thoughts after :shower: 

That `answer.correctAnswerIndices()[i]` each time in the loop is a little silly - it rebuilds that map each time. So perhaps instead of storing the `answer`, store the `map[int]bool` from `correctAnswerIndices()` and look that up instead of what you currently have of generating a new map?

**Comment by camh-** [comment]
I've pushed `camh/learn-with-marked` which has a commit on top of your branch showing what I meant above. Feel free to ignore or to fold into your commit if you want.

Also, not sure if you fixed it locally and have not pushed, but the second commit still says "lean" not "learn" as the title prefix.

**Review by camh- — APPROVED** [review]
:heavy_check_mark:

File: learn/pkg/question/question.go (hunk 1 above)
Comment (juliaogris) [reply learn/pkg/question/question.go]: ok, i have squashed you change.
Fun fact: I had nearly exactly what you're suggesting at first and removed it feeling there is too much going on in the single function.
thanks for taking the time and posting the commit.

File: learn/pkg/question/question.go [hunk 2]
```diff
@@ -225,6 +238,19 @@ func (m *Model) printAnswerChoicesHTML(list *markdown.List, buf *bytes.Buffer) {
...
+}
+
+func (m *Model) checkedByIndex() (map[int]string, error) {
+	answer, err := m.Frontmatter.getAnswer(m.privateKey)
+	if err != nil {
+		return nil, err
+	}
+	checkedByIndex := map[int]string{}
+	for idx := range answer.correctAnswerIndices() {
+		checkedByIndex[idx] = "checked"
```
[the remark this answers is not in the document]
juliaogris: done.

PR #359, opened 2024-05-22
playsite: Fix hsl usage in events sample

juliaogris:
Fix hsl usage in events sample to be composed with the color function.
We have updated hsl to return a string in commit 9b2c83c0 but forgot
to update the example. Let's do it!

While at it, remove all hsl strings in other samples.

---

Compare: https://play.evy.dev/#events
To: https://evy-lang-stage-play--359-whn7zd5c.web.app/#events

use arrow keys to change the color of the dot.


camh- (approved):
:green_circle:

PR #360, opened 2024-05-23
playsite: Revisit the splash of trig example

juliaogris:
Revisit the splash of trig example using the new `hsl` builtin and repetition
operator to initialize array. We've reworked the example so that the dots are
now stored in a two dimensional array: an array of rings, where each ring
array holds the dot within the ring. Arguably this is a little easier to
reason about. Additionally, we calculate by default the radius from the orbit
count, and the orbit value. We've also made the animation fully fit within
the canvas.

This changes should make it easier for people playing with it to:

- Add extra dots by adding an extra color
- Add extra orbits by adding an extra speed

I'm not 100% convinced that his is really making things easier to play with
though.

@fcostin - if you have time to look at this, please let me know if this is
still close enough to the original spirit in which you wrote this cool
animation.

### DIFF
--- a/frontend/play/samples/animate/splashtrig.evy
+++ b/frontend/play/samples/animate/splashtrig.evy
@@ -1,76 +1,43 @@
-sa := -0.007
-sb := -0.006
-sc := -0.005
-sd := -0.004
-se := -0.003
-radius := 5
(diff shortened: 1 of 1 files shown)


camh- (approved):
:green_circle:

PR #361, opened 2024-05-23
lexer: Reserve pkg and import keywords

juliaogris:
Reserve pkg and import keywords for later use in packages and dependency
management. I really, really, really want to do this eventually.

### DIFF
--- a/pkg/lexer/token.go
+++ b/pkg/lexer/token.go
@@ -87,6 +87,9 @@ const (
 	WHILE  // while
 	BREAK  // break
 	END    // end
+
+	PKG    // package keyword, reserved for later use
+	IMPORT // import keyword, reserved for later use
 )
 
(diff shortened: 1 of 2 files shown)


pkg/lexer/token.go
```diff
@@ -87,6 +87,9 @@ const (
 	WHILE  // while
 	BREAK  // break
 	END    // end
+
+	PKG    // package keyword, reserved for later use
+	IMPORT // import keyword, reserved for later use
```
camh-, line 92: every token has a test in `TestSingleToken()`. I guess these should be added too.

camh- (approved):
:green_circle:

pkg/lexer/token.go
juliaogris: Good point! Addressed.

Issue #362, opened 2024-05-23
Reassignment mutates all references for mutable types


joshcarp:
I think @juliaogris and I already came across this. Not sure if this is expected but it seems like variable references are mutable and not just the underlying data.
For example in this case I would expect `orig` to never change from it's original value because it doesn't ever get re-assigned:

```evy
orig := {val:0 next:{}}
copy := orig

print "orig" orig // orig {val:0 next:{}}
print "copy" copy // copy {val:0 next:{}}

copy = copy.next.({}any)

print "orig" orig // orig {}
print "copy" copy // copy {}
```

This is the equivalent python code:

```python
orig = {"val": 0, "next": {}}
copy = orig

print("orig", orig)  # orig {'val': 0, 'next': {}}
print("copy", copy)  # copy {'val': 0, 'next': {}}

copy = copy["next"]

print("orig", orig)  # orig {'val': 0, 'next': {}}
print("copy", copy)  # copy {}
```

This also seems to be the case with slices:

```
orig := [1 2 3]
copy := orig

print "orig" orig // orig [1 2 3]
print "copy" copy // copy [1 2 3]

copy = copy[1:]

print "orig" orig // orig [2 3]
print "copy" copy // copy [2 3]

```

camh-:
This is also broken for elements of arrays and maps. Some slightly simpler test cases:

Maps: https://play.evy.dev/#content=H4sIAAAAAAAAA0tUsLJVqE60MqzlSgIxE7mSFGwVqpOsjGq5Cooy80oUlBKVFBIV9PUVijPyS3NSFCCiYC0KSaUlEH4xqpYkJYUkLgC741kvWwAAAA==
```
a := {a: 1}
b := a
b = {b: 2}
print "a" a // should print {a:1} but prints {b:2}
print "b" b
```

Arrays:
https://play.evy.dev/#content=H4sIAAAAAAAAA0tUsLJViDaM5UoCMRK5khRsFaKNYrkKijLzShSUEpUUEhX09RWKM/JLc1IUIKLRhrEKSaUlEF4xsvIkJYUkLgArmnwwUwAAAA==
```
a := [1]
b := a
b = [2]
print "a" a // should print [1] but prints [2]
print "b" b
```

Map elements:
[...]

PR #363, opened 2024-05-24
evaluator: Fix assignment of composites with copied references

juliaogris:
Fix assignment of composites with copied references, so that that

	a := [1]
	b := a
	b = [2]
	print a b // [a] [b]

now does NOT update the a variable anymore. The assignment now always creates
a new value, rather than updating the existing one.

Fixes: https://github.com/evylang/evy/issues/362

### DIFF
--- a/pkg/evaluator/scope.go
+++ b/pkg/evaluator/scope.go
@@ -1,5 +1,7 @@
 package evaluator
 
+import "fmt"
+
 type scope struct {
 	values map[string]value
 	outer  *scope
@@ -28,4 +30,18 @@ func (s *scope) set(name string, val value) {
 		return
 	}
 	s.values[name] = val
+}
+
+func (s *scope) update(name string, val value) {
+	if name == "_" {
+		return
+	}
+	if _, ok := s.values[name]; ok {
+		s.values[name] = val
+		return
+	}
+	if s.outer == nil {
+		panic(fmt.Errorf("%w: update of unknown variable %q", ErrAssignmentTarget, name))
+	}
+	s.outer.update(name, val)
 }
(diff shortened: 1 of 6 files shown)


pkg/evaluator/scope.go
```diff
@@ -29,3 +31,17 @@ func (s *scope) set(name string, val value) {
 	}
 	s.values[name] = val
 }
+
+func (s *scope) update(name string, val value) {
+	if s == nil {
+		panic(fmt.Errorf("%w: cannot update nil scope for %q", ErrAssignmentTarget, name))
+	}
+	if name == "_" {
+		panic(fmt.Errorf("%w: cannot update _ variable", ErrAssignmentTarget))
```
camh-, line 40: I wonder whether it might be better to just return doing nothing for this case. Assigning to the empty var is Go is really just a no-op, so you could make that happen here. Then you would not need the checks in the various rangers.

I was going to suggest doing the same in `Set()`, but it looks like it already does just this!

pkg/evaluator/scope.go
```diff
@@ -29,3 +31,17 @@ func (s *scope) set(name string, val value) {
 	}
 	s.values[name] = val
 }
+
+func (s *scope) update(name string, val value) {
+	if s == nil {
```
camh-, line 36: Rather than doing this check here (which is not done for the other methods), perhaps move the nil check below before calling `s.outer.update()` so you can panic with a slightly better error message.

```
if s.outer == nil {
  panic(fmt.Errorf("%s: update of unknown variable %q", ErrAssignmentTarget, name))
}
```

camh- (approved):
:bug:

PR #364, opened 2024-05-27
playsite: Allow sharing code with parse errors

juliaogris:
Allow sharing Evy code via playground URLs even if the code cannot be
formatted. This should help with help, e.g. why does this code not work
`<link>`?

It'd be neat to also add sharing to the CLI, e.g. `evy share file.evy`
generating URL or `evy run --url`. We might leave this for later.

### DIFF
--- a/frontend/play/index.html
+++ b/frontend/play/index.html
@@ -184,25 +184,12 @@
           <button class="icon-close"></button>
         </header>
         <main>
+          <p class="note hidden">⚠️ Code has <span class="err">errors</span>!</p>
           <div class="copy">
             <input type="text" value="/play#gzipped+and+base64+encoded+evy+code+content" />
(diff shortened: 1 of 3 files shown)


frontend/play/index.html
```diff
@@ -184,6 +184,7 @@ <h1>Share</h1>
           <button class="icon-close"></button>
         </header>
         <main>
+          <p class="note">⚠️ Code has <span class="err">parse errors</span>!</p>
```
camh-, line 187: Is "parse error" the best term here? It's technically correct, but a bit jargony.

juliaogris: i'll change it to just `Code has errors!` ok?

camh- (approved):
:green_salad: LGTM

PR #365, opened 2024-05-27
cli: Add hidden "compile" command to CLI

camh-:
Add a `compile` command to the CLI that is hidden, like the `tokenize`
and `parse` commands. `compile` outputs the bytecode for the evy program
specified on the command line (or stdin). It is intended for debugging
the compiler and for deeper understanding of what the bytecode for a
program looks like.

### DIFF
--- a/main.go
+++ b/main.go
@@ -49,6 +49,7 @@ import (
 	"strconv"
 	"time"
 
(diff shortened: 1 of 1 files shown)


juliaogris (approved):
lgtm 🟢

PR #366, opened 2024-05-27
builtins: Add `repr` function

juliaogris:
Add `repr` function to the evaluator. The function returns a space-separate
string representation of its values that can be copied back in as valid Evy
code for most parts, similar to Python's `repr` function.

Use it in `assert` function so that `assert "2" 2` prints a less confusing
error message.

### DIFF
--- a/pkg/evaluator/value.go
+++ b/pkg/evaluator/value.go
@@ -5,12 +5,14 @@ import (
 	"strconv"
 	"strings"
 
+	"evylang.dev/evy/pkg/lexer"
 	"evylang.dev/evy/pkg/parser"
 )
 
 type value interface {
 	Equals(value) bool
 	String() string
+	Repr() string
 	Set(value)
 }
 
@@ -50,6 +52,7 @@ type breakVal struct{}
 type noneVal struct{}
 
 func (n *numVal) String() string { return strconv.FormatFloat(n.V, 'f', -1, 64) }
+func (n *numVal) Repr() string   { return n.String() }
 func (n *numVal) Equals(v value) bool {
 	n2, ok := v.(*numVal)
 	if !ok {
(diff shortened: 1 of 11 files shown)


pkg/evaluator/value.go
```diff
@@ -300,6 +323,27 @@ func (m *mapVal) String() string {
...
 
+func (m *mapVal) Repr() string {
+	pairs := make([]string, 0, len(m.Pairs))
+	for _, key := range *m.Order {
+		pairs = append(pairs, keyRepr(key)+":"+m.Pairs[key].Repr())
+	}
+	return "{" + strings.Join(pairs, " ") + "}"
+}
+
+func keyRepr(s string) string {
```
camh-, line 334: food for thought. It seems to me perhaps there could be a function `lexer.IsIdent(s)` that this function could be written in terms of:
```
func keyRepr(s string) string {
  if lexer.IsIdent(s) {
    return s
  }
  return strconv.Quote(s)
}
```

Then the `lexer` package could have `IsIdent()` which makes sense as a place for this to belong (and it already has `isLetter(r)` (which allows `_`) and `isDigit(r)`:
```
func IsIdent(s) bool {
  if s == "" {
    return false
  }
  for i, r := range s {
    if !isLetter(r) && (i > 0 && !isDigit(r)) {
      return false
    }
  }
  return true
}
```

camh- (approved):
Just a suggestion, but otherwise :green_apple: LGTM if you want to merge as is (if you do, just resolve the comment - I'm good with that)

pkg/evaluator/value.go
juliaogris: I like this a lot! Adopted - thank you.

camh- (approved):
:green_apple: LGTM

PR #367, opened 2024-05-29
parser: Fix confusing error message

juliaogris:
Fix confusing error message for "invalid return type: end of line" that
printed and associated the wrong tokens. We have now simplified the message
to `invalid return type` and associate it with the first token of the return
type declaration.


camh- (approved):
:green_salad: LGTM

PR #369, opened 2024-05-30
playsite: Add `editor=none` options to hash

juliaogris:
Add `editor=none` options to hash in preparation for classroom work: I'd like
to show students a program in execution without giving away how it's
programmed - find out what this thing does by just observing it.

This is a hack. Truth is the playground is ripe for a re-think and re-write,
ideally based around web-components. But that's a big job and this will do
for now. I hope.

---

Try with:
https://evy-lang-stage-play--369-a5xbmcq8.web.app/#sample=welcome&editor=none
https://evy-lang-stage-play--369-a5xbmcq8.web.app/#sample=splashtrig&editor=none

### DIFF
--- a/frontend/play/css/index.css
+++ b/frontend/play/css/index.css
@@ -200,10 +200,16 @@ button.share > div {
   display: flex;
   overflow: clip;
   transition: translate 0.3s ease-in-out;
+  justify-content: center;
 }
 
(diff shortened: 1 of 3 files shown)


camh- (approved):
:green_salad: Seems to work!

PR #370, opened 2024-06-01
learn: Refactor utilities in preparation for exercise

juliaogris:
Refactor various utilities in preparation for exercise modelling:

- Move out various markdown utilities from build-tools and add tests to stay above coverage threshold.
- Fix the `levy unseal` command which wasn't doing anything.
- Refactor answer key and required question file hierarchy.
- Update embedded CSS styles for questions so that they can be combined in an exercise.html file.
- Rework learn/pkg/question/testdata to prepare for only one exercise markdown per exercise dir.


camh- (approved):
:broccoli: LGTM

PR #376, opened 2024-06-05
learn: Refactor question model and MD utilties

juliaogris:
Refactor question model and various utilities in preparation
for exercise, unit, unittest, quiz and course models, see
#377 .
Add a pluggable `cls` function so that we can use `bytes.Buffer`
to capture output.
Add testdata for follow-up commits with more comprehensive course
markdown.

---

Rendered new testdata:


camh- (approved):
:leafy_green:

PR #378, opened 2024-06-12
learn: Add course catalog exporting

juliaogris:
Add course catalog exporting as single JSON file. Catalog contains all
information to generate a question set for an exercise, quiz or unittest
identified by composite IDs. It also holds the maximumScore achievable for
each course and unit. An exercise has a predefined score of 10. This will
allow for progress tracking as percentage.

Sneak in a couple of preparatory and clean-up commits: fix learn/README.md
update go dependencies and add Name field to some markdown model types.

### DIFF
--- a/learn/cmd/levy/main.go
+++ b/learn/cmd/levy/main.go
@@ -63,7 +63,7 @@ func main() {
 }
 
 type exportCmd struct {
-	ExportType        string `arg:"" enum:"html,answerkey,all" help:"Export target: one of html, answerkey, all."`
+	ExportType        string `arg:"" enum:"html,answerkey,catalog,all" help:"Export target: one of html, answerkey, catalog, all."`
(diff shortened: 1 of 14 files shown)


learn/cmd/levy/main.go
```diff
@@ -91,12 +91,13 @@ type unsealCmd struct {
 }
 
 func (c *exportCmd) Run() error {
-	if c.ExportType == "answerkey" && c.WithAnswersMarked {
+	if (c.ExportType == "answerkey" || c.ExportType == "catalog") && c.WithAnswersMarked {
```
camh-, line 94: I would change this condition to match the error message now that there are equal number of options either way:

    if c.WithAnswersMarked && c.ExportType != "all" && c.ExportType != "html"

camh- (approved):
:shamrock: LGTM

learn/cmd/levy/main.go
juliaogris: yeah, agreed. More readable too. thank you fixed.

PR #380, opened 2024-06-12
learn: Generate HTML with templates

juliaogris:
Generate learn resource HTML with templates so that we can optionally use
embedded and linked CSS in preparation for deploying to learn.evy.dev.

In a follow up commit we should be ready to deploy to learn.evy.dev.

As usual start with a couple smaller preparatory, fix-up and refactor
commits.


camh- (approved):
:green_book: LGTM

PR #381, opened 2024-06-14
cli: Support txtar format as input

juliaogris:
Add support for the txtar file format as Evy source input. This will come in
handy for creating course work with multiple choice answers where we need
many small source files.

### DIFF
--- a/main.go
+++ b/main.go
@@ -55,6 +55,7 @@ import (
 	"evylang.dev/evy/pkg/lexer"
 	"evylang.dev/evy/pkg/parser"
 	"github.com/alecthomas/kong"
+	"golang.org/x/tools/txtar"
 )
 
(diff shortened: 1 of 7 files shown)


main.go
```diff
@@ -164,6 +165,24 @@ func (c *runCmd) Run() error {
 	return nil
 }
 
+func (c *runCmd) fileBytes() ([]byte, error) {
+	b, err := fileBytes(c.Source)
+	if err != nil {
+		return nil, err
+	}
+	if filepath.Ext(c.Source) == ".txtar" && c.Txtar != "" {
```
camh-, line 173: Perhaps add another case where `c.Txtar != ""` and the source is not a .txtar file and return an error?

camh- (approved):
:leafy_green: just one little comment/suggestion

PR #382, opened 2024-06-17
learn: Add SubQuestions generated from question template

juliaogris:
Add SubQuestions generated from question.md template file, where questions and
answers are all represented in .txtar files. We then "generate" a
sub-question from each possible correct answer.

In a preparatory commit Add Filename() method to model interface.

---

Testdata is now ready for review (generic feedback also welcome in view of up-coming fundamentals course; deployment base on follow-up PR ~#383~ #384 ):


learn/pkg/learn/questionfm.go
```diff
@@ -72,3 +84,21 @@ func (f *questionFrontmatter) Unseal(privateKey string) error {
...
 	return nil
 }
+
+func validateGeneratedQuestions(gen string) error {
+	if gen == "" {
+		return nil
+	}
+	questions := splitTrim(gen)
+	if len(questions) == 1 && questions[0] != "all" {
+		return fmt.Errorf("%w: more than one question generated for generate-questions", ErrInvalidFrontmatter)
```
camh-, line 94: nit: add "must" - "more that one question must be generated..."

learn/pkg/learn/renderer.go
```diff
@@ -180,6 +181,50 @@ func (s *evySource) RenderHTML(buf *bytes.Buffer) {
...
+	b, err := os.ReadFile(filename)
+	if err != nil {
+		return nil, err
+	}
+	archive := txtar.Parse(b)
+	m := map[string]bool{}
+	for _, file := range archive.Files {
+		content := string(file.Data)
+		if m[content] {
+			return nil, fmt.Errorf("%w: found duplicate content", ErrBadTxtar)
```
camh-, line 194: can you put `file.Name` in this error message so they know which is one of the duplicates?

learn/pkg/learn/renderer.go
```diff
@@ -180,6 +181,50 @@ func (s *evySource) RenderHTML(buf *bytes.Buffer) {
...
+	}, nil
+}
+
+type txtarContent struct {
+	data       []byte
+	ResultType ResultType
+	archive    *txtar.Archive
+}
+
+// RenderOutput executes.
```
camh-, line 211: short comment? I guess you cannot render txtar content? perhaps the comment should say that?

learn/pkg/learn/renderer.go
```diff
@@ -180,6 +181,50 @@ func (s *evySource) RenderHTML(buf *bytes.Buffer) {
...
+
+// RenderHTML prints evy source code.
+func (*txtarContent) RenderHTML(buf *bytes.Buffer) {
+	buf.WriteString("*** txtar Content ERROR ***")
+}
+
+func (t *txtarContent) Archive() *txtar.Archive {
+	if t.archive == nil {
+		t.archive = txtar.Parse(t.data)
+	}
```
camh-, line 224: How come this is here? `newTxtarContent` sets the field so it should not be nil.

learn/pkg/learn/renderer.go
```diff
@@ -206,23 +251,30 @@ func newRendererFromEvyFile(filename string, resultType ResultType) (Renderer, e
 	if err != nil {
 		return nil, err
 	}
+	if strings.HasSuffix(filename, ".txtar") {
+		return newTxtarContent(filename, resultType)
+	}
```
camh-, line 256: This could go before reading all the bytes so you dont read them twice.

learn/pkg/learn/questionfm.go
juliaogris: done.

learn/pkg/learn/renderer.go
juliaogris: done, great idea.

learn/pkg/learn/renderer.go
juliaogris: ok, done. I've also added more to the type.

learn/pkg/learn/renderer.go
juliaogris: that was probably a left over from the initial implementation, apologies. I've removed it.

learn/pkg/learn/renderer.go
juliaogris: Good point! thank you. fixed.

juliaogris:
@camh- if you don't mind, could you please give it another look?

camh- (approved):
:green_apple: LGTM

PR #383, opened 2024-06-17
learn: Add verification field to question frontmatter

juliaogris:
Add verification field to question frontmatter, with values:

    - match          # default
    - none           # skip verification
    - parse-error    # correct answer(s) have parse errors
    - no-parse-error # correct answer(s) have no parse errors

I started on the fundamentals course materials and I realized I really wanted
to also have students check which program they think compiles or not. This
was a relatively easy addition.

---

Testdata is now ready for review (generic feedback also welcome in view of up-coming fundamentals course):


learn/pkg/learn/frontmatter.go
```diff
@@ -9,12 +9,14 @@ type (
...
 	answerType      string
 	difficulty      string
+	verification    string
 )
 
 var (
 	validFrontmatterTypes = []string{"course", "unit", "unittest", "quiz", "exercise", "question"}
 	validAnswerTypes      = []string{"single-choice", "multiple-choice", "free-text", "multiple-free-texts", "program"}
 	validDifficulties     = []string{"easy", "medium", "hard", "retriable"}
+	validVerifications    = []string{"match" /* default */, "none", "parse-error", "no-parse-error"}
```
camh-, line 19: I had a little trouble grokking at first the difference between `match` and `no-parse-error` since a `match` answer should not have parse errors. But now I realise the two parse-error forms are for "pick the program that compiles" and "pick the program that doesn't compile". I think. I'm not sure if there's a place to note that, as there's not really any docs for the frontmatter in more detail. Perhaps that can come later.

:pineapple:

camh- (approved):
:green_apple: LGTM

PR #384, opened 2024-06-19
learn: Add support for text-based questions

juliaogris:
Add support for text-based questions on top of multiple and single choice, for example:

    What does this program output?

    `print "hello"`

    Write your answer output below:

    ```

    ```

For this to work we must now handle the `text` question
types in the question frontmatter. We will also store expected answer text in
the answer key.

---

- [Demo deployment with answers](https://learn.evytest.dev/all/unit1/exercise-text/index.html)
- [Demo deployment without answers](https://learn.evytest.dev/html-no-private-key/unit1/exercise-text/index.html)


learn/pkg/learn/question.go
```diff
@@ -455,6 +543,23 @@ func (m *QuestionModel) buildAnswerChoicesField(block markdown.Block) error {
 	return nil
 }
 
+func (m *QuestionModel) buildAnswerTextField(block markdown.Block) error {
+	if m.AnswerText != nil || m.Question == nil {
```
camh-, line 557: :pineapple: I am finding all these checks of `m.AnswerText`, `m.AnswerChoices`, `m.Question` against `nil` (either nil or not nil) somewhat confusing. There seems to be a bit of a state machine here and it's a little hard to understand the progression exactly.

learn/pkg/learn/question.go
```diff
@@ -455,6 +543,23 @@ func (m *QuestionModel) buildAnswerChoicesField(block markdown.Block) error {
...
+		return nil
+	}
+	var err error
+	m.AnswerText, err = NewRenderer(block, answerField, m.Filename())
```
camh-, line 561: it seems odd that NewRenderer can return nil without an error - a little unexpected perhaps

learn/pkg/learn/question.go
```diff
@@ -548,6 +657,10 @@ func (m *QuestionModel) isParserErrorQuestion() bool {
 	return m.Frontmatter.Verification == "parse-error" || m.Frontmatter.Verification == "no-parse-error"
 }
 
+func (m *QuestionModel) isTextQuestion() bool {
```
camh-, line 670: this function is open-coded in one other place (although in that place I did suggest replacing it with a switch). Either way use the function everywhere or nowhere I think

learn/pkg/learn/question.go
```diff
@@ -265,21 +265,31 @@ func (m *QuestionModel) printTextAnswerHTML(buf *bytes.Buffer, withAnswersMarked
...
 	if !strings.HasSuffix(content, "\n") {
 		content += "\n"
 	}
 	return content, nil
 }
 
+func removeCommentTaggedLines(s string) string {
+	lines := strings.Split(s, "\n")
+	for i, line := range lines {
+		if strings.HasSuffix(line, " //levy:blank") {
```
camh-, line 286: would `hide` be better than `blank`? I'm really not sure what the purpose of this annotation is - the commit does not say why you want this.

camh- (commented):
Some comments, but otherwise LGTM. I'm sorry but I really could not get into depth here - I tried but I think it might be best to just push forward and look at the results once all your work is done, and at that point I think I will have a better mental model to be able to assess the structure. Until then, my comments will be mostly narrowly focused on lines and simple things I think.

camh-:
The PR description says something about handling `multiple-texts`, but I don't see anything in this PR for that. Should the PR description be updated?

learn/pkg/learn/question.go
juliaogris: ok, after discussion in person: `hide` doesn't sit right for me as my CSS bent 
brain reads hide as don't show at all including new line. Here the annotation
is meant to mean: `turn it into a "fill-in-the-blank" type exercise`.

I've updated the commit message to better reflect this but left the 
annotation as is.

juliaogris:
> The PR description says something about handling `multiple-texts`, but I don't see anything in this PR for that. Should the PR description be updated?

Fixed.

learn/pkg/learn/question.go
```diff
@@ -358,6 +413,39 @@ func generateParseErrors(t *txtarContent) []bool {
 	return parseErrors
 }
 
+func (m *QuestionModel) buildQuestionModel() error {
+	if m.Frontmatter.AnswerType == "text" {
+		return m.buildQuestionModelForText()
+	}
+	return m.buildQuestionModelForChoice()
```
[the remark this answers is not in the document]
juliaogris: done.

learn/pkg/learn/question.go
```diff
@@ -286,6 +322,13 @@ func (m *QuestionModel) getVerifiedAnswer() (Answer, error) {
 }
 
 func (m *QuestionModel) verifyMatch(answer Answer) error {
+	if len(m.AnswerChoices) != 0 {
```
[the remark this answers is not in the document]
juliaogris: done.

learn/pkg/learn/question.go
juliaogris: ok, i just cannot not respond to a 🍍 .
I think you have point. The comparison doesn't belong here and shows again that this code needs a holistic revision when done.
I've hoisted it up, which should make it clearer hopefully.

learn/pkg/learn/question.go
juliaogris: ok, after some debate we settled on renaming it to `AsRenderer`

learn/pkg/learn/question.go
juliaogris: ok, removed

camh- (approved):
:green_book:

PR #385, opened 2024-06-22
learn: Prepare for txtar support by levy

juliaogris:
- Factor out `QuestionModel.correctAnswerIndices` method for later reuse by
  txtar.
- Remove embedMD type in QuestionModel as it is not used other than in tests.
- Add Filename() method to model interface, which is necessary in
  QuestionModel based on txtar sub-questions that don't have an underlying
  filename.
- Ensure testdata descriptions match the question and answer.

And further testdata and error message improvements.


camh- (approved):
:four_leaf_clover: LGTM

PR #388, opened 2024-06-26
learn: Add sidebar to learn pages

juliaogris:
Create sidebar, one per course with absolute URL paths, so that we can re-use
the same sidebar for any sub-entries of the course (unit, exercise,....).
Create the sidebar structure similar to the sidebar of the doc-site and
following the hierarchy of

    course -> unit -> exercise

Use absolute paths for linking so that we can use the same sidebar for any
file within this hierarchy.

In preparatory commits, factor out sidebar.css from docs, wrap badges in their
own `<div>` for styling, and add a Document() method to the model interface.


camh- (approved):
:green_apple: I didn't look much at the css/html/js stuff, :moneybag:

PR #389, opened 2024-06-26
deploy: Use special root path re-write for learn subdomain

juliaogris:
Use special root path re-write for learn subdomain where we replace
`/lear/PATH` with `/PATH`. This allows for absolute paths without domain used
as links, in particular in sidebar links.


camh- (approved):
:green_apple: I assume some of this will need a further update for the new `labs` subdomain?

PR #390, opened 2024-07-02
learn-site: Add SPA links, loading only page fragments

juliaogris:
Capture `onclick` event of course link: links starting with the same
courseRoot directory derived from the sidebar heading and all other relative
links.

Instead of default link navigation load fragment of href (href + "f") and
replace inner HTML (`...`) of

    <main>
      <div class="max-width-wrapper">
        ...
      </div>
    </main>

Fiddle with HS History API so that the browser back and forwards buttons
still work as expected. Fun.

In preparatory commit generate HTML Fragments with .htmlf ending - the same
files as the regular .html files without doctype, html, head, body, and other
tags that are not needed when we only want to replace the main content of the
page:

    <!doctype html>
    <html lang="en">
      <head>...</head>
      <body>
        <header>...</header>
        <nav>...</nav>
        <main>
           <div class="max-width-wrapper">
                
           </div>
        </main>
      </body>
    </html>

Sneak in a mini style change making the content narrower.

### DIFF
--- a/learn/cmd/levy/main.go
+++ b/learn/cmd/levy/main.go
@@ -70,7 +70,6 @@ type exportCmd struct {
 	PrivateKey        string `short:"k" help:"Secret private key to decrypt sealed answers." env:"EVY_LEARN_PRIVATE_KEY"`
 	WithAnswersMarked bool   `short:"m" help:"Include marked answers in HTML output. Cannot be used with export target answerkey."`
 	SelfContained     bool   `short:"c" help:"Ensure .html files are self-contained (embed CSS, JS, Favicon)." negatable:""`
-	SkipQuestions     bool   `short:"q" help:"Skip exporting individual questions, export full exercise only."`
 	RootDir           string `short:"r" default:"/" help:"Root directory for HTML links, e.g. /learn/."`
 }
 
@@ -103,7 +102,6 @@ func (c *exportCmd) Run() error {
 		WriteCatalog:      c.ExportType == "catalog" || c.ExportType == "all",
 		WithAnswersMarked: c.WithAnswersMarked,
 		SelfContained:     c.SelfContained,
-		SkipQuestions:     c.SkipQuestions,
 		RootDir:           c.RootDir,
 	}
 	modelOptions := getOptions(c.IgnoreSealed, c.PrivateKey)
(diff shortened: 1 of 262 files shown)


learn/pkg/learn/export.go
```diff
@@ -195,11 +187,20 @@ func writeHTMLFile(model model, sidebarContent string, srcDir, destDir string, o
 	if err != nil {
 		return fmt.Errorf("%w: %w: %s", ErrInconsistentMdoel, err, model.Filename())
 	}
-	htmlFile := filepath.Join(destDir, md.HTMLFilename(mdFile))
-	content, err := model.ToHTML(opts.WithAnswersMarked)
+	content, err := model.ToHTML(opts.WithAnswersMarked) // without answers marked
```
camh-, line 190: I dont understand this comment you've added - the parameter specifies whether to export with answers marked or not. How does this comment apply?

learn/cmd/levy/main.go
```diff
@@ -70,7 +70,6 @@ type exportCmd struct {
 	PrivateKey        string `short:"k" help:"Secret private key to decrypt sealed answers." env:"EVY_LEARN_PRIVATE_KEY"`
 	WithAnswersMarked bool   `short:"m" help:"Include marked answers in HTML output. Cannot be used with export target answerkey."`
 	SelfContained     bool   `short:"c" help:"Ensure .html files are self-contained (embed CSS, JS, Favicon)." negatable:""`
-	SkipQuestions     bool   `short:"q" help:"Skip exporting individual questions, export full exercise only."`
```
camh-, line 73: Why have you removed this? It does not seem related to the commit as described in the commit message.

camh- (approved):
:green_book: LGTM with just a little confusion.

Also the 4th commit message says "HS History API" - I don't know what HS is. Perhaps it was meant to be JS, but that's a given since it is browser code. Either expand "HS" to what it means (and Google didn't help me here) or perhaps just remove it.

juliaogris:
> 📗 LGTM with just a little confusion.
> 
> Also the 4th commit message says "HS History API" - I don't know what HS is. Perhaps it was meant to be JS, but that's a given since it is browser code. Either expand "HS" to what it means (and Google didn't help me here) or perhaps just remove it.

yeah, it was meant to be JS History API. nvm, I've removed it.

learn/pkg/learn/export.go
juliaogris: it doesn't. removed, thank you.

learn/cmd/levy/main.go
juliaogris: I guess I have removed this, IIRC, because I now see that I won't need it. I will have to export individual questions _always_ as htmlf. This will become clearer when I implement exercises/quiz/unittest question selection. I've updated the commit.

camh- (approved):
:green_book:

PR #391, opened 2024-07-02
doc-site: Use shared spa and sidebar JS modules

juliaogris:
Use shared spa and sidebar JS modules so that the learn-site and the
docs-site work exactly the same way. This shook out a few bugs in the JS
modules and seems to be the cleaner approach.

In preparation, add minor docs content fixes and fix mobile sidebar
scroll-bugs on both learn and docs sites.

Further, generate docs fragment and prepare docs deployment in separate public
directory just as we did in previous commits for the learn-site.

Fully factor our sidebar JS for reuse in docs and learn sites.

---

https://evy-lang-stage-learn--391-bcb5iuw2.web.app/foundation
https://evy-lang-stage-docs--391-f1sq8u1p.web.app/


camh- (approved):
:green_book: LGTM

PR #392, opened 2024-07-07
bytecode: Implement local variables

Rossiar:
Add scopes to the compiler so that it can differentiate whether a
variable is global or local. A local variable is one that is not defined
at the top-level of the program - i.e. it is either in a block at the
top-level of the program or in a function (although functions are not
yet implemented in the compiler).

Space is allocated at the top of the stack for the locals needed, with
the stack pointer initialised to start after that. The space allocated
for locals is reused where it can be, for example, when two blocks at
the same level have a local variable, they can share the same space as
they both cannot be in scope at the same time.

### DIFF
--- a/pkg/bytecode/code.go
+++ b/pkg/bytecode/code.go
@@ -10,11 +10,11 @@ const (
 	// OpConstant defines a constant that will be referred to by index
 	// in the bytecode.
 	OpConstant Opcode = iota
-	// OpGetGlobal retrieves a symbol from the symbol table at the
+	// OpGetGlobal retrieves a symbol from the global symbol table at the
 	// specified index.
 	OpGetGlobal
-	// OpSetGlobal adds a symbol to the specified index in the symbol
-	// table.
+	// OpSetGlobal adds a symbol to the specified index in the global
+	// symbol table.
 	OpSetGlobal
 	// OpDrop pops and discards the top N elements of the stack.
 	OpDrop
--- a/pkg/bytecode/code.go
+++ b/pkg/bytecode/code.go
@@ -18,6 +18,14 @@ const (
 	OpSetGlobal
 	// OpDrop pops and discards the top N elements of the stack.
 	OpDrop
+	// OpGetLocal retrieves a symbol from the current local-scoped
+	// symbol table at the specified index. This index also corresponds
+	// to its location on the VM stack.
+	OpGetLocal
+	// OpSetLocal adds a symbol to the specified index in the current
+	// local-scoped symbol table. This index also corresponds to its
+	// location on the VM stack.
+	OpSetLocal
 	// OpAdd instructs the virtual machine to perform an addition.
 	OpAdd
 	// OpSubtract instructs the virtual machine to perform a subtraction.
@@ -137,6 +145,8 @@ var definitions = map[Opcode]*OpDefinition{
 	OpGetGlobal: {"OpGetGlobal", []int{2}},
 	OpSetGlobal: {"OpSetGlobal", []int{2}},
 	OpDrop:      {"OpDrop", []int{2}},
+	OpGetLocal:  {"OpGetLocal", []int{2}},
+	OpSetLocal:  {"OpSetLocal", []int{2}},
 	// Operations like OpAdd have no operand width because the virtual
 	// machine is expected to pop the values from the stack when reading
 	// this instruction.
--- a/pkg/bytecode/compiler.go
+++ b/pkg/bytecode/compiler.go
@@ -30,7 +30,7 @@ var (
 type Compiler struct {
 	constants    []value
 	instructions Instructions
-	globals      *SymbolTable
+	symbolTable  *SymbolTable
 	// breaks tracks the positions of break statements in the inner-most loop.
 	breaks []int
 }
@@ -43,7 +43,7 @@ type Bytecode struct {
 
 // NewCompiler returns a new compiler.
 func NewCompiler() *Compiler {
-	return &Compiler{globals: NewSymbolTable()}
+	return &Compiler{symbolTable: NewSymbolTable()}
 }
 
 // Compile accepts an AST node and renders it to bytecode internally.
@@ -279,7 +279,7 @@ func (c *Compiler) compileForStatement(stmt *parser.ForStmt) error {
 		// declare the loop var with a noneVal to start with. The top
 		// of the loop will overwrite this with a value of the correct type
 		// for the loop.
-		symbol := c.globals.Define(stmt.LoopVar.Name)
+		symbol := c.symbolTable.Define(stmt.LoopVar.Name)
 		if err := c.emit(OpNone); err != nil {
 			return err
 		}
@@ -302,7 +302,7 @@ func (c *Compiler) compileForStatement(stmt *parser.ForStmt) error {
 	// Assign the current loop value to the loop var. The rangeOp has left it
 	// on the stack if we told it we have a loop var (hasLoopVar).
 	if stmt.LoopVar != nil {
-		symbol, ok := c.globals.Resolve(stmt.LoopVar.Name)
+		symbol, ok := c.symbolTable.Resolve(stmt.LoopVar.Name)
 		if !ok {
 			return fmt.Errorf("%w %s", ErrUndefinedVar, stmt.LoopVar.Name)
 		}
@@ -317,7 +317,6 @@ func (c *Compiler) compileForStatement(stmt *parser.ForStmt) error {
 	if err := c.Compile(stmt.Block); err != nil {
 		return err
 	}
-
 	if err := c.emit(OpJump, topOfLoop); err != nil {
 		return err
 	}
@@ -487,7 +486,7 @@ func (c *Compiler) compileDecl(decl *parser.Decl) error {
 	if err := c.Compile(decl.Value); err != nil {
 		return err
 	}
-	symbol := c.globals.Define(decl.Var.Name)
+	symbol := c.symbolTable.Define(decl.Var.Name)
 	return c.emit(OpSetGlobal, symbol.Index)
 }
 
@@ -497,7 +496,7 @@ func (c *Compiler) compileAssignment(stmt *parser.AssignmentStmt) error {
 	}
 	switch target := stmt.Target.(type) {
 	case *parser.Var:
-		symbol, ok := c.globals.Resolve(target.Name)
+		symbol, ok := c.symbolTable.Resolve(target.Name)
 		if !ok {
 			return fmt.Errorf("%w %s", ErrUndefinedVar, target.Name)
 		}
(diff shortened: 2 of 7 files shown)


pkg/bytecode/vm.go
```diff
@@ -70,6 +70,14 @@ func (vm *VM) Run() error {
 			globalIndex := ReadUint16(vm.instructions[ip+1:])
 			ip += 2
 			vm.globals[globalIndex] = vm.pop()
+		case OpSetLocal:
```
juliaogris, line 73: reverse order with `OpGetLocal` for consistency with `OpGet|SetGlobal`

pkg/bytecode/symbol.go
```diff
@@ -27,20 +31,37 @@ func NewSymbolTable() *SymbolTable {
...
+}
+
 // Define adds a symbol definition to the table or returns an
 // already defined symbol with the same name.
 func (s *SymbolTable) Define(name string) Symbol {
 	if existing, found := s.store[name]; found {
 		return existing
 	}
-	symbol := Symbol{Name: name, Index: len(s.store), Scope: GlobalScope}
+	symbol := Symbol{Name: name, Index: len(s.store)}
```
juliaogris, line 46: We think `Index` should be the some of the length of all the parent stores.

pkg/bytecode/vm_test.go
```diff
@@ -595,6 +595,22 @@ func TestIf(t *testing.T) {
...
+				x := 2
+				x = x + 1
+			else if false
+				x := 3
+				x = x + 1
+			else if false
+				x := 4
+				x = x + 1
+			end
+			x = x`,
```
juliaogris, line 611: How about a test like:

```
x := 3
if true
   x := 2
   x = x
end
if x == 3
   // this branch should be taken, 
   // we think it isn't because shadowing isn't implemented:
    x = 1 
end
// ensure x is 1 NOT 2
```

pkg/bytecode/compiler.go
```diff
@@ -30,7 +30,7 @@ var (
 type Compiler struct {
 	constants    []value
 	instructions Instructions
-	globals      *SymbolTable
+	symbolTable  *SymbolTable
```
juliaogris, line 33: Can this rename please be done in initial separate commit?

pkg/bytecode/code.go
```diff
@@ -10,14 +10,22 @@ const (
 	// OpConstant defines a constant that will be referred to by index
 	// in the bytecode.
 	OpConstant Opcode = iota
-	// OpGetGlobal retrieves a symbol from the symbol table at the
+	// OpGetGlobal retrieves a symbol from the global symbol table at the
```
juliaogris, line 13: Can this rewording be done together with the rename in initial separate commit?

juliaogris (commented):
Sorry @Rossiar that it took so long to give you any feedback. @camh- and I did have a start on it several times but this is actually a little bit tricky stuff.

* Would be great to have the renames as a separate commit (stuff around global symbol table rename) - that way it's like: this commit is just the rename, the following has the meat (easier for us to take in).
* We think you are trying to implement block scoping and local scopes in one hit, but you should proabably address proper block scoping first and alone. We think there are some issues with block scoping 
* We believe Variable shadowing is not addressed properly.

### Examples for shared understanding

#### **Local scope**  - does not exist yet

```
a := a

func f1:
  a1 := "a1"
  print a1 a
end

func f2:
  a2 := "a2" // a1 NOT known here (local); a known here (global)
  print a2 a
end
```

#### **Block scope** - DOES exist

```
x := 12
print x
if x == 0 
   y := "banana"
   print x y
end
// y not available here
```

#### **Variable shadowing** - DOES exist

```
x := 12
print x
if x == 0 
   x := 14
   print x
end
print x // prints 14 we think, but should be 12. could be tested via conditionals.
```

Rossiar:
@juliaogris @camh- thanks for the review, I have split out the renaming of `globals` and fixed shadowing. I had missed one invocation of `emitSymbol` inside `compileDecl`, which was leading to the shadowing test that you provided failing. Good catch!

On block scope vs local scope - I think we can treat them in the same way (this is what Thorsten does in the book). Symbols in a block have a scoped lifetime just like local symbols in a function. Perhaps the naming is confusing, but I have another [compiler-functions](https://github.com/evylang/evy/compare/main...compiler-functions) that is building on what I have here with support for local scopes inside functions. I could have done a better job of explaining "the vision" here, as it gets much more tricky in that PR where I introduce stack frames to the VM.

pkg/bytecode/symbol.go
```diff
@@ -27,20 +31,37 @@ func NewSymbolTable() *SymbolTable {
...
+	if s.Outer == nil {
```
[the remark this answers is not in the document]
camh-: I think you are right, reading the spec and looking at the evaluator. A "global" is only at the very top level. Block-scoped variables for global-scoped code are not global variables. But this implementation does not seem to reserve space on the stack for the block-level sort-of-global variables, perhaps. I'll look at it again with that updated understanding in mind.

pkg/bytecode/vm_test.go
camh-: There is an existing test that covers shadowing, so I'll just leave that.

pkg/bytecode/symbol.go
```diff
@@ -27,20 +37,40 @@ func NewSymbolTable() *SymbolTable {
 	}
 }
 
+func newEnclosedSymbolTable(outer *SymbolTable) *SymbolTable {
+	return &SymbolTable{
+		store: make(map[string]Symbol),
+		Outer: outer,
+		index: outer.index,
```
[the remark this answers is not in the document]
camh-: I've addressed this in the way mentioned - when outer is the global symbol table, `index` starts at zero, not `outer.index`.

pkg/bytecode/vm.go
```diff
@@ -70,6 +70,14 @@ func (vm *VM) Run() error {
...
 			ip += 2
 			vm.globals[globalIndex] = vm.pop()
+		case OpGetLocal:
+			idx := ReadUint16(vm.instructions[ip+1:])
+			ip += 2
+			err = vm.push(vm.stack[idx])
+		case OpSetLocal:
+			idx := ReadUint16(vm.instructions[ip+1:])
+			ip += 2
+			vm.stack[int(idx)] = vm.pop()
```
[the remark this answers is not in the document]
camh-: I've addressed this by adding a `SymbolTable.Pop()` method that propagates the index back up the scopes, summing and maxing, so we can tell how much space is needed for all the locals. That max is used to initialise the VM stack pointer so there is space for those locals.

pkg/bytecode/compiler.go
```diff
@@ -533,3 +538,20 @@ func (c *Compiler) compileIndexExpression(expr *parser.IndexExpression) error {
...
+	c.symbolTable = newEnclosedSymbolTable(c.symbolTable)
+}
+
+func (c *Compiler) leaveScope() {
+	c.symbolTable = c.symbolTable.Outer
+}
+
+// emitSymbol will emit an OpSetLocal or an OpSetGlobal depending upon
+// the scope of the provided symbol.
+func (c *Compiler) emitSymbol(symbol Symbol) error {
```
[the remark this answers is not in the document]
camh-: Done

pkg/bytecode/symbol.go
```diff
@@ -27,20 +37,40 @@ func NewSymbolTable() *SymbolTable {
...
 // Resolve returns the Symbol with the specified name,
-// or false if there is no such Symbol.
+// or false if there is no such Symbol. If this symbol table
+// is enclosed within another, then it will attempt to resolve
+// the symbol name with the outer table.
 func (s *SymbolTable) Resolve(name string) (Symbol, bool) {
 	obj, ok := s.store[name]
+	if !ok && s.Outer != nil {
+		obj, ok = s.Outer.Resolve(name)
+		return obj, ok
```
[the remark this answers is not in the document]
camh-: Done

camh- (approved):

camh-:
I've squashed my changes into the main commit and rebased on main. I added a couple more small changes to address some of the other comments, and renamed `Outer` to `outer` as it was the only field in `SymbolTable` that was exported, and that didn't make much sense.

I'll merge this tomorrow (Monday).

pkg/bytecode/symbol.go
```diff
@@ -18,6 +21,17 @@ type Symbol struct {
 // SymbolTable is a mapping of string identifiers to symbols.
 type SymbolTable struct {
 	store map[string]Symbol
+	// index keeps track of the number of locals inside this scope,
+	// an enclosed SymbolTable will inherit a starting index from
+	// its outer table.
+	index int
+	// nestedMaxIndex is the maximum index of any inner symbol
+	// tables. It propagates up to the top-level symbol table
+	// for a function or block-scoped globals.
```
juliaogris, line 30: I think we use the word "global" differently. In the VM I understand it to mean: top level variables available everywhere excluding top-level block-scoped variables. So this comment confuses me a little as `globals` seem to include block-scoped top-level variables.

Might be worth while to step back and write a clean description on the purpose and workings of the `SymbolTable` as it seems to be such a crucial piece in understanding how all variables and scopes and their handling inside the compiler and VM hang together.

(summarizing an IRL discussion with @camh- who's volunteered to provide such a description if i haven't misunderstood).

pkg/bytecode/symbol.go
```diff
@@ -27,20 +41,61 @@ func NewSymbolTable() *SymbolTable {
...
 	if existing, found := s.store[name]; found {
 		return existing
 	}
-	symbol := Symbol{Name: name, Index: len(s.store), Scope: GlobalScope}
+	symbol := Symbol{Name: name, Index: s.index}
+	s.index++
+	if s.outer == nil {
+		symbol.Scope = GlobalScope
+	} else {
+		symbol.Scope = LocalScope
```
juliaogris, line 84: So just to be clear: local means any block scope or function level scoped variable, correct?

pkg/bytecode/symbol.go
```diff
@@ -27,20 +41,61 @@ func NewSymbolTable() *SymbolTable {
...
+	}
 	s.store[name] = symbol
 	return symbol
 }
 
 // Resolve returns the Symbol with the specified name,
-// or false if there is no such Symbol.
+// or false if there is no such Symbol. If this symbol table
+// is enclosed within another, then it will attempt to resolve
+// the symbol name with the outer table.
```
juliaogris, line 93: recursively with the outer tables?

juliaogris (commented):
lgtm, I think - I still think this stuff is pretty tricky to follow along sort of from the passenger seat, but I will do my best.
(this is no criticism on the code, just an observation on the nature of this task).
lgtm module some potential comment improvements.

pkg/bytecode/symbol.go
camh-: Done. I've taken your description of SymbolTable and added that and re-worded this comment to remove "block-scope globals"

pkg/bytecode/symbol.go
camh-: Yes - anything that is not at the top-level of the program.

pkg/bytecode/symbol.go
camh-: Re-worded this as it was not really correct. Have mentioned recursing.

camh-:
@juliaogris PTAL. I have responded to all your comments.

PR #397, opened 2024-07-22
github: Add linguist-generated git attribute

juliaogris:
Add linguist-generated git attribute for generated .html and .htmlf files.
This is in attempt to make PR reviews more manageable when the templates for
the docs or learn sites get updated.


camh- (approved):
:green_circle:

PR #398, opened 2024-07-22
ai: Remove ai subdomain

juliaogris:
Remove ai subdomain and all related files, as it looks like we are still a
good way out from actually working on it. Josh is pretty busy with other
things. In the meantime there's no need to add a deployment on every PR push
and merge to master.


camh- (approved):
:broccoli:

PR #400, opened 2024-08-01
e2e: Rename golden snapshots for docker platform

juliaogris:
Rename golden snapshots for docker platform from `linux` to `docker`. This is so
that linux devs 🐣🐧 can have different local Linux snapshots to what is
generated via the playwright jammy docker image. In my case they seem to be
different because I have locally the Fira Code font pre-installed. I have tried
waiting for fonts to be loaded without luck 🤷‍♂.


camh- (approved):
:green_apple:

PR #401, opened 2024-08-01
play-site: Improve initialization concurrency

juliaogris:
Improve initialization concurrency, so that we load sample.json, sample content
in parallel to the large evy.wasm file. The difference can be observed in
browser developer tools with network throttling enabled and watching the
waterfall - instead of the previous:

	|--- evy.wasm -------------------------|
	                                       |-- sample.json -|
	                                                        |-- welcome.evy -|

we now have

	|--- evy.wasm -------------------------|
	|-- sample.json -|
	                 |-- welcome.evy -|

This means that for any hash other than #welcome the editor contents are visible
and syntax colored faster.

### DIFF
--- a/frontend/play/index.js
+++ b/frontend/play/index.js
@@ -22,10 +22,8 @@ let editorHidden = false
 
 // --- Initialize ------------------------------------------------------
 
-await initWasm()
-initUI()
-initCanvas()
-initThemeToggle("#dark-theme", "theme")
+await Promise.all([initWasm(), initUI()])
+await format()
 
 // --- Wasm ------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


frontend/play/index.js
```diff
@@ -342,6 +340,8 @@ function clearOutput() {
 // --- UI: initialization ----------------------------------------------
 
 async function initUI() {
+  initCanvas()
+  initThemeToggle("#dark-theme", "theme")
```
camh-, line 344: These used to be done after all the other stuff in initUI, but now they are being done first before the other init functions at the end. Should these init calls all be grouped together at the end?

juliaogris: nope, i have decided they are ok to be done first.
they are not async and therefore take "no" time.

juliaogris: as i played everything in slow I also realized this way there is one race condition less, though I forgot which one now...

camh- (approved):
:broccoli:

PR #406, opened 2024-08-06
fmt: Create temp file in same dir as source file

camh-:
Create the temporary file for formatting evy source in the same
directory as the source file so that we can rename it over the original
source file if necessary. A rename cannot be across filesystems so the
only valid place for the temp file is in the same directory as the file
that will be overwritten by the rename. Without that, you may get an
error like:

    evy: error: hsl.evy: rename /tmp/evy3084141908 hsl.evy: invalid cross-device link


juliaogris (approved):
:four_leaf_clover:

PR #412, opened 2024-08-17
wasm: Set lower bound for sleep function

phy1um:
When calling `sleep` in the JS runtime, enforce a minimum allowed
interval of 1 millisecond.

This change is motivated by watching students use evy. I found they
experimented with very large and very small numbers. When writing a loop
with 10,000,000,000+ iterations containing `sleep 0`, their browser tabs
crash. For example, run this program against `main`:

```evy
i:num

clear "blue"
while i < 100000
    move i 50
    color "yellow"
    circle 5
    i = i + 0.1
    sleep 0
end
```

This is enough to freeze the browser tab on my PC for a few seconds, and
the evy UI is unresponsive even after the render finishes. In my branch,
it runs smoothly - and slowly - with the STOP button still functional.

I experimented with a few values and found a minimum sleep of 1
millisecond (0.001 seconds) is enough to ensure that the browser remains
responsive.

---

## Questions
- Is this the correct approach to this problem? Should we enforce this at all, or should the compiler (?) emit a warning (?) for cases where we see `sleep 0`.
- Does other hardware behave the same as my setup? Please run this locally and let me know!
- I have reused `minSleepDur`, should I define a second constant?

### DIFF
--- a/pkg/wasm/imports.go
+++ b/pkg/wasm/imports.go
@@ -26,11 +26,19 @@ func newJSRuntime() *jsRuntime {
 	return &jsRuntime{yielder: newSleepingYielder()}
 }
 
-func (rt *jsRuntime) Yielder() evaluator.Yielder       { return rt.yielder }
-func (rt *jsRuntime) Print(s string)                   { jsPrint(s) }
-func (rt *jsRuntime) Cls()                             { jsCls() }
-func (rt *jsRuntime) Read() string                     { return rt.yielder.Read() }
-func (rt *jsRuntime) Sleep(dur time.Duration)          { rt.yielder.Sleep(dur) }
+func (rt *jsRuntime) Yielder() evaluator.Yielder { return rt.yielder }
+func (rt *jsRuntime) Print(s string)             { jsPrint(s) }
+func (rt *jsRuntime) Cls()                       { jsCls() }
+func (rt *jsRuntime) Read() string               { return rt.yielder.Read() }
+
+func (rt *jsRuntime) Sleep(dur time.Duration) {
+	// Enforce a lower bound to stop browser tabs from freezing.
+	if dur < minSleepDur {
+		dur = minSleepDur
+	}
+	rt.yielder.Sleep(dur)
+}
+
 func (rt *jsRuntime) Move(x, y float64)                { move(x, y) }
 func (rt *jsRuntime) Line(x, y float64)                { line(x, y) }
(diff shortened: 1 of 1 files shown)


camh- (commented):
We should update the built-in docs with this too I think. When i get a moment later today or tomorrow, I'll reformat the PR description so it works properly as a commit message (it will get used for the merge commit).

juliaogris:
> We should update the built-in docs with this too I think. When i get a moment later today or tomorrow, I'll reformat the PR description so it works properly as a commit message (it will get used for the merge commit).

@phy1um - so that means word-wrapping at 80 char, commit message type heading,
indent over tripple-back tick and for editorial content, ie stuff that should _not_
go in the merge commit we use a thematic break `---` - so that should probably go
before `## Questions`

camh-:
> so that means word-wrapping at 80 char

Actually, 72 chars for commit messages. Anyway I've reformatted it to 72-char lines and added the editorial separator, and updated the title to use imperative mood, as per https://cbea.ms/git-commit/ 

When we merge this, the commit message of the individual commit will be ignored as PRs with a single commit get squash merged to main for which the commit message comes from the PR. We use merge commits when there is more that one commit in a PR, which implies when you update this PR, please fold the changes into the existing commit.

We don't merge with the GitHub merge button, but use an external merge script instead. If you're interested: https://github.com/foxygoat/git-scripts/blob/master/git-pr-merge

juliaogris:
I've cherry-picked the commit of this PR and the excellent 
PR description into PR #418, it is already merged.

PRs from fork don't pass CI, because we deploy to Firebase on 
each push, which is great for reviewing new features live.
It also requires the Firebase deployment secrets which are kept
as GitHub secrets which are for security reasons not exposed to
PRs from fork. 

@camh- and I have tried to find a workaround to that, but it is
rather involved, so we will just stick with the PR from  evylang/evy
branch rather than fork for now. I'm working on a CONTRIBUTING.md
that will cover this.

juliaogris:
I've cherry-picked the commit of this PR and the excellent 
PR description into PR #418, it is already merged.

PRs from fork don't pass CI, because we deploy to Firebase on 
each push, which is great for reviewing new features live.
It also requires the Firebase deployment secrets which are kept
as GitHub secrets which are for security reasons not exposed to
PRs from fork. 

@camh- and I have tried to find a workaround to that, but it is
rather involved, so we will just stick with the PR from  evylang/evy
branch rather than fork for now. I'm working on a CONTRIBUTING.md
that will cover this.

PR #413, opened 2024-08-17
pkg/cli/svg: fix minor typo

phy1um:
Storke -> Stroke


juliaogris:
I've cherry-picked the commit of this PR into PR #418, it is 
already merged.

PRs from fork don't pass CI, because we deploy to Firebase on 
each push, which is great for reviewing new features live.
It also requires the Firebase deployment secrets which are kept
as GitHub secrets which are for security reasons not exposed to
PRs from fork. 

@camh- and I have tried to find a workaround to that, but it is
rather involved, so we will just stick with the PR from  evylang/evy
branch rather than fork for now. I'm working on a CONTRIBUTING.md
that will cover this.

PR #414, opened 2024-08-21
wip: Deploy from fork

juliaogris:
This is all just experimentation at this stage - but can only be done via PR, so
here we go:

THIS IS A TEST PR


camh- (approved):

juliaogris:
RIGHT. this is too hard.
for the full PR to pass it must also in the future continued to be created from a
branch on evylang/evy. This will make outside contriubtions harder.
We will carry the pain.

PR #415, opened 2024-08-21
wip: Refactor deployment in preparation for on-fork deployment

juliaogris:


juliaogris:
RIGHT. this is too hard.
for the full PR to pass it must also in the future continued to be created from a
branch on evylang/evy. This will make outside contriubtions harder.
We will carry the pain.

PR #416, opened 2024-08-21
lab: Create `if` statement lab

juliaogris:
Create `if` statement lab. This is our third lab. I've tested it with my
daughter Mali (thank you!). She managed to do pretty much all of it by
herself with the code hints. The Monte Carlo Pi simulation was the
trickiest I think, but that's more intended as a walk through on the
facilitators side anyway.

The content is on the wordy side but I intend to add some [[ NEXT ]]
buttons within the lab notes in the follow up commit.

We start off with a couple of minor fix-ups to little bugs that were discovered
when building this lab:

* Fix SVG exports if initial position is not explicitly set to `0 0`
* Tweak CSS for landing page text.

---

Preview also available at:
https://lab.evytest.dev#welcome-ifs


camh- (approved):
:green_apple: :green_book:

PR #418, opened 2024-08-22
builtin: Rename evy builtin `assert` to `test`

juliaogris:
Rename evy builtin `assert` to `test` as test seems to be a more
intuitive name than assert. Full disclosure @camh- said all along it
should `test`, but I felt the concept of a test is an abstraction level
higher, a collection of many asserts or so. I had to rename all
human-eval samples from `func test` to `func testAll`.

Still it seems with a bit of distance and reading the builtin docs
afresh: test is the more intuitive name.

Fix up all examples and re-generate docs in separate commits for better
reviewability.


docs/builtins.md
```diff
@@ -511,75 +511,75 @@ line 4 column 5: scale must be positive
...
-`assert` is used to check if a condition is true or two values – `want` and
+`test` is used to check if a condition is true or two values – `want` and
 `got` – are the same. If the condition is not true or the two values are not
-the same, the program will print the failed assertion and terminate with exit
+the same, the program will print the failed test and terminate with exit
 status 1.
 
-`assert` optionally takes a message or a formatted message with arguments to
-print along with the failed assertion as a third and following arguments:
+`test` optionally takes a message or a formatted message with arguments to
```
camh-, line 521: I think "formatted message" is wrong here. It should say something like "format string". The text below is more specific about sprintf, etc so I think this general term is fine.

camh- (approved):
:cactus:

PR #419, opened 2024-08-22
e2e: Reduce dialog snapshot test flakiness

juliaogris:
Reduce dialog snapshot test flakiness, or at least attempt it. We just
add a 1 sec wait before we take the dialog snapshot which seems to have
removed any local flakiness. I'm really unsure what the right way to
fix this would be, but snapshot tests have been useful (and painful) in
the past and found real issues so we'll stick with them for now.

### DIFF
--- a/e2e/docs/test.js
+++ b/e2e/docs/test.js
@@ -23,6 +23,7 @@ for (const baseURL of baseURLs) {
       }
 
       await page.getByRole("button", { name: "About" }).click()
+      await new Promise((resolve) => setTimeout(resolve, 1000)) // unflakes snapshot test :(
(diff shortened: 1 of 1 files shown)


camh- (approved):
:green_apple:

juliaogris:
I think I found a better fix for this!
e2e has again found a real issue. YAY sob.
For me on linux a get a flicker when opening _any_ dialog
in Firefox or Chrome. Removing the following lines fixes it.
https://github.com/evylang/evy/blob/7dbd68c964a829568130f88a10f9355d5e25586e/frontend/css/dialog.css#L18-L32

I will raise this as part of a separate PR.

PR #420, opened 2024-08-22
playsite: Improve sharing UX

juliaogris:
The primary button in the share dialog said Done, without having done the
main action of sharing: copying the URL. We have changed it to say
`Copy`, copying the URL. If you don't wish to copy the URL, just hit the
close button of the dialog. As a result we had to tweak JS, CSS and the
snapshot tests.

While playing with dialogs I've noticed a flicker on when the dialog
gets rendered which might have also caused flaky tests. We've removed
the fade-in for HTML dialogs.


camh- (approved):
:green_apple:

PR #422, opened 2024-08-24
tools: Upgrade deps and tooling

juliaogris:
Upgrade deps and tooling as described in docs/development/upgrade.md. Refine the
documentation along the way and add a last manual step ensuring all external
URLs still work.

The recent upgrade wasn't long ago, but firebase had an upgrade which nagged on
every local deploy and go had a new release as well, so I thought, let's just
get it done.

Most things went smooth this time, except for the golangci-lint upgrade.


camh- (approved):
:frog:

PR #425, opened 2024-08-26
labsite: Tweak CSS and scrolling behavior

juliaogris:
Refactor labsite CSS to use more modern CSS features, especially nesting.

Finally fix the broken looking light-theme sample selection on the playsite and
labsite.

Tweak the scrolling behavior for Next button clicks: we always scroll to top of
notes section. We also use "smooth" scrolling (rather than immediate jumping to
target), which makes it more obvious what's going on. Tested to work on iOS and
MacOS, Chrome, Safari, Firefox. Unfortunately no dice on my Ubuntu laptop with
Smooth scrolling on any browser 🤷‍♀.

Link: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_nesting/Using_CSS_nesting

---

Deployed at https://lab.evytest.dev

### DIFF
--- a/frontend/lab/css/overrides.css
+++ b/frontend/lab/css/overrides.css
@@ -26,6 +26,12 @@ html {
   height: 100%;
   padding: 8px 16px var(--editor-padding-bottom) 16px;
   scroll-behavior: smooth;
+  position: relative;
+}
+.notes:after {
+  content: "";
+  display: block;
+  height: 80%;
 }
 
 @media (min-width: 768px) {
--- a/frontend/lab/css/overrides.css
+++ b/frontend/lab/css/overrides.css
@@ -55,19 +55,6 @@ html {
   font-size: 0.875rem;
 }
 
-.notes details {
-  margin-bottom: 12px;
-}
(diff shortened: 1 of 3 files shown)


camh- (approved):
:green_book: 

It scrolls smoothly for me - chrome on linux.

I'm not sure why you wanted it to scroll leaving part of the next button chopped off. I think I would have scrolled the button all the way off the top. But just preference so merge away!

juliaogris:
> 📗
> 
> It scrolls smoothly for me - chrome on linux.
> 
> I'm not sure why you wanted it to scroll leaving part of the next button chopped off. I think I would have scrolled the button all the way off the top. But just preference so merge away!

ok, I've changed it. I only had it that way so it was more obvious what's going on
if you don't have smooth scrolling. Will ask @jasonstrachan what he thinks next I see him,
thanks for looking!

camh-:
> I only had it that way so it was more obvious what's going on
> if you don't have smooth scrolling

That makes sense. If you feel that is needed you can stick with it.

PR #428, opened 2024-08-28
lint: Ignore unwrapped errors in main package

juliaogris:
Ignore unwrapped errors in Go `main` packages. We only use Sentinel
errors when provided by packages. Disable err113 linter for main
packages and remove inline `//nolint:` directives.

As I fixed this up I revisited building the go sub-module in `learn/`
and made it consistent in docs and Makefile.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -8,6 +8,15 @@ issues:
     - "^G304: Potential file inclusion via variable"
     - "^G306: Expect WriteFile permissions to be 0600 or less"
     - "^G301: Expect directory permissions to be 0750 or less"
+  exclude-rules:
+    # Dynamic errors are OK in main packages. Packages should use sentinels.
+    - path: "^([^/]*|build-tools/.*)$"
+      linters:
+        - err113
+    # Used with `cd learn; golangci-lint run`
+    - path: "^cmd/.*$"
+      linters:
+        - err113
 
 linters:
   enable-all: true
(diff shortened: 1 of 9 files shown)


.golangci.yaml
```diff
@@ -8,6 +8,11 @@ issues:
     - "^G304: Potential file inclusion via variable"
     - "^G306: Expect WriteFile permissions to be 0600 or less"
     - "^G301: Expect directory permissions to be 0750 or less"
+  exclude-rules:
+    # Dynamic errors are OK in main packages. Packages should use sentinels.
+    - path: "^([^/]*|build-tools/.*|cmd/.*)$"
```
camh-, line 13: I'm not sure how the `cmd/` part works. It is anchored to the start of the path, yet the file with the lint error is `learn/cmd/levy/ain.go`.

Ah, I see now - for `learn` you first `cd learn`. I would consider duplicating this rule for the things in `learn/` with an appropriate comment.

camh- (commented):
Just one suggestion.

But also: https://github.com/evylang/evy/blob/main/learn/pkg/learn/learn.go#L129 - it seems there is one instance of this ignore not in a main package.

.golangci.yaml
juliaogris: ok, fair enough.
updated. I've also added a Makefile cleanup commit to this pr to used `go CMD -C dir`
everywhere. seems related.

camh- (approved):
:green_apple:

PR #429, opened 2024-08-28
labsite: Tweak UI after session with school

juliaogris:
- Make Hint content un-copyable
- Remove copy icon
- Remove extra notes space without editor. There won't be Next buttons (overview only)
- Tweak smooth scrolling offset


camh- (approved):
:green_apple:

PR #430, opened 2024-08-29
make: Rework make targets for parallelism

juliaogris:
Rework make targets for parallelism, so that we can run

    make ci -j $(nproc)

to speed up the CI builds locally and on GitHub Actions. This required a
little of renaming of files used for markdown generation. It also seems
that the node tool prettier and stylelint cannot be installed / run via npx
in parallel.

In a preparatory step add check-uptodate to CI in a fully separate job and
remove it from the test job, to speed up the test CI job. Parallelizing make
does not help with this as `check-uptodate` interferes with testing and needs
to be run sequentially - or on different instances in parallel.

FWIW, hyperfine tells my that my local builds sped up around 2.4 times:

    Summary
      'make ci -j 16' ran
        2.41 ± 0.04 times faster than 'make ci'

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -151,22 +151,22 @@ conform: install
 # --- Docs ---------------------------------------------------------------------
 doc: doctest godoc toc usage
 
-DOCTEST_CMD = ./build-tools/doctest.awk $(md) > $(O)/out.md && mv $(O)/out.md $(md)
+DOCTEST_CMD = ./build-tools/doctest.awk $(md) > $(O)/doctest-out.md && mv $(O)/doctest-out.md $(md)
 DOCTESTS = docs/builtins.md docs/spec.md docs/syntax-by-example.md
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -16,7 +16,7 @@ all: build-full test lint
 
 test: test-go test-tiny test-cli check-coverage
 
-lint: lint-go lint-sh check-prettier check-style check-fmt-evy conform
+lint: lint-go lint-sh check-style check-fmt-evy conform
```
camh-, line 19: Why has `check-prettier` been removed from this? Was that from experimenting with parallelism and not put back?

juliaogris: thanks. fixing.

juliaogris: addressed.

camh- (approved):
:four_leaf_clover:

PR #431, opened 2024-08-30
make: Fix ci-check-uptodate for parallelism

juliaogris:
Fix ci-check-uptodate for parallelism, it seems Node is still stepping
on its toes when check-prettier and check-style are run in parallel.
Fix it by running `make install-node-deps` only once _before_ any
prettier of stylelint executions with a .WAIT in the
`ci-check-uptodate` target.

This was an oversight in previous commit and PR (my bad).

Related-PR: https://github.com/evylang/evy/pull/430
Related-Commit: fab5fc320b9e9db62d5f53d8b82026e79d2d4594


camh- (approved):
:four_leaf_clover:

PR #435, opened 2024-09-06
evy: Improve version information

juliaogris:
Improve version information to use debug.BuildInfo if version string is
not explicitly set by compiler flags. This should provide Version
information when installed via

	go install evylang.dev/evy@latest


camh- (approved):
:broccoli:

PR #436, opened 2024-09-08
labsite: Fix mobile three column layout

juliaogris:
Add tweaks so the three columns

- Notes
- Editor
- Output

if specified are displayed properly.

Add logic to navigate between them.

---

I've deployed these changes to https://lab.evytest.dev. Ideally tested manually
on mobile or Browser dev tools with <767px width view / responsive enabled.


camh- (approved):
:four_leaf_clover:

PR #439, opened 2024-09-15
Change the license to MIT

juliaogris:
Change the license to MIT because it is shorter, simpler and easier to
understand than Apache 2.0 and therefore more aligned with what Evy
tries to do. I have checked with all code contributors that this is
ok.


camh- (approved):
:green_book:

PR #440, opened 2024-09-16
evy: Add `pi` as global constant to Evy

juliaogris:
Add `pi` as global variable to Evy. Pi is often needed
when calculating radians from degrees and/or placing shapes in a
circle.

In all examples where we defined `pi` as a local constant, we now use
the pre-defined global one.


frontend/lab/samples/ifs/pi.md
```diff
@@ -124,7 +124,6 @@ And at the bottom of the loop, add:
 ```evy
 while // ...
   // ...
-  pi := reds / i * 4
```
camh-, line 127: This is estimating π, isn't it? So you should not be removing this I think. Perhaps this sample could show the delta between the builtin π and the estimated π?

camh- (commented):
A couple of thoughts outside individual lines:

The PR/commit message says "constant" but there are no constants in Evy. It really is a variable that could be re-assigned if wanted.

π should be added to the builtins docs somewhere.

frontend/lab/samples/ifs/pi.md
juliaogris: that's a great idea!

juliaogris: fixed.

pkg/evaluator/builtin.go
```diff
@@ -138,9 +147,19 @@ func newBuiltins(rt Runtime) builtins {
 		"input":   {Name: "input", Params: inputParams},
 		"animate": {Name: "animate", Params: numParam},
 	}
-	globals := map[string]*parser.Var{
-		"err":    {Name: "err", T: parser.BOOL_TYPE},
-		"errmsg": {Name: "errmsg", T: parser.STRING_TYPE},
+	globals := map[string]global{
+		"err": {
+			val:       &boolVal{V: false},
+			parserVar: &parser.Var{Name: "err", T: parser.BOOL_TYPE},
```
camh-, line 153: nit: can you swap these two over (and below). Then it matches the order declared in the struct but also doesn't seem backwards to me: var then value is the ordering I'd expect.

camh- (approved):
:green_apple: with comment

pkg/evaluator/builtin.go
juliaogris: of course, agreed. fixed.

PR #441, opened 2024-09-16
labsite: Change book-icon to back-icon for mobile

juliaogris:
Change book-icon to back-icon for mobile: When on mobile only one of
these three screens is showed at a time: Notes, Code, Output. Clicking
the primary button lets you move from Notes -> Code <-> Output. To get
back to the Notes we had a book icon in the top right corner which
caused confusion. The back button is hopefully more logical and also
aligns with the whole screen swipe to to left.


camh- (approved):
:green_book: on the arrow, but as discussed there may be other ways coming that address the unity of navigation devices.

PR #442, opened 2024-09-16
playsite: Remove slide back transition on navigation

juliaogris:
When navigating to the next playground sample or lab challenge while on
the output view on mobile, we had this awkward: slide back to notes or
code first, than load new code. It seems more natural and intuitive to
just load the new contents - so for hashChange transition remove
animation by adding and eventually removing a CSS class that sets the
translate transition duration for the main container to 0 seconds.

### DIFF
--- a/frontend/play/css/index.css
+++ b/frontend/play/css/index.css
@@ -267,11 +267,14 @@ button.share > div {
   .main .editor-wrap {
     margin-top: 4px;
   }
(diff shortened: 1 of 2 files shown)


camh- (approved):
:green_book: I can't say I properly understand this stuff, but playing with it on my mobile, it works fine, so LGTM

PR #445, opened 2024-09-19
make: Fix race generating markdown docs

camh-:
Sequence the `doctest` and `toc` prerequisites in the `doc` target as
both of the prerequisites change the same files. If they run in parallel
and start with the same input, and one changes a file but the other does
not and the one that does not finishes last, you end up with an
unchanged file, wiping out the changes of the first. This can cause the
`check-uptodate` target to falsely report the docs are up to date.

This seems to have happened when adding the `Pi` variable to the
built-in docs. The PR successfully passed CI, yet the table of contents
did not contain the entry for `Pi`. Looking at the CI run [1], both
`doctest.awk` and `toc.awk` ran close together and the toc should have
been updated causing the workspace to be dirty, but because
`doctest.awk` takes a lot longer to run on `builtins.md` (as it has to
run and format many evy programs), it is likely it generated a file
without the updated toc and it all appeared clean.

Sequencing these two prerequisites should let `doctest` run to
completion first, then `toc` runs, using the output of `doctest` as
input. If either make changes, we should now fail the up-to-date
check.

Update the table of contents of `builtin.md` that got missed due
to this race.

Link: https://github.com/evylang/evy/actions/runs/10901756581/job/30252189731#step:3:117

### DIFF
--- a/docs/builtins.md
+++ b/docs/builtins.md
@@ -34,7 +34,7 @@ see [syntax by example](syntax-by-example.md).
 8. [**Random**](#random)  
    [rand](#rand), [rand1](#rand1)
 9. [**Math**](#math)  
-   [min](#min), [max](#max), [abs](#abs), [floor](#floor), [ceil](#ceil), [round](#round), [pow](#pow), [log](#log), [sqrt](#sqrt), [sin](#sin), [cos](#cos), [atan2](#atan2)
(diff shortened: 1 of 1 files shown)


juliaogris (approved):
Perfect, thank you so much!
:four_leaf_clover:

PR #446, opened 2024-09-20
play-site: Add Game of Life to interact samples

camh-:
Add an implementation of Conway's Game of Life to the interact
samples. While this is titled "game", it is not really a game so does
not really belong in the games section. It features a few different
modes of interaction, so let's place it there.

Link: https://en.wikipedia.org/wiki/Conway%27s_Game_of_Life


juliaogris (approved):
This reads so nice and runs really well, thank you!
I've tried a bit to find anything to improve, but no luck...
My only comment: this is snake_case where we have settled on 
camelCase for most samples - but I'm happy to reconsider.
  🐉
    🐉
🐉🐉🐉

camh-:
> My only comment: this is snake_case where we have settled on
> camelCase for most samples - but I'm happy to reconsider.

Yeah I meant to change that. For whatever reason I seem to write
snakes when writing evy. Fixed.

PR #450, opened 2024-10-16
labsite: Re-write relative URLs

juliaogris:
Re-write img and link URLs in labsite-gen tool from correct location of relative
to source .md file to correct relative location to lab/index.html file when
deployed.

This change improves readability and generation of Markdown lab source files.

In a preparatory step mark `lab/**/*.htmlf` as generated for linguist and PR
reviews.


camh- (approved):
:green_book:

PR #451, opened 2024-10-24
labsite: Rework Markdown source for Next button

juliaogris:
Rework Markdown source for Next button. Previously we required a
paragraph that only contained:

	[Next]

This did not look great as rendered Markdown on GitHub. The Thematic
Break or `<hr/>` is a more natural fit and has not used in any other
way in the lab so far.


camh- (approved):
:green_book:

PR #458, opened 2024-12-19
all: Rename "runtime" to "platform"

camh-:
Rename uses of "runtime" where it refers to the CLI runtime and the browser
runtime to "platform". Runtime usually refers to the thing doing the
running, such as the evaluator or bytecode VM and has started to cause a
little confusion or cumbersome language now that the bytecode VM is being
developed. Initial talk of perhaps targeting micro-controllers as
"platforms" also suggests this should be called platform.

There are still a few uses of the word "runtime" left - these refer to 
things like "runtime errors" where the term still makes sense, as it does
not refer to the Evy abstraction for runtime/platform.

Fix a lint error in `learn/pkg/learn/markdown.go` since golangci-lint wants
to complain about this now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ Use [Homebrew] to install `evy`.
 The Evy interpreter is written in [Go] and built using the Go and
 [TinyGo] compilers. TinyGo targets [WebAssembly], which allows Evy
 source code to be parsed and run in a web browser. The browser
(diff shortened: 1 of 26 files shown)


juliaogris (approved):
Looks great to me, I checked out the branch and double checked all remaining uses of "runtime" and they seem a lot more fitting now.

Thank you!

PR #459, opened 2024-12-20
tools: Upgrade

juliaogris:
Upgrade hermitised and node tools according to docs. It all went smoothly apart
from Playwright which required a bunch of snapshot updates again and is still
not working properly.

docs: https://github.com/evylang/evy/blob/5b196cbe154c60d68b2c2deffa41e118985ffba9/docs/development/upgrade.md


camh- (approved):
:broccoli:

PR #460, opened 2025-01-20
parser: Implement for niladic function calls without parens

juliaogris:
Implement niladic function calls without parens, so that the following Evy
code is now valid:

   if rand1 > 0.5
    print "Heads"
end

   print read

    print rand1 rand1+10

which reads better than `if (rand1) > 0.5` etc.

Along the way fix typo in name of builtin `"rand1"` (name is used in error
messages only). Improve error message where a function is called and expected
to have a return value.

### DIFF
--- a/pkg/parser/expression.go
+++ b/pkg/parser/expression.go
@@ -49,20 +49,22 @@ var precedences = map[lexer.TokenType]precedence{
 
 func (p *parser) parseTopLevelExpr() Node {
 	tok := p.cur
-	if tok.Type == lexer.IDENT && p.funcs[tok.Literal] != nil {
-		return p.parseFuncCall()
+	if tok.Type == lexer.IDENT && p.funcs[tok.Literal] != nil && !p.funcs[tok.Literal].isNiladic() {
+		return p.parseFuncCall(true)
 	}
 	return p.parseExpr(lowestPrec)
 }
 
-func (p *parser) parseFuncCall() Node {
+func (p *parser) parseFuncCall(isTopLevel bool) Node {
 	fc := &FuncCall{token: p.cur, Name: p.cur.Literal}
 	p.advance() // advance past function name IDENT
 	funcDef := p.funcs[fc.Name]
 	funcDef.isCalled = true
 	fc.FuncDef = funcDef
-	fc.Arguments = p.parseExprList()
-	p.assertArgTypes(fc.FuncDef, fc.Arguments)
+	if isTopLevel || !funcDef.isNiladic() {
+		fc.Arguments = p.parseExprList()
+		p.assertArgTypes(fc.FuncDef, fc.Arguments)
+	}
 	return fc
 }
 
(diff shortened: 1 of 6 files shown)


pkg/parser/expression.go
```diff
@@ -122,6 +124,14 @@ func (p *parser) isAtExprEnd() bool {
 	return p.isAtEOL()
 }
 
+func (p *parser) parseIdentExpr() Node {
+	funcDef := p.funcs[p.cur.Literal]
+	if funcDef != nil && funcDef.isNiladic() {
```
camh-, line 129: In `parseTopLevelExpr()` you don't bother with the `funcDef` var - you just index `p.funcs[]` twice. I suggest you do the same here or use the var over there.

pkg/parser/expression_test.go
```diff
@@ -300,3 +304,76 @@ func TestParseTopLevelExpressionErr(t *testing.T) {
...
+	for name, input := range inputs {
+		t.Run(name, func(t *testing.T) {
+			parser := newParser(input, testBuiltins())
+			_ = parser.parse()
+			assertNoParseError(t, parser, input)
+		})
+	}
+}
+
+type inputWithError struct {
```
camh-, line 339: This type could move into the `ParseNiladicErr()` function. It could even be anonymous there too, but up to you.

pkg/parser/expression_test.go
```diff
@@ -300,3 +304,76 @@ func TestParseTopLevelExpressionErr(t *testing.T) {
...
+		"rand1":       "print rand1",
+		"rand1-twice": "print rand1 rand1",
+		"rand1-expr":  "print rand1+10",
+		"rand1-toplevelexpr": `
+n := rand1+0.5
+print n
+`,
+		"rand1-toplevelexpr-space": `
+if rand1 > 0.5
+  print "big"
```
camh-, line 319: all the other test programs use a tab for indenting the block where there is one. this uses two spaces.

camh- (approved):
A few minor formatting comments, plus this one:

* The second commit message has some code indented with a tab and some with 4 spaces.

pkg/parser/expression.go
juliaogris: hmm, I tried this. Above I did it because it allowed me drop an extra nested if statement, which i think reads clearer, but hear I feel the local var reads clearer 
as it is not conditioned on `tok.Type == lexer.IDENT` - what do you think?

pkg/parser/expression_test.go
juliaogris: ok, i've done that. i usually don't do this either - along with my personal no nested funcs / types preference and a preference for shorter funcs. i know this is one of the few things we disagree on...

pkg/parser/expression_test.go
juliaogris: great, done, thanks.

pkg/parser/expression.go
camh-: All good. Keep as is.

camh- (approved):
:leafy_green: LGTM

PR #463, opened 2025-02-08
make: Fix prettier installation race

juliaogris:
Targets `docs`, `learn` and `labs` all use prettier which gets installed with

    npx --prefix $(NODEPREFIX) -y prettier # ...

Occasionally this seems to fail on CI when run concurrently. Fix by adding an
explicit dependency and prettier installation and a `.WAIT`.


camh- (approved):
:broccoli: LGTM

PR #464, opened 2025-02-08
svg: Prevent code with fused instructions

camh-:
Add a floating point conversion when multiplying when scaling coordinates
so that the multiplication does not get fused with an adjacent addition or
subtraction. This causes two instructions to be generated, a MUL and a
ADD/SUB, instead of a single FMADD or FMSUB. A fused instruction has
different rounding so ends up giving slightly different results to two
instructions. This was causing different values in the SVGs generated
between AMD64 and ARM64.

Revert "labs: Rework random rectangle generation" as we no longer need
to work around this by rounding in Evy code.

Remove the parens in `randrect.evy` as they are not needed any more due
to a recent change.

I've tested this on AMD64 and ARM64 and the same SVGs are now generated.

Link: https://go.dev/ref/spec#Floating_point_operators


juliaogris (approved):
Beautiful, thank you!💚

PR #466, opened 2025-03-05
play: Fix dash reset issues

juliaogris:
Fix dash reset issues with an explicit check for `dash`. Previously it resulted
in the incorrect:

	canvas.ctx.setLineDash([0])

Now it results in the correct:

	canvas.ctx.setLineDash([])

---

See https://evy-lang-stage-play--466-09qkylmw.web.app/#source=https://gist.githubusercontent.com/juliaogris/628d08c39d2710b04ce2fe0b17e7827f/raw/newton.evy


camh- (approved):
:leafy_green:

